PR #20, opened 2020-06-16
Add web client to docker compose

mvandenburgh:
This PR adds the web client to the docker-compose configuration. nginx is used to serve static files and proxy requests to the backend when appropriate.


web/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM node:lts-alpine as build-stage
+WORKDIR /app
+COPY package*.json ./
```
jbeezley, line 3: Probably should copy `yarn.lock` as well.

jbeezley:
This is great, thanks.

I'm seeing something that looks like css not being applied correctly.  Are you getting that as well, or is it just my build?  At a glance, I'm not sure what's causing it.  It almost looks like the vuetify css isn't loaded.

One other small thing.  Could you make sure to end files in a new line?  Sometimes programs don't read files correctly when that happens.  So it's good practice to configure your editor to do it.

mvandenburgh:
> I'm seeing something that looks like css not being applied correctly. Are you getting that as well, or is it just my build? At a glance, I'm not sure what's causing it. It almost looks like the vuetify css isn't loaded.

This should be fixed now. I didn't configure MIME types properly in nginx, which caused css to not be displayed. I just assumed the UI was still a work in progress when I saw that :)

> One other small thing. Could you make sure to end files in a new line? Sometimes programs don't read files correctly when that happens. So it's good practice to configure your editor to do it.

Ok, I added a new line to all the files I edited here.

jbeezley:
Looks good, thanks.

PR #29, opened 2020-07-09
Vuex, Typescript, Upgrade Vuetify, Vue router

subdavis:
Update stuff to latest versions, add some D&A boilerplate eslint config.


subdavis:
This isn't 100% functional, but it's a really solid foundation for the new feature work that has been requested (IMO).

This actually breaks faceted search left panel interactions because they're about to be rewritten and it wasn't worth the time to make them work in the old style.

PR #35, opened 2020-07-17
Add pipeline processing tables

jbeezley:
This is WIP until I get the final data files, but it seems to be working.


subdavis:
@jbeezley this broke docker-compose build

a "versions" directory is expected

```
backend_1  | FileNotFoundError: [Errno 2] No such file or directory: '/app/migrations/versions'
```

Maybe stick a `.gitignore` or a `README` in there to preserve the dir in git?

PR #40, opened 2020-07-17
Return attribute types in database summary

jbeezley:
@subdavis This updates the summary response to include the attribute type.  I replaced the per attribute count because I'm not sure it's that useful, but I can make it a nested object type if needed.

### DIFF
--- a/nmdc_server/aggregations.py
+++ b/nmdc_server/aggregations.py
@@ -1,0 +1,72 @@
+from typing import cast, Dict, Type
+
+from sqlalchemy import Column, func
+from sqlalchemy.orm import Session
+
+from nmdc_server import models, schemas
+
+
+def get_annotation_summary(
+    db: Session, model: Type[models.AnnotatedModel],
+) -> Dict[str, schemas.AttributeSummary]:
+    attribute = func.jsonb_object_keys(model.annotations)
+
+    # TODO: Figure out the correct type, or remove json aggregations
+    q = db.query(attribute, func.count(),).group_by(attribute)
+
+    attributes: Dict[str, schemas.AttributeSummary] = {}
+    for r in q:
+        attributes[r[0]] = schemas.AttributeSummary(count=r[1], type=schemas.AttributeType.string,)
+
+    return attributes
+
+
+def get_column_count(db: Session, column: Column) -> int:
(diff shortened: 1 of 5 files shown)


subdavis:
Those per-attribute counts are where these numbers come from

Screenshot from 2020-07-17 17-06-50

Should I just remove those boxes?

subdavis:
I just pulled them from the aggregate total, then I replaced all the things at the bottom with the facets from `project.omics_type`:

Screenshot from 2020-07-17 17-17-12

This is probably wrong.

jbeezley:
I suspect we will need some custom aggregations to fill in many of those boxes.  I can work on that next.  For now, at least I have kept the counts and added min/max to help populate the widgets.

E.g.
```
    "biosample": {
        "attributes": {
            "add_date": {
                "count": 964,
                "max": "2019-02-07T18:08:49.941000",
                "min": "2013-11-25T00:00:00",
                "type": "date"
            },
            "collection_date": {
                "count": 963,
                "type": "string"
            },
            "community": {
                "count": 964,
                "type": "string"
            },
            "depth": {
                "count": 392,
                "max": 2000.0,
                "min": 0.0,
                "type": "float"
            },...
```

jbeezley:
A couple of other notes about this:
1. I'm not attempting to infer the types for json attributes, so they are all strings.  I'm thinking it will be best to remove them from the query interface anyway at some point.
2. This implementation is not at all scalable.  I think moving all of this information into a materialized view will be most appropriate when the time comes.

subdavis:
> This implementation is not at all scalable. I think moving all of this information into a materialized view will be most appropriate when the time comes.

Is this a performance scaling concern?  I've tried to lazy-load as much as possible while keeping the interface responsive. 

If these queries are going to be too slow to perform requests every time a checkbox updates, I may have to rethink a bit of the UI state management.

Or do you mean it won't scale in terms of technical complexity?

jbeezley:
Specifically referring to the `/api/summary` endpoint.  It runs a full table scan for nearly every column in the database, and it is executed every time the page is loaded.  It's not an immediate concern, but will be the first place where we get performance issues.

subdavis:
Oh, got it.  So that could easily be a separate summary table that gets populated during migrations or something in the future.

Thanks.

Issue #53, opened 2020-07-20
api/project/search 500 error


subdavis:
## Request

``` bash
curl -X POST "http://localhost:8080/api/project/search?limit=100" -H  "accept: application/json" -H  "Content-Type: application/json" -d "{\"conditions\":[]}"
```

## Response:

```
backend_1  | 172.21.0.1:47238 - "POST /api/project/search?offset=0&limit=100 HTTP/1.1" 500
backend_1  | [2020-07-20 14:16:13 +0000] [24] [ERROR] Exception in ASGI application
backend_1  | Traceback (most recent call last):
backend_1  |   File "/usr/local/lib/python3.8/site-packages/uvicorn/protocols/http/httptools_impl.py", line 385, in run_asgi
backend_1  |     result = await app(self.scope, self.receive, self.send)
backend_1  |   File "/usr/local/lib/python3.8/site-packages/uvicorn/middleware/proxy_headers.py", line 45, in __call__
backend_1  |     return await self.app(scope, receive, send)
backend_1  |   File "/usr/local/lib/python3.8/site-packages/fastapi/applications.py", line 181, in __call__
backend_1  |     await super().__call__(scope, receive, send)
backend_1  |   File "/usr/local/lib/python3.8/site-packages/starlette/applications.py", line 102, in __call__
backend_1  |     await self.middleware_stack(scope, receive, send)
backend_1  |   File "/usr/local/lib/python3.8/site-packages/starlette/middleware/errors.py", line 181, in __call__
backend_1  |     raise exc from None
backend_1  |   File "/usr/local/lib/python3.8/site-packages/starlette/middleware/errors.py", line 159, in __call__
backend_1  |     await self.app(scope, receive, _send)
backend_1  |   File "/usr/local/lib/python3.8/site-packages/starlette/exceptions.py", line 82, in __call__
backend_1  |     raise exc from None
backend_1  |   File "/usr/local/lib/python3.8/site-packages/starlette/exceptions.py", line 71, in __call__
backend_1  |     await self.app(scope, receive, sender)
backend_1  |   File "/usr/local/lib/python3.8/site-packages/starlette/routing.py", line 550, in __call__
backend_1  |     await route.handle(scope, receive, send)
backend_1  |   File "/usr/local/lib/python3.8/site-packages/starlette/routing.py", line 227, in handle
backend_1  |     await self.app(scope, receive, send)
[...]

PR #54, opened 2020-07-20
Fix query error due to incorrect schema

jbeezley:
There is a problem coercing sqlalchemy association proxies to lists.
Because this is a blocker, I'm commenting these fields out until I come
up with a solution.


jbeezley:
Fixes #53

PR #56, opened 2020-07-20
Fix 'open in gold' link

jbeezley:


jbeezley:
Seems to fix https://github.com/microbiomedata/pilot/issues/48, though I think there are some entities now that don't come from GOLD, so the field is optional.

PR #67, opened 2020-07-21
Server study PI images

### LINKED ISSUE #58: Add profile pictures to db

jbeezley:
@subdavis 
Adds a derived field `principal_investigator_image_url` to the study response.  This contains an absolute URL to something like `/api/principal_investigator/<uuid>` which returns a jpeg image.

Fixes #58


jbeezley:
Also of note, all of the association proxy fields (including the principal investigator name) are not queryable at moment.  I'm working on a fix.

PR #73, opened 2020-07-21
Convert strings to numbers if required

subdavis:
@jbeezley because the URL query string causes all numbers to be converted to strings, on page reload, if you had a numeric condition, the value would become a string.

It seems like the server is handling this, but I thought it wasn't, and wrote this to deal with it.  I don't really like this solution, because it relies on loading the results from `/api/summmary` to get the type info before the search can be dispatched.

LMK what you think.

### DIFF
--- a/web/src/data/api.ts
+++ b/web/src/data/api.ts
@@ -106,7 +106,7 @@ export interface MetagenomeAnnotationResult extends DerivedDataResult {
 
 export type MetaproteomicAnalysisResult = DerivedDataResult
 
-interface AttributeSummary {
+export interface AttributeSummary {
   count: number;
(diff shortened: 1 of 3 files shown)


jbeezley:
Pydantic has some magical ways of coercing values.  In this case, the first type it tries is a number.  This is actually how it deals with datetimes as well.  This actually leads to a potential bug where you are searching a string field, but the value happens to be a number.  I'll deal with it if it ever comes up.

Could you not json encode the query string?  It's already kind of a mess of escaped characters, so I don't know that human readability is a concern.

subdavis:
Yeah I can do that instead I think

Issue #112, opened 2020-07-28
Remove faceting fields that have unique values


jbeezley:
E.g. anything like `name` or `description`.  We still want `principal_investigator_name`.

subdavis:
did some of this, I'll do more.

subdavis:
Turns out I did this already and forgot.

Issue #110, opened 2020-07-28
Add search by data object


jbeezley:
![data_objects1](https://user-images.githubusercontent.com/31890/88670197-c40a8000-d0b2-11ea-9578-d91687a487dd.png)
![dataobjects2](https://user-images.githubusercontent.com/31890/88670205-c4a31680-d0b2-11ea-81d4-e43b97f575eb.png)

subdavis:
Issue out of date.

PR #118, opened 2020-07-28
Adds data_object, fix many aspects of individual results

subdavis:


subdavis:
fixes #112, first pass on #110 until we get more info

PR #120, opened 2020-07-29
Add endpoints to list data objects

jbeezley:
@subdavis I started this to work out how to show the data objects.  I got in a little over my head on the web client changes.  There are a couple of issues:

1. The counts returned by the standard search interface are for **all** related data objects joined by project_id whereas the data object list is only for outputs of the given analysis (my best guess for the information desired).
2. The API interface is necessarily different from the other result lists, so it breaks some of the abstractions in the components.

I'm starting to think that just putting the data object list directly onto the individual result component for the relevant types.  There are never more than a handful of data objects associated with a single pipeline entity.

Do you have any suggestions?  You can take over this PR and remove my client commit if you want.


subdavis:
Hmm.  Could you point to what you got stuck on?  That may point to an over-complication in the client.

> I'm starting to think that just putting the data object list directly onto the individual result component for the relevant types.

I like this idea.  The "relevant types" would be project, biosample, and study right?

jbeezley:
> Hmm. Could you point to what you got stuck on? That may point to an over-complication in the client.

I didn't put a lot of effort into it, but I stopped when I found [AttributeList.vue#L10](https://github.com/microbiomedata/nmdc-server/blob/list-data-objects/web/src/views/IndividualResults/AttributeList.vue#L10) would have to create custom behavior for this one button.  Not the end of the world, but more than I was willing to take on last night.

> > I'm starting to think that just putting the data object list directly onto the individual result component for the relevant types.
> 
> I like this idea. The "relevant types" would be project, biosample, and study right?

No, project, reads_qc, metagenome_assembly, metagenome_annotation, and metaproteomic_analysis.  They are the tables that have data object outputs.

subdavis:
@jbeezley ready for your review.  Merge if you're happy with it.

jbeezley:
Seems to be missing `components/DataObjectsList.vue`.

subdavis:
Updated.

PR #146, opened 2020-08-20
Client/tree

### LINKED ISSUE #121: Tree view for GOLD terms
Should start with a mockup

subdavis:
Screenshot from 2020-08-20 12-03-50

## problems with this approach

GOLD is a series of 5 actual facets, so every leaf on the tree always has a depth of 5, and when I transform between the tree and the condition list, I have to describe the selected nodes on the tree in terms of their actual facet.  Right now, I've made a meta-facet type thing called "GOLD Classification" for the purposes of adding this tree menu to the sidebar.
ENVO is different.  Each field is a flat list of all the node values.  Currently the server doesn't provide enough information for me to construct a tree, but I imagine it will produce something similar to the sankey API.  If it does not produce something similar (for example, because the tree is not a fixed depth) then both the inputs and the outputs are sufficiently dissimilar that they can't share code.

This is compounded by the fact that v-treeview  reports selection state in an incredibly unhelpful way.   If you select Environment  at level 0, for example, the tree component reports that there are 15 selections: all the leaves underneath Environment.   The naive way to report this user event to the server would be to create something like 30 conditions for all the levels in the tree that got selected.  In order to discover the common ancestor and select the minimum number of conditions, you have to take that flat list of 15 leaf nodes, (which is effectively a subset of the same data shape that the sankey api returns), construct another tree from that, then do tree diffing of some kind.

I actually don't see a way aronud the tree diffing if you want to build a query based on common ancestors: either we do it on the server or we do it on the client.

fixes #121


subdavis:
Screenshot from 2020-08-20 14-31-35

Good enough for a first pass.

PR #162, opened 2020-09-09
Add a condition schema for tree queries on gold terms

jbeezley:
This adds an explicit condition type to be used for queries on the Gold environment hierarchy.  ~~It looks like this:~~
```
{
	"table": "biosample",
	"field": "gold_tree",
	"op": "tree",
	"value": {
		"ecosystem": ["ecosystem 1", ...],
		"ecosystem_category": ["category 1", ...],
		"ecosystem_type": ["type 1", ...],
		"ecosystem_subtype": ["subtype 1", ...],
		"specific_ecosystem": ["specific ecosystem 1", ...]
	}
}
```
~~In this implementation, the query it produces is an "OR" of all individual attributes (even between distinct attributes).  This would not be possible under the existing query schema because all filters on distinct attributes are combined with an "AND".~~

~~I *think* this is the most logical way to implement the desired behavior, but I would be interested hearing @subdavis's opinion before finalizing it.~~

See the updated schema below.

### DIFF
--- a/nmdc_server/query.py
+++ b/nmdc_server/query.py
@@ -1,7 +1,7 @@
 from datetime import datetime
 from enum import Enum
 from itertools import groupby
-from typing import Any, Dict, Iterator, List, Tuple, Union
+from typing import Any, Dict, Iterator, List, Optional, Tuple, Union
 
 from pydantic import BaseModel
 from sqlalchemy import and_, distinct, func, inspect, or_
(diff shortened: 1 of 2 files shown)


jbeezley:
I Modified the condition schema to this:
```json
{
    "table": "biosample",
    "field": "gold_tree",
    "op": "tree",
    "value": [ 
        { 
            "ecosystem": "ecosystem_0",
            "ecosystem_category": "category_0",
            "ecosystem_type": "type_0",
            "ecosystem_subtype": "subtype_0",
            "specific_ecosystem": "specific_0",
        },
        { 
            "ecosystem": "ecosystem_0",
            "ecosystem_category": "category_0",
            "ecosystem_type": "type_1"
        }
    ]
}
```
The query is generated as an "AND" condition inside the object, and an "OR" condition between each object.  Though not required by the API, it is expected the client will resolve parent entities (e.g. if a specific "ecosystem_type" is chosen, it will also resolve "ecosystem_category" and "ecosystem" as in the example).

@subdavis Let me know if for any reason this doesn't work for the web client, we can make adjustments.

subdavis:
This looks like it will fit the existing UI really well.  Thanks.

PR #183, opened 2020-10-27
Client/181 usability

### LINKED ISSUE #181: Usability feedback action items.
# Priority: high

- [x] Toggle selects for filtering for all appropriate graphics.
- [x] Animations when filters change
- [x] Info/Guide tooltips on some visual elements (Sankey, bar, range selector)

# Priority: lower

Mostly because these would involve forking an upstream library, so the implications for cost/benefit for longer-term maintenance are different.  I'll take a look at the options of the library and see if there are tweaks we can make.

https://github.com/riophae/vue-treeselect hasn't been updated in a year and is overflowing with open issues, so it may be abandonware at this point.

- [ ] Remove chips or add full hierarchy to graph chips in GOLD list
- [ ] Right-align counts 

# Requires discussion

Feedback that doesn't necessarily make sense to us to address at this time.

- [ ] Move filters to top of the search result page.  Brandon thinks this goes against convention for faceted search.  In most interfaces, users will expect the filters in a left sidebar.  I think Faiza agrees, but please feel free to comment.
- [ ] Switch to drop-down menu for category.  Brandon objects to this mostly because it's taking a 1-click operation and turning it into a 2-click operation without much justification.  Faiza raised a more nuanced usability concern related to understanding that facets persist through data categories.
- [ ] Search bar for results: Brandon made this lower priority because it's not clear yet that this would have utility.  In many of the categories, the result item title has very low variance (they're basically all the same), so the current display is already not ideal. At some point, again particularly on some of the lower 5 (Environment Onward), it was noted that showing individual results may not even be the goal, because users will probably be using this tool to search over much more data than a tabular search would provide value on, and will be basing their query exclusively on study and aggregate search metrics.   For full text search, I want to consider implementing this as a facet (such as for study title, where we replace the checkboxes with a new freeform text search). Otherwise, it's a special edge case where it works as a facet but gets applied to the filters separately.  Would this search only filter the table but leave aggregate queries the same?  Does it get captured in the URL for copy-paste?  Requires discussion.

@faiza-a please let me know if any of that needs revising.

subdavis:
partially addresses #181

### DIFF
--- a/web/src/components/charts/Histogram/Histogram.vue
+++ b/web/src/components/charts/Histogram/Histogram.vue
@@ -68,7 +68,10 @@ export default {
       return this.bin(this.dataValues);
     },
     binnedDomain() {
-      return this.binned.map((b) => b.length);
+      return this.binned.map((b) => ({
+        height: this.y(b.length),
+        key: b.x0,
+      }));
     },
     dataValuesRaw() {
       return this.data.map((d) => d.valueOf());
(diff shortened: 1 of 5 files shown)


subdavis:
Missing info tooltips, I'll add those in a follow-up.

PR #203, opened 2021-01-27
Add gene functions from metagenome annotations

jbeezley:
This add the gene function relations for metagenome annotations.  As I suspected, this makes running ETL on deploy (and testing it in github actions) unsustainable as it will take at least 15 minutes (probably more).  I'm going to have look into a new strategy for manually triggering the ETL.

### DIFF
--- a/nmdc_server/cli.py
+++ b/nmdc_server/cli.py
@@ -1,3 +1,4 @@
+import logging
 from pathlib import Path
 
 from alembic import command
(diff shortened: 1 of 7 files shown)


jbeezley:
It looks like the full import of gene functions from metagenome annotations is going to take at least 3 hours.  For now, I'm going to limit the number of gene functions imported per metagenome annotation to 100.  This keeps the ingest time short and will allow us to more quickly iterate on everything else.

At some later point, I'll move the ingest out of the deployment... perhaps a celery task triggered from an endpoint.

PR #205, opened 2021-01-28
Invert the relationship between biosample and project

jbeezley:
This was a mistake in the data model that needs to be corrected now that there are multiple projects per biosample.

A couple of weird things about the new schema:
* biosample_id is nullable on the project because we don't have biosample information on all projects
* study_id exists on both biosample and project so that queries can be joined both ways.  This will be unnecessary if we make biosample_id on project non-nullable.

I'm only 80% sure the way the tables are being joined now produces the expected result.  It's tricky with two different ways to join with the study table.  We will want to have the people who know the data better vet the search results once we have the data import ironed out.

@subdavis I don't think this will break anything you are working on.  The only big difference is that the biosample search results will now contain multiple projects in the list rather than always one.

It may break when you try to upgrade your database.  I'm having a strange problem where alembic's migrations don't apply.  Until I figure that  out, you can just run `docker-compose run backend psql postgres -c 'drop database nmdc;'` and it will start the database from scratch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ NMDC_MONGO_PASSWORD=changeme
 
 ```
 nmdc-server truncate
-nmdc-server migrate
+alembic -c nmdc_server/alembic.ini upgrade head
 nmdc-server ingest
 ```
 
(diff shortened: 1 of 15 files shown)


jeffbaumes:
I believe the goal is to have biosample records for all projects. @dehays where are we on this? If it be more straightforward to always create dummy biosamples for any biosample ID that doesn't have one, I'd say do that.

PR #256, opened 2021-02-11
Use inner joins for study and biosample queries

jbeezley:
I don't think outer joins between many of these tables is necessary (at least given the current schema).  This speeds everything up quite a bit, so I'm going to try deploying this without the pre-baked study response to see what it's like.

There are a few other places where I think I can get away with similar optimizations, so I might not need a full refactor to get performance good enough for the current data size.


jbeezley:
I decided to keep the mocked study call.  With KEGG id's it's still too slow, I'm seeing what I can do to improve that query next.  If I can't come up with anything, I'll just update the response for tomorrow.

PR #277, opened 2021-02-13
Refactor table joining for more efficient queries

jbeezley:
I still need to implement faceting before it can be merged.  This is looking very promising with study search responses taking around 350ms even when including gene function conditions.  Biosample searches take around 500ms with the extra time being spent primarily in serialization of the response.


subdavis:
It's pretty trivial to switch the client back to live results.

Feel free to uncomment the lines in data/api.ts

Otherwise, I can do it after this merges.

jbeezley:
I'm going to merge this now.  There may be some bugs due to combinations of join conditions that I didn't implement.  Also, I'm not sure the omics aggregations on the study response is returning the desired quantities.

In any case, the responsiveness on most of the interactions is near instant now, so it should be easier for the users to vet the results.

PR #349, opened 2021-04-26
Incorporate UX Design Feedback Round 2

### LINKED ISSUE #341: Incorporate UX design feedback
This is feedback shared by @pvangay from the UX review.

subdavis:
Screenshot from 2021-04-26 12-21-59
Screenshot from 2021-04-26 11-55-36
Screenshot from 2021-04-26 11-55-12
Screenshot from 2021-04-26 11-54-50

I think this takes care of all the feedback we decided to move on from https://www.figma.com/file/h3v0xOrdZSYWSCRZhxQcXw/Improvements-based-on-'Mockup-Report-SGCI-NMDC-Usability-Consulting.pdf'?node-id=0%3A1

* Still using lorem ipsum for hover text.  I can switch these to full dialogs with a dismiss button if you think they need more than just a sentence or two.
* I went with sort of a hybrid for the download table change.  I liked the dedicated workflow header more than just the darkened background.  I also added a duplicate workflow header on the top of the N > 1th page of results so you don't have to remember what workflow section you're in page-to-page.
  * I didn't make it collapsible because there's already a lot going on in here and it seemed premature.  Also, only some omics types have this workflow header now (the column is removed) so it'd be kinda awkward to have to expand some tables and others just be normal tables.  Feedback welcome.
 

fixes #341

### DIFF
--- a/web/src/colors.js
+++ b/web/src/colors.js
@@ -12,6 +12,7 @@ export default {
   green,
   blue,
   purple,
+  grey: colors.grey,
   primary: purple,
   secondary: red,
   accent: orange,
--- a/web/src/components/Presentation/ConditionChips.vue
+++ b/web/src/components/Presentation/ConditionChips.vue
@@ -124,6 +124,7 @@ export default Vue.extend({
                 field: group.field,
                 table: group.table,
                 isOpen: menuState[group.key],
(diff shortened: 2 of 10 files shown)


subdavis:
Menu dismiss

Screenshot from 2021-04-28 16-59-59

Issue #221, opened 2021-02-03
Add units (cm) for depth and possibly other fields


jeffbaumes:

subdavis:
done in #225

PR #351, opened 2021-04-28
Use units from database

### LINKED ISSUE #338: Provide mechanism for providing units to the client
Ultimately just involves communication between client and server on the portal, but may need a bit of discussion with @cmungall as to how data with units should be stored/presented vs. normalized for query.

subdavis:
@jbeezley this is ready to go, but the depth unit should be corrected to "centimeters" per https://github.com/microbiomedata/nmdc-server/issues/221

I've made the change here, but I'm not sure if this is the only change I need to make.

fixes #338

### DIFF
--- a/nmdc_server/attribute_units.py
+++ b/nmdc_server/attribute_units.py
@@ -2,7 +2,6 @@ from typing import Dict, Optional
 
 from pint import Unit, UnitRegistry
 
-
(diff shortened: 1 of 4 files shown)


jbeezley (approved):
:+1:  Yeah, I guess a depth of 2.5km doesn't make much sense.  :)

PR #467, opened 2021-07-28
Refactor database and test code to remove global state

zachmullen:
- [x] Test that prod ingest task still works


zachmullen:
@subdavis I plan to merge this today, PTAL if you want. This is a refactor that will affect the server, dev ingest, prod ingest, and testing.

PR #471, opened 2021-08-03
Envo tree

zachmullen:
Screen Shot 2021-08-03 at 7 31 27 PM


jeffbaumes:
This looks good to me as a MVP of the feature that will let @subdavis start on the frontend. There will be some iteration, but I'd suggest the bar for getting this first PR merged is low. I don't mind the heavy-but-realistic test, but will also not be running it so @subdavis is free to yell if he would like it shortened.

subdavis:
My only concern is the increase in CI time, which is barely more than it takes the web tests to run, so I'm fine with this change.

Merge at your convenience.

zachmullen:
Regarding the test time, I've placed a pytest mark on the new test to indicate that it's expensive beyond what we would expect of a unit test. That gives us the option to disable it from regular CI, perhaps only running it on merge, or nightly, or only locally.

> This looks good to me as a MVP

I'll go ahead and merge this as a proof of concept, but I will accentuate the "M" in MVP. I suspect having trees with hundreds of nodes might not be what we want -- we may wish to add an additional preprocessing step that pares down nodes of the tree that are not present in the biosample dataset.

Issue #474, opened 2021-08-04
Envo Tree API improvements


subdavis:
It would be good to:

* eliminate nodes that aren't present in the biosamples
* indicate in the response what tree belongs to what facet (I can't tell currently how to pair trees in the list with the existing menus)

zachmullen:
IIUC, the mapping from each root to the facet is hard-coded based on what @jeffbaumes showed.

* env_broad_scale_id -> ecosystem (ENVO:01000254)
* env_local_scale_id -> geographic feature (ENVO:00000000)
* env_medium_id -> environmental material (ENVO:00010483)

We could put it in the response, but it might make more sense to re-encode that mapping in the client since it's a fixed law of the domain.

subdavis:
Works for me.

zachmullen:
Actually, on second thought, it probably makes more sense to just change the response from an Array to an Object that maps facet identifier to the root node. I will do that.

subdavis:
Getting this error in dev environment

```
raceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/uvicorn/protocols/http/httptools_impl.py", line 396, in run_asgi
    result = await app(self.scope, self.receive, self.send)
  File "/usr/local/lib/python3.8/site-packages/uvicorn/middleware/proxy_headers.py", line 45, in __call__
    return await self.app(scope, receive, send)
  File "/usr/local/lib/python3.8/site-packages/fastapi/applications.py", line 208, in __call__
    await super().__call__(scope, receive, send)
  File "/usr/local/lib/python3.8/site-packages/starlette/applications.py", line 112, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/usr/local/lib/python3.8/site-packages/starlette/middleware/errors.py", line 181, in __call__
    raise exc from None
  File "/usr/local/lib/python3.8/site-packages/starlette/middleware/errors.py", line 159, in __call__
    await self.app(scope, receive, _send)
  File "/usr/local/lib/python3.8/site-packages/starlette/middleware/sessions.py", line 75, in __call__
    await self.app(scope, receive, send_wrapper)
  File "/usr/local/lib/python3.8/site-packages/sentry_sdk/integrations/asgi.py", line 106, in _run_asgi3
    return await self._run_app(scope, lambda: self.app(scope, receive, send))
  File "/usr/local/lib/python3.8/site-packages/sentry_sdk/integrations/asgi.py", line 152, in _run_app
[...]

PR #475, opened 2021-08-04
Improve response format of envo tree

zachmullen:
Refs #474


nmdc_server/schemas.py
```diff
@@ -597,4 +597,4 @@ class EnvoTreeNode(BaseModel):
 
 
 class EnvoTreeResponse(BaseModel):
-    trees: List[EnvoTreeNode]
+    trees: Dict[str, EnvoTreeNode]
```
zachmullen, line 600: @subdavis

PR #477, opened 2021-08-05
ENVO Tree UI

subdavis:
I haven't integrated the counts back in yet, I'll do this once the pruning happens.

@zachmullen please see the python changes here.  I did this as a workaround to avoid a multi-hour ingest

```
nmdc-server truncate envo_ancestor
nmdc-server ingest --cheap-only
```

Screenshot from 2021-08-06 11-37-22

### DIFF
--- a/nmdc_server/cli.py
+++ b/nmdc_server/cli.py
@@ -1,6 +1,8 @@
 import logging
+from typing import List
 
 import click
+from sqlalchemy.sql import text as sa_text
 
 from nmdc_server import jobs
 from nmdc_server.config import Settings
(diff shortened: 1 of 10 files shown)


zachmullen:
Let's remove the python changes from this branch.

subdavis:
> Let's remove the python changes from this branch.

I need these changes.  In general, I think we need to be able to run some kind of fast-path ingest that skips all the heavy stuff for development.  Doing ingests every time the schema changes is unsustainable for development.

I'll move them to another branch.

zachmullen:
I like the idea of making scripts for common tasks, but there are so many branches in the ingest all that I'm not comfortable trying to maintain direct CLI support. I want to promote a culture of using the shell or one-off scripts for more fine-tuned tasks, so that it won't introduce a maintenance burden.

PR #478, opened 2021-08-06
Infer per-facet envo tree roots at ingest time

zachmullen:
The tree endpoint will now return an array of valid roots for each envo facet.


nmdc_server/schemas.py
```diff
@@ -597,4 +597,4 @@ class EnvoTreeNode(BaseModel):
 
 
 class EnvoTreeResponse(BaseModel):
-    trees: Dict[str, EnvoTreeNode]
+    trees: Dict[str, List[EnvoTreeNode]]
```
zachmullen, line 600: @subdavis

zachmullen:
The test is failing because we now need biosamples loaded for our tree to be built successfully. I'm going to scrap the test and plan to create one with a small fake dataset.

PR #480, opened 2021-08-06
Change celery reference

subdavis:
Fixes bug introduced in #478 `celery.py -> celery_config.py` rename broke container build.


zachmullen:
Sorry about this!

The reason for the change is that I ran into an issue where my interpreter tried to do a relative import of celery rather than an absolute one, which was happening when I tried to import `get_logger`.

Thanks for the fix.

zachmullen (approved):

PR #497, opened 2021-08-30
Prune internal envo tree nodes statically

zachmullen:
I believe this is now equivalent functionality to the algorithm @jeffbaumes laid out. The only thing it doesn't do is delete root nodes, which could be done as a separate step if we want the ability to remove one extra node. (I had originally tried to do it all at once, but the function became complicated and difficult to reason about.)

I would love for someone to come up with a proof about the need to detect changes and run multiple passes. I tried reasoning it out but my brain broke and I gave up so I could get this merged since it gets us 99% to the goal.

### DIFF
--- a/nmdc_server/ingest/envo.py
+++ b/nmdc_server/ingest/envo.py
@@ -174,6 +174,9 @@ def _prune_useless_nodes(
     """
     Remove useless internal (non-root) nodes from a tree.
 
+    A useless node is one that is not in the terms present in the biosample set, and that has
+    only a single child node.
+
     This modifies the structure in-place, returning the new root. Returns whether or not
     any modifications were made to the tree.
     """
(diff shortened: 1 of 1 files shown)


nmdc_server/ingest/envo.py
```diff
@@ -165,6 +165,32 @@ def _nested_envo_subtree(
...
 
 
+def _prune_useless_nodes(
+    node: EnvoTreeNode,
+    present_terms: Set[str],
+    parent: Optional[EnvoTreeNode] = None,
+    modified: bool = False,
+) -> bool:
+    """
+    Remove useless internal (non-root) nodes from a tree.
```
subdavis, line 175: Could you provide a definition of "useless" in this docstring?

zachmullen: I'll give my best attempt, but I don't want to sound too authoritative with our working definition because I just don't feel terribly confident in the underlying math.

zachmullen:
FWIW, with our current dataset, this cut the size of the tree roughly in half.

PR #507, opened 2021-09-10
Protobuf query history

subdavis:
This is something I've wanted to try for a while.   B64 encoded protobuf for fairly dense compression of the query state into the query string.

# Large example 
```
http://localhost:8081/?q=Ci4IARAGGAQiJlszNi4xMDI1OTUzMjg3ODM4MzUsNDIuMjI4NzE3OTUxNTYzMzJdCjAIARAHGAQiKFstMTEzLjcxMjA0MzMxNDgzMDU3LC05My4yMzM1Mjc2ODk4MzA1NV0KKQgAEAIYAiIhIk9yZ2FuaWMgTWF0dGVyIENoYXJhY3Rlcml6YXRpb24i
```
encodes
```
[{"op":"between","field":"latitude","table":"biosample","value":[36.102595328783835,42.22871795156332]},
{"op":"between","field":"longitude","table":"biosample","value":[-113.71204331483057,-93.23352768983055]},
{"op":"==","field":"omics_type","table":"omics_processing","value":"Organic Matter Characterization"}]
```

# Small example

```
http://localhost:8081/?q=ChUIABAPGAMiDSJFb2luIEJyb2RpZSI=
```
encodes
```
[{"op":"==","field":"principal_investigator_name","value":"Eoin Brodie","table":"study"}]
```

@zachmullen @jeffbaumes does this make sense as a way to preserve state between login and to allow people to share query links?


zachmullen:
This is kind of a lot of engineering, I think it would only make sense if it's solving a real issue. What is the distribution of sizes of these query strings? Are we running into length limits?

zachmullen:
Sorry, now I am noticing that the motivation is for sharing links. I think this is a reasonable approach, since it's just confined to the front-end.

subdavis:
This state also needs to be passed through oauth, and b64 strings are significantly cleaner than any other string encoding for structured data of this kind I could come up with.

zachmullen:
In theory we could b64 encode the regular old JSON, but maybe that's going to create really long strings.

web/src/data/protobuf-descriptor.json
```diff
@@ -0,0 +1,88 @@
...
+                "ecosystem_path_id": 14,
+                "principal_investigator_name": 15,
+                "gold_classification": 16,
+                "ecosystem": 17,
+                "ecosystem_category": 18,
+                "ecosystem_type": 19,
+                "ecosystem_subtype": 20,
+                "specific_ecosystem": 21,
+                "env_broad_scale": 22,
+                ",env_local_scale": 23,
```
jeffbaumes, line 79: Is this comma supposed to be here?

subdavis: No, and that's kinda important not to mess up.

jeffbaumes:
I like this approach.

PR #512, opened 2021-09-14
Envo bug

### LINKED ISSUE #510: Searching by ENVO term seems to reset query
To reproduce:

* Filter by a non-ENVO facet.
* Filter by an ENVO facet. The filter will be replaced, not extended (i.e. all other purple chips go away and are replaced by the ENVO filter.

### LINKED ISSUE #511: For bulk download, display number of files and download size when not logged in
It would be nice to still display the number of files and total size even when not logged in.

image

subdavis:
fixes #510
fixes #511


web/src/components/FilterTree.vue
```diff
@@ -63,11 +63,12 @@ export default defineComponent({
       return resp;
     });
 
-    const { loading, request } = useReqest();
+    const { loading, request } = useRequest();
     request(getTreeData);
 
     async function setSelected(values: string[]) {
       const c = cloneDeep(otherConditions.value);
+      console.log('🚀 ~ file: FilterTree.vue ~ line 71 ~ setSelected ~ c', c);
```
zachmullen, line 71: Intentional?

subdavis: It's why we have CI :)

Thanks.

PR #516, opened 2021-09-14
Include definitions from schema

### LINKED ISSUE #505: Facet tooltips
Pull the description of fields from the schema and display them in the UI. A simple approach would be to just display on hover like the following:

image

subdavis:
Screenshot from 2021-09-14 14-11-27

# Notes for reviewer

I found this approach simpler than a backend ingest.

* Install https://github.com/microbiomedata/nmdc-schema as a git submodule
* use `postinstall` to install this submodule on `yarn install`
* Bundle the schema JSON file using webpack

This comes with the added benefit that typescript can enforce the mappings declared in `encoding.ts` because of `resolveJsonModule`.

See how, in the screenshot below, you can get a nice literal type by introspecting the values in the json object.  So if a name changes in a future version of the NMDC schema, it will actually break the build.
Screenshot from 2021-09-14 13-44-05

This is probably less code than going the ingest route

fixes #505

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -18,6 +18,8 @@ jobs:
       -
         name: Checkout code
         uses: actions/checkout@v2
+        with:
+          submodules: 'recursive'
       -
         name: Login to GitHub Container Registry
         uses: docker/login-action@v1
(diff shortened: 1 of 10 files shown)


jeffbaumes:
This approach sounds like a great one. People will need to remember to do the git submodule init, but I see you've already added it to the docs.

Issue #362, opened 2021-05-05
Portal Search - KEGG


ssarrafan:
Enable search by modules and the 3 Pathway Levels. KO term is very granular. Oftentimes we're interested in understanding whether there's presence of an entire pathway (sometimes this may be specific, like "fatty acid metabolism" which is a level 3 pathway, but other times we may be interested in "metabolism" in general which is a level 1 pathway)

Priority - High
Urgency - Medium

kfagnan:
In today's meeting Jeff indicated Jon has the information needed to work on this issue. So we'll leave in this sprint for now.

ssarrafan:
This issue is dependent on the fix to be done on another issue - https://github.com/microbiomedata/nmdc-metadata/issues/342 per @pvangay

ssarrafan:
@zachmullen @subdavis I'm just checking to see if you have everything you need for this issue.  Would you be able to get this completed and available in the portal by the end of this September?  Please let me know if you need anything from anyone on the team or if you're blocked by anything.

zachmullen:
Hi @ssarrafan , I am not aware of any blockers at this time but I'm in a preliminary stage on this. I will raise any issues if I encounter them.

ssarrafan:
Moving to October's sprint since this is being tested now and part of the October release.

ssarrafan:
Released yesterday to production. Closing this issue.

PR #518, opened 2021-09-15
Allow resolving KEGG pathways and modules to KO term sets

zachmullen:
Refs #362 

This provides backend support that will enable parts of KEGG advanced search functionality.


subdavis:
Couple questions:

* Isn't it usually better to have a single migration per pull request?  Why break it up into 3?
* Could you provide a bit of information about how to use this?

zachmullen:
> Isn't it usually better to have a single migration per pull request? Why break it up into 3?

I think that's a philosophical question. I used to hold the opinion that we should be minimalist about number of migrations, but I've kind of lost that mindset and now I don't really care. Reverting them is a pain so I now just treat them as immutable in dev and ask, "what's the worst that could happen"?

zachmullen:
> Could you provide a bit of information about how to use this?

This exposes two new REST endpoints. They can be seen in swagger under the `kegg` tag. One accepts a module identifier as a parameter and returns a list of all KO terms under that module. The other accepts a pathway identifier as a parameter and returns a list of all KO terms under that pathway. This allows us to query by module or pathway in two stages: the first stage resolves to the set of terms, and the second stage filters by terms, which was already supported. This was to avoid engineering complexity on the backend of accepting KO modules or pathways directly in the search query.

PR #520, opened 2021-09-22
Kegg search

subdavis:
@zachmullen I really quite disliked the client-side transformation of KEGG terms, so I did it on the server side.

This implementation isn't perfect, and it should probably be some kind of join, but it's relatively well isolated in the `transformCondition` function IMO.

It also makes the search functions case-insensitive.

This is a basic version where you can paste in the new terms and they'll be transformed on the server.

### DIFF
--- a/nmdc_server/crud.py
+++ b/nmdc_server/crud.py
@@ -201,12 +201,12 @@ def list_omics_processing_data_objects(db: Session, id: str) -> Query:
 
 
 def list_ko_terms_for_module(db: Session, module: str) -> List[str]:
-    q = db.query(models.KoTermToModule.term).filter(models.KoTermToModule.module == module)
+    q = db.query(models.KoTermToModule.term).filter(models.KoTermToModule.module.ilike(module))
     return [row[0] for row in q]
 
 
 def list_ko_terms_for_pathway(db: Session, pathway: str) -> List[str]:
-    q = db.query(models.KoTermToPathway.term).filter(models.KoTermToPathway.pathway == pathway)
+    q = db.query(models.KoTermToPathway.term).filter(models.KoTermToPathway.pathway.ilike(pathway))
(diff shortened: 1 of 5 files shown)


nmdc_server/crud.py
```diff
@@ -201,12 +201,12 @@ def list_omics_processing_data_objects(db: Session, id: str) -> Query:
...
 
 def list_ko_terms_for_module(db: Session, module: str) -> List[str]:
-    q = db.query(models.KoTermToModule.term).filter(models.KoTermToModule.module == module)
+    q = db.query(models.KoTermToModule.term).filter(models.KoTermToModule.module.ilike(module))
     return [row[0] for row in q]
 
 
 def list_ko_terms_for_pathway(db: Session, pathway: str) -> List[str]:
-    q = db.query(models.KoTermToPathway.term).filter(models.KoTermToPathway.pathway == pathway)
+    q = db.query(models.KoTermToPathway.term).filter(models.KoTermToPathway.pathway.ilike(pathway))
```
zachmullen, line 209: These two are only returning the "term" column, is it OK not to be able to distinguish which resulting term corresponds to which module?

Is this a real use case, where someone knows some module / pathway number, but only wants to use a prefix of it?

zachmullen: Sorry I just realized you weren't using `%` here... this is just for case insensitivity then?

subdavis: I don't understand the question.  This is to fix no results returned when I search for `M00000` instead of `m00000`.  There are no wildcards in the query, so I don't' understand the prefix question.

zachmullen:
This is awesome, thanks for digging into it. I like the design and the implementation is fairly straightforward.

zachmullen (approved):

PR #521, opened 2021-09-23
Add database connection pool tuning parameters

zachmullen:
These can be set via the environment if we need to dynamically tune them.

### DIFF
--- a/nmdc_server/config.py
+++ b/nmdc_server/config.py
@@ -15,6 +15,10 @@ class Settings(BaseSettings):
     ingest_database_uri: str = "postgresql:///nmdc_testing"
     testing_database_uri: str = "postgresql:///nmdc_testing"
 
+    # database tuning knobs
+    db_pool_size: int = 20
+    db_pool_max_overflow: int = 40
+
     # for orcid oauth
     secret_key: str = "secret"
     client_id: str = "oauth client id"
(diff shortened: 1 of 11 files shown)


nmdc_server/config.py
```diff
@@ -15,6 +15,10 @@ class Settings(BaseSettings):
     ingest_database_uri: str = "postgresql:///nmdc_testing"
     testing_database_uri: str = "postgresql:///nmdc_testing"
 
+    # database tuning knobs
+    db_pool_size: int = 20
```
zachmullen, line 19: These numbers are still pretty conservative for the hardware we're running on. Once this is deployed I'll mark the sentry errors as resolved so we'll be notified if they crop up again.

nmdc_server/jobs.py
```diff
@@ -20,25 +19,34 @@ def ping():
...
         alembic_cfg.attributes["configure_logger"] = True
 
-        context = MigrationContext.configure(db.connection())
-        head = context.get_current_revision()
-        if head is None:
-            metadata.create_all(engine)
-            command.stamp(alembic_cfg, "head")
-        else:
-            command.upgrade(alembic_cfg, "head")
+        with db.connection() as conn:
```
zachmullen, line 42: This change has a reasonable probability of fixing the connection leaking issue. `db.connection()` is a context manager that is supposed to be closed at the end.

zachmullen:
I've tested this locally and everything seems to work. Given that dev deployment is pretty broken currently, it seems low risk to merge this.

PR #522, opened 2021-09-23
Remove ingest db

subdavis:
This service is unused and was confusing me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,13 +93,13 @@ The fix is to drop the existing database and to restart the service.  In docker-
 can be done by running
 
 ```bash
-docker-compose run backend psql postgres -c 'drop database nmdc;'
+docker-compose run backend psql postgres -c 'drop database nmdc_a;'
 ```
 
 On Spin, you can use the web interface to start a shell on the `db` instance and run:
 
 ```bash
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -93,13 +93,13 @@ The fix is to drop the existing database and to restart the service.  In docker-
 can be done by running
 
 ```bash
-docker-compose run backend psql postgres -c 'drop database nmdc;'
+docker-compose run backend psql postgres -c 'drop database nmdc_a;'
```
zachmullen, line 96: Locally, I've had to also set my alembic.ini database uri to `nmdc_a` rather than `nmdc` -- should we also make that change in the repo?

subdavis: Yes.

zachmullen (approved):

nmdc_server/alembic.ini
```diff
@@ -37,7 +37,7 @@ script_location = nmdc_server/migrations
 
 # TODO: When running alembic from the CLI this value must be
 #       hard coded.
-sqlalchemy.url = postgresql:///nmdc
+sqlalchemy.url = postgresql:///nmdc_a
```
zachmullen, line 40: This is already in main FWIW

PR #523, opened 2021-09-23
KEGG text search

subdavis:
Include exact matches for the kegg term.


nmdc_server/crud.py
```diff
@@ -213,7 +213,7 @@ def list_ko_terms_for_pathway(db: Session, pathway: str) -> List[str]:
 def kegg_text_search(db: Session, query: str, limit: int) -> List[models.KoTermText]:
     q = (
         db.query(models.KoTermText)
-        .filter(models.KoTermText.text.ilike(f"%{query}%"))
+        .filter(models.KoTermText.text.ilike(f"%{query}%") | models.KoTermText.term.ilike(query))
```
zachmullen, line 216: The `%` signs match anything, including the empty string, so wouldn't the full terms already be captured?

PR #525, opened 2021-09-24
deconflict

subdavis:
@zachmullen I reverted your `SessionLocal` changes because I want to be able to change the ingest destination so that I can test ingests and blue/green locally

### DIFF
--- a/nmdc_server/cli.py
+++ b/nmdc_server/cli.py
@@ -6,7 +6,7 @@ import click
 
 from nmdc_server import jobs
 from nmdc_server.config import Settings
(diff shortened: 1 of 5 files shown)


nmdc_server/cli.py
```diff
@@ -24,13 +24,17 @@ def cli(ctx):
 @cli.command()
 def migrate():
     """Upgrade the database schema."""
-    jobs.migrate()
+    jobs.migrate(obj["settings"].ingest_database_uri)
```
zachmullen, line 27: Can you rebase, I fixed this line (it was actually a bug).

PR #526, opened 2021-09-24
Remove regex timeout

subdavis:


nmdc_server/ingest/pipeline.py
```diff
@@ -26,29 +26,25 @@ def __call__(self, db: Session, obj: Dict[str, Any], **kwargs: Any) -> LoadObjec
...
     db.add(row)
     db.flush()
 
     annotations: Collection = kwargs["annotations"]
 
-    query = annotations.find(
-        {
-            "was_generated_by": pipeline.id,
-            "has_function": {
-                "$regex": ko_regex,
```
zachmullen, line 40: I'm curious; now that the `was_generated_by` is indexed, does keeping the regex in the query slow it down materially? Right now we know it's not filtering many rows, but is there potential for it to be a stronger filter in the future?

nmdc_server/ingest/pipeline.py
```diff
@@ -26,29 +26,25 @@ def __call__(self, db: Session, obj: Dict[str, Any], **kwargs: Any) -> LoadObjec
...
+    query = annotations.find({"was_generated_by": pipeline.id})
     if kwargs.get("function_limit"):
         query = query.limit(kwargs["function_limit"])
 
     gene_functions: Dict[str, models.GeneFunction] = {}
     mga_gene_functions: List[Tuple[UUID, str, str, str]] = []
     for gf in query:
         function_id = gf["has_function"]
+        if not re.match(ko_regex, function_id):
+            logger.warn(f"Unexpected function {function_id} for pipeline {pipeline.id}")
```
zachmullen, line 46: I think it's a good practice for warnings to end up in Sentry. If we don't leave the function check out of the query, let's instead keep a running total of how many are being excluded, and only issue a log warning if it's above a certain threshold. That seems like something we'd want to get an alert about.

zachmullen (changes requested):

nmdc_server/ingest/pipeline.py
subdavis: It's definitely a lot slower (not 4 hours, but 10+ min)

zachmullen: Roger that. Did you try it with the regex but also including `.hint('was_generated_by')`? I think that's janky too, I'm just very curious.

(Don't feel the need to spend any more time on this if you don't want, BTW.)

zachmullen: (The right solution is a compound index anyway.)

nmdc_server/ingest/pipeline.py
subdavis: We don't have sentry in celery, this is currently a few hundred messages, and I think the individual values are valuable for knowing what's wrong.  Can we make a second pass for better handling and reporting later?

zachmullen: Fine by me. I'll file a ticket; we should definitely have sentry setup for celery.

zachmullen (approved):

nmdc_server/ingest/pipeline.py
subdavis: I'm trying the hint.  I'd prefer that to the client side filter.

subdavis: No dice, merging.

PR #527, opened 2021-09-24
Fix CLI commands to use ingest database

zachmullen:


subdavis (changes requested):
Need to change the ingest URI in docker env

zachmullen:
> Need to change the ingest URI in docker env

Do you want them to be the same? [Currently we use two different databases](https://github.com/microbiomedata/nmdc-server/blob/main/.docker-env)

subdavis:
If it isn't changed, then the documentation is wrong because after you do a CLI ingest, you'll start the server and it will have no data because ingest and server are pointed at different DBs.

For local development, I think they should default to the same value so that an ingest happens in `nmdc_a`.

subdavis (approved):

PR #530, opened 2021-09-27
Optimize MGA ingest

zachmullen:
This is my attempt to find all the "low-hanging fruit" optimizations within the MGA ingest inner loop. 

* Use a projection operator to minimize network data transfer from Mongo
* Use `bulk_save_objects` rather than the dialect `insert`; at least on my local docker-compose environment, this was about 3x faster.
* Instead of doing 1-2 queries per gene function being added (due to `get_or_create`), gather them all up and use an insert with `ON CONFLICT DO NOTHING` behavior. This will massively increase the performance of ingesting the gene functions, but that was admittedly not the hot path here.

### DIFF
--- a/nmdc_server/ingest/pipeline.py
+++ b/nmdc_server/ingest/pipeline.py
@@ -1,6 +1,5 @@
 import re
-from typing import Any, Dict, List, Tuple, cast
-from uuid import UUID, uuid4
+from typing import Any, Dict, List, Set, cast
 
 from pymongo.collection import Collection
 from pymongo.cursor import Cursor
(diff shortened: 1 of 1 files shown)


subdavis:
I did what you said Friday and made these changes too, the only difference being I gathered the new objects into a set and don't pass them to bulk ingest if they have been added already.  It may or may not be faster.


The real cost here comes from the MGA Gene Functions where there are usually only dozens or hundreds of functions but tens of thousands of duplicates with different subjects.

Since I have no idea what a subject is and users don't need to query for it, I don't see the point of having it in our DB at all.  I think we might be able to save millions of records of ingest by denormalizing whatever join requires the subject to just use the function ID instead.

subdavis:
FWIW, this appears to be the entire purpose of https://github.com/microbiomedata/nmdc-server/blob/main/nmdc_server/models.py#L719, but that happens after ingest.

It doesn't make a lot of sense to me for us to ingest that magnitude of data just to reduce the relationship to `gene_function_id`, so perhaps there's an argument to be made that this denationalization belongs in Mongo.

subdavis (approved):

zachmullen:
I'll merge this and kick off an ingest on dev; I really want to see how long it takes up there vs my laptop.

PR #534, opened 2021-09-28
More database tuning

zachmullen:
I've already decreased the worker count on prod. My hypothesis is that that (along with this change) will fix the connection limit errors we've seen in Sentry. I will resolve them once I merge.


subdavis:
Cool!

PR #561, opened 2021-12-06
Fix ingest download stats

### LINKED ISSUE #551: Portal download stats persistence bug
When re-ingest happens, a bug causes download stats to be lost.  We don't want to lose stats (at least in prod), so this is blocking reingests in production right now.

subdavis:
* Fixes ingest where data objects disappear and cause a foreign key issue on the download stats tables.
* Adds "fast ingest" as an option through the swagger page.

fixes #551

### DIFF
--- a/.docker-env
+++ b/.docker-env
@@ -7,4 +7,4 @@ PGDATABASE=nmdc_a
 NMDC_DATABASE_URI="postgresql://postgres:postgres@db/nmdc_a"
 NMDC_CELERY_BACKEND="redis://redis:6379/0"
 NMDC_CELERY_BROKER="redis://redis:6379/0"
-NMDC_INGEST_DATABASE_URI="postgresql://postgres:postgres@db/nmdc_a"
+NMDC_INGEST_DATABASE_URI="postgresql://postgres:postgres@db/nmdc_b"
--- a/.docker-env
+++ b/.docker-env
@@ -7,4 +7,4 @@ PGDATABASE=nmdc_a
 NMDC_DATABASE_URI="postgresql://postgres:postgres@db/nmdc_a"
 NMDC_CELERY_BACKEND="redis://redis:6379/0"
 NMDC_CELERY_BROKER="redis://redis:6379/0"
-NMDC_INGEST_DATABASE_URI="postgresql://postgres:postgres@db/nmdc_b"
+NMDC_INGEST_DATABASE_URI="postgresql://postgres:postgres@db/nmdc_a"
--- a/nmdc_server/api.py
+++ b/nmdc_server/api.py
@@ -383,14 +383,19 @@ async def ping_celery(token: Token = Depends(admin_required)) -> bool:
     tags=["jobs"],
     responses=login_required_responses,
 )
-async def run_ingest(token: Token = Depends(admin_required), db: Session = Depends(get_db)):
+async def run_ingest(
+    token: Token = Depends(admin_required),
+    db: Session = Depends(get_db),
+    function_limit: int = None,
+    skip_annotation: bool = False,
+):
     lock = db.query(IngestLock).first()
     if lock:
         raise HTTPException(
             status_code=409,
             detail=f"An ingest started at {lock.started} is already in progress",
         )
-    jobs.ingest.delay()
+    jobs.ingest.delay(function_limit=function_limit, skip_annotation=skip_annotation)
     return ""
 
 
--- a/nmdc_server/api.py
+++ b/nmdc_server/api.py
@@ -385,9 +385,8 @@ async def ping_celery(token: Token = Depends(admin_required)) -> bool:
 )
 async def run_ingest(
     token: Token = Depends(admin_required),
+    params: schemas.IngestArgumentSchema = schemas.IngestArgumentSchema(),
     db: Session = Depends(get_db),
-    function_limit: int = None,
-    skip_annotation: bool = False,
 ):
     lock = db.query(IngestLock).first()
     if lock:
@@ -395,7 +394,7 @@ async def run_ingest(
             status_code=409,
             detail=f"An ingest started at {lock.started} is already in progress",
         )
-    jobs.ingest.delay(function_limit=function_limit, skip_annotation=skip_annotation)
+    jobs.ingest.delay(function_limit=params.function_limit, skip_annotation=params.skip_annotation)
     return ""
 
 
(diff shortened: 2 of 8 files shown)


nmdc_server/ingest/common.py
```diff
@@ -47,3 +50,14 @@ def extract_extras(
         if key not in fields and key not in exclude:
             values["annotations"][key] = extract_value(value)
     return values
+
+
+def merge_or_delete(ingest_db: Session, query):
+    logger = logging.getLogger()
```
zachmullen, line 56: Rather than use the root logger, let's move `logger` up to module scope and do `logger = logging.getLogger(__name__)`. (That's the typical practice in Python.)

nmdc_server/api.py
```diff
@@ -383,14 +383,19 @@ async def ping_celery(token: Token = Depends(admin_required)) -> bool:
     tags=["jobs"],
     responses=login_required_responses,
 )
-async def run_ingest(token: Token = Depends(admin_required), db: Session = Depends(get_db)):
+async def run_ingest(
+    token: Token = Depends(admin_required),
+    db: Session = Depends(get_db),
+    function_limit: int = None,
+    skip_annotation: bool = False,
```
zachmullen, line 390: For consistency, let's use a JSON-encoded body for POST request params. (This will mean adding a new schema in the `schemas` module and setting `response_model=...` in the decorator above.)

.docker-env
```diff
@@ -7,4 +7,4 @@ PGDATABASE=nmdc_a
 NMDC_DATABASE_URI="postgresql://postgres:postgres@db/nmdc_a"
 NMDC_CELERY_BACKEND="redis://redis:6379/0"
 NMDC_CELERY_BROKER="redis://redis:6379/0"
-NMDC_INGEST_DATABASE_URI="postgresql://postgres:postgres@db/nmdc_a"
+NMDC_INGEST_DATABASE_URI="postgresql://postgres:postgres@db/nmdc_b"
```
zachmullen, line 10: Is this change intentional?

nmdc_server/config.py
```diff
@@ -36,7 +36,8 @@ class Settings(BaseSettings):
     host: Optional[str] = None  # sets the host name for the oauth2 redirect
 
     # mongo database used for ingest
-    mongo_host: str = "nmdc-metadata.polyneme.xyz"
+    mongo_host: str = "mongo-loadbalancer.nmdc-runtime-dev.development.svc.spin.nersc.org"
+    mongo_port: int = 27017
```
zachmullen, line 40: Do any of our environments use a nonstandard mongo port? This is a good chance, I was just curious.

Another option we might want to consider for the future is to just accept a URI connection string that can encode lots of different options, rather than enumerating the options explicitly.

nmdc_server/ingest/common.py
```diff
@@ -47,3 +50,14 @@ def extract_extras(
...
+    for row in query:
+        try:
+            ingest_db.merge(row)
+            ingest_db.commit()
+        except IntegrityError:
+            logger.error("Error: data object with download history was removed.")
```
zachmullen, line 62: `merge_or_delete` seems like a pretty generic function, but this error message is quite specific to the case of download history. Maybe we just need to rename this function if it's really just designed for this specific case?

zachmullen (changes requested):
I'd be OK merging this, but I'd like to investigate whether we can achieve this with a bulk operation, which would likely be significantly faster. [This SO post](https://stackoverflow.com/questions/36038751/can-insert-on-conflict-be-used-for-foreign-key-violations) shows a clever approach using a join to skip rows that would cause a FK integrity violation. (ON CONFLICT can't be used for FK constraints, sadly.)

nmdc_server/config.py
subdavis: Sorta.  You have to do SSH tunneling now to run this locally.

And that command doesn't actually work, I had to modify it a bit, which I'll document elsehwhere.

https://github.com/microbiomedata/nmdc-runtime/blob/44fbb11f69b51866a410a463f1f5678a2391cd8b/Makefile#L65

nmdc_server/ingest/common.py
subdavis: `logging.getLogger()` is used in 7 places. This change seems out of scope for this bugfix, I'll open a different PR.

nmdc_server/api.py
subdavis: This is the only function intended to be used by non-Kitware folks through the swagger API, so using query params means that there's UI to enter just a number and "true/false" dropdown.  JSON api will make this harder to use.

I'll change it if you think it's important, but I think this way is more user-friendly.

nmdc_server/schemas.py
```diff
@@ -632,3 +632,8 @@ class Config:
 
 class KeggTermTextListResponse(BaseModel):
     terms: List[KeggTermText]
+
+
+class IngestArgumentSchema(BaseModel):
+    skip_annotation: bool = False
+    function_limit: int = 0  # Default to no limit
```
zachmullen, line 639: Is passing `0` the same as passing `None`?

nmdc_server/jobs.py
```diff
@@ -67,23 +68,18 @@ def ingest():
...
-                    ingest_db.merge(row)
-                ingest_db.commit()
-            except Exception:
-                ingest_db.rollback()
-                raise
+            ingest_db.execute("select truncate_tables()").all()
+
+            # ingest data
+            logger.info(
+                f"Load with function_limit={function_limit}, skip_annotaiton={skip_annotation}"
```
zachmullen, line 75:
```suggestion
                f"Load with function_limit={function_limit}, skip_annotation={skip_annotation}"
```

zachmullen (changes requested):

nmdc_server/schemas.py
subdavis: They result in the same behavior, if that's what you mean.

`Optiona[int]` would hide the option from the generated example JSON, which makes it pretty much undiscoverable

nmdc_server/ingest/common.py
```diff
@@ -47,3 +50,14 @@ def extract_extras(
...
+
+
+def maybe_merge_download_artifact(ingest_db: Session, query):
+    logger = logging.getLogger()
+    for row in query:
+        try:
+            ingest_db.merge(row)
+            ingest_db.commit()
+        except IntegrityError:
+            logger.error("Error: data object with download history was removed.")
```
zachmullen, line 62:
```suggestion
            logger.info("Error: data object with download history was removed.")
```
For the sake of our alerting (to the extent that it exists), let's make this a `logger.info`. Upstream deletion shouldn't cause error alerts from our app.

zachmullen (changes requested):
LGTM after you accept the one suggestion

zachmullen (approved):

PR #564, opened 2021-12-07
Remove NMDC_MONGO_HOST from instructions

zachmullen:
The default value will be used in development.

@subdavis I think this is no longer needed, please correct me if that's wrong.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,6 @@ level directory containing mongo credentials.
 
 ```bash
 # .env
-NMDC_MONGO_HOST=changeme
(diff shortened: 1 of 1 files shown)


subdavis:
The default value is incorrect if you use SSH tunneling (which you have no choice but to use right now).  You have to use either `localhos` or the docker bridge gateway or your DHCP-assiged IP if you're in a container.

zachmullen:
Oh right. I will get it working locally and update these instructions with the required steps. Thanks.

PR #565, opened 2021-12-07
Client/december minor updates

### LINKED ISSUE #547: NDMC logo link update
Update upper-left NMDC logo link on data.microbiomedata.org to data.microbiomedata.org (instead of microbiomedata.org)

### LINKED ISSUE #548: EnvO triad update on portal
Update EnvO triad on the portal search (biome, feature, material) to match latest version of MIxS (and the NMDC schema): broad-scale environmental context, local environmental context, environmental medium

### LINKED ISSUE #552: Explicit units in data object
Make sure that units are explicit in the data object (i.e. Depth search is in cm, but this field is stored in meters so the search doesn't work correctly). This may be related to [the depth range issue.](https://github.com/microbiomedata/nmdc-metadata/issues/355)

From #448

### LINKED ISSUE #562: Don't use root logger everywhere
Prefer the normal python convention of module-scoped loggers using `__name__`.

subdavis:
* Switch depth to meter.  the deepest sample (2000) is actually fracking fluid, so that's probably right, but we should check.  Almost all others are <1 meter in depth.  There are some "microbial soil" samples with a depth of 15 which seems more likely to be CM than meters.   `igsn:IEWFS000B`

Fixes #562 
Fixes #547 
Fixes #552 
Fixes #548


zachmullen:
Looks like you may have forgotten to check in the new logger file

zachmullen (approved):

PR #566, opened 2021-12-07
Add docs for tunneling mongodb in development

zachmullen:
@subdavis would you mind testing this out for me? Based on [this](https://stackoverflow.com/questions/31324981/how-to-access-host-port-from-docker-container), we may need to add 

```yaml
extra_hosts:
  - "host.docker.internal:host-gateway"
```

to docker-compose config, but only on Linux. It works without it on MacOS.

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -5,6 +5,10 @@ services:
       - "./nmdc_server:/app/nmdc_server"
     command: /start-reload.sh
 
+  db:
+    environment:
+      - POSTGRES_DB=nmdc_a
+
   worker:
     volumes:
       - "./nmdc_server:/app/nmdc_server"
--- a/.docker-env
+++ b/.docker-env
@@ -8,3 +8,6 @@ NMDC_DATABASE_URI="postgresql://postgres:postgres@db/nmdc_a"
 NMDC_CELERY_BACKEND="redis://redis:6379/0"
 NMDC_CELERY_BROKER="redis://redis:6379/0"
 NMDC_INGEST_DATABASE_URI="postgresql://postgres:postgres@db/nmdc_a"
+# See docs/mongo_ssh_tunnel.md for how to access Mongo in development
+NMDC_MONGO_HOST=host.docker.internal
+NMDC_MONGO_PORT=27027
--- a/.github/docs/mongo_ssh_tunnel.md
+++ b/.github/docs/mongo_ssh_tunnel.md
@@ -1,0 +1,7 @@
+1. If you haven't already, [set up MFA on your NERSC account](https://docs.nersc.gov/connect/mfa/)
+   (it's required for SSHing in).
+1. Run the following SSH tunneling command, substituting your actual NERSC username:
+
+   ```ssh -L 127.0.0.1:27027:mongo-loadbalancer.nmdc-runtime-dev.development.svc.spin.nersc.org:27017 <YOUR_NERSC_USERNAME>@dtn01.nersc.gov '/bin/bash -c "while [[ 1 ]]; do echo heartbeat; sleep 300; done"'```
+
+1. When prompted, provide your NERSC password concatenated with your MFA OTP code.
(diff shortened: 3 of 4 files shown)


nmdc_server/config.py
```diff
@@ -35,9 +35,9 @@ class Settings(BaseSettings):
     oauth_token_endpoint: str = "https://orcid.org/oauth/token"
     host: Optional[str] = None  # sets the host name for the oauth2 redirect
 
-    # mongo database used for ingest
-    mongo_host: str = "mongo-loadbalancer.nmdc-runtime-dev.development.svc.spin.nersc.org"
-    mongo_port: int = 27017
+    # mongo database used for ingest (see README for how to SSH tunnel to the mongo DB)
+    mongo_host: str = "host.docker.internal"
```
zachmullen, line 39: It looks like we aren't setting this via the environment in our dev and prod deployments. Would it be preferable to have the default value here be the deployment value, or keep the default as the docker value and set env vars on the deployments?

subdavis:
I'll test this when I get in, but I [don't expect it will work](https://stackoverflow.com/questions/48546124/what-is-linux-equivalent-of-host-docker-internal/61001152)

docker-compose.dev.yml
```diff
@@ -5,6 +5,10 @@ services:
       - "./nmdc_server:/app/nmdc_server"
     command: /start-reload.sh
 
+  db:
+    environment:
+      - POSTGRES_DB=nmdc_a
```
subdavis, line 10: Is this necessary? It's a departure from the conditions in production (no default database).  Could this cause something to work by accident in local dev that would break in prod?

subdavis (changes requested):
This doesn't work on linux.

```
pymongo.errors.ServerSelectionTimeoutError: host.docker.internal:27027: [Errno -2] Name does not resolve, Timeout: 30s, Topology Description: <TopologyDescription id: 61b117621d856fe0d712d00e, topology_type: Unknown, servers: [<ServerDescription ('host.docker.internal', 27027) server_type: Unknown, rtt: None, error=AutoReconnect('host.docker.internal:27027: [Errno -2] Name does not resolve')>]>
```

There's also no way to know what the bridge IP will be.    You can force it in docker compose like this, and hope that docker hasn't already allocated that range:

``` yaml
networks:
  public:
    driver: bridge
    ipam:
      driver: default
      config:
      - subnet: 172.29.0.0/16
        gateway: 172.29.0.1
```

But you still have a problem that the port isn't bound properly with SSH tunnelling unless you specify the exact IP

```
/app # telnet 172.29.0.1 27027
telnet: can't connect to remote host (172.29.0.1): Connection refused
/app # telnet 10.171.2.216 27027
telnet: can't connect to remote host (10.171.2.216): Connection refused
```

The only cross-platform, static option I can see is this:

```
ssh -L 172.29.0.1:27027:mongo-loadbalancer.nmdc-runtime-dev.development.svc.spin.nersc.org:27017 bdavis@dtn01.nersc.gov '/bin/bash -c "while [[ 1 ]]; do echo heartbeat; sleep 300; done"'

# Now telnet works
```

It's realistically just us, so that's not the worst thing ever, but if anyone has remapped the docker network space to exclude that IP things will break.  This is unlikely.

subdavis:
Weird idea: what if we run the SSH tunnel in a container too?  `ldc run ssh-tunnel-thing`

Then you wouldn't need any of this because the hostname would be static.

docker-compose.dev.yml
zachmullen: The reason I set this is because `nmdc_a` appears to be what's hardcoded everywhere else in the dev environment, but [these lines](https://github.com/microbiomedata/nmdc-server/blob/main/docker-compose.yml#L15-L19) would otherwise cause it to be set to `nmdc`, and I have no idea why.

zachmullen:
Did the `extra_hosts` option not work for you?

> Weird idea: what if we run the SSH tunnel in a container too?

I'm fine with that if it solves the networking issue.

subdavis:
Closing in favor of #569

PR #582, opened 2022-01-18
Switch from Alpine to debian

subdavis:
Alpine and python don't mix well [1] [2]

[1] https://github.com/tiangolo/uvicorn-gunicorn-docker#-alpine-python-warning
[2] https://www.notion.so/A-Comprehensive-Guide-for-Switching-to-Poetry-b452ad4d816a4cdda3726847c74fc21f#f3405cef8fc043508dfcdd2115b14cd4

Because there are no wheels for alpine, builds are taking forever.  Switch to debian base images.  Drop all the manual installation of build requirements and just use wheels.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,8 +28,8 @@ setup(
         "requests",
         "sentry-sdk[celery,sqlalchemy]",
         "sqlalchemy>=1.4",
-        "starlette",
-        "typing-extensions",
+        "starlette==0.17.1",
+        "typing-extensions==4.0.1",
     ],
     include_package_data=True,
     entry_points="""
--- a/setup.py
+++ b/setup.py
@@ -28,8 +28,8 @@ setup(
         "requests",
         "sentry-sdk[celery,sqlalchemy]",
         "sqlalchemy>=1.4",
-        "starlette==0.17.1",
-        "typing-extensions==4.0.1",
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -28,8 +28,8 @@
         "requests",
         "sentry-sdk[celery,sqlalchemy]",
         "sqlalchemy>=1.4",
-        "starlette",
-        "typing-extensions",
+        "starlette==0.17.1",
+        "typing-extensions==4.0.1",
```
zachmullen, line 32: Can we add a comment as to why these need to be pinned?

zachmullen (changes requested):

setup.py
subdavis: Everything is pinned, these were just missed last time.  Comment is here:

https://github.com/microbiomedata/nmdc-server/blob/0c0c8bb42df29ffad19deafa1a2dc19ca5890259/setup.py#L9-L10

subdavis: They don't.  Unpinned in https://github.com/microbiomedata/nmdc-server/pull/582/commits/7da8c6b403ca7a3eda2a885d6ab52188a17002ec

zachmullen (approved):

PR #594, opened 2022-02-03
Submission Portal Revision 3

subdavis:
* Adds endpoint and DB table for submitting to server.
* Credit Association Enum added to Study Contributor Form Element.


web/src/views/SubmissionPortal/Components/ValidateSubmit.vue
```diff
@@ -29,18 +16,19 @@ export default defineComponent({
...
     function downloadJson() {
-      saveAs('nmdc_study.json', payload.value);
+      saveAs('nmdc_study.json', submitPayload.value);
     }
 
-    function submit() {
-      // console.log('whatever');
-    }
+    const { request, loading: submitLoading, count: submitCount } = useRequest();
+    const doSubmit = request(submit);
```
mcovalt, line 23:
```suggestion
    const doSubmit = () => request(submit);
```

The `request` function executes its input function immediately. It returns a promise. As a result, the `submit` function is executed when the component is setup and the submit button is always in a disabled state.

This suggestion changes `doSubmit` **_from_** a promise of `submit`'s execution **_to_** a closure that executes `request`. This will allow the user to execute `doSubmit` by clicking the submit button.

mcovalt (changes requested):
I don't believe this works as intended. I would think the submission can be reviewed at the last step, but the submission is sent automatically. I've suggested a change that I think will achieve the intended behavior.

web/src/views/SubmissionPortal/Components/ValidateSubmit.vue
subdavis: Thank you!  I changed this at the last second after testing with it written a different way and didn't test again.

subdavis: BTW, this pattern comes from https://www.notion.so/Composition-func-adds-loading-and-error-state-to-arbitrary-requests-305472ee7227400986da4e04ae22a52c

mcovalt: Very cool. I like this a lot and will use it in other projects.

mcovalt (approved):

PR #642, opened 2022-04-08
Add Metatranscriptomics Gene Function Join Path

### LINKED ISSUE #618: Fix metaT and metaP annotation search
Currently this search returns no matches on metaT in the bar chart:

image

We need to discover why metaT are not being populated and fix it so that this search returns 33 matches for metaT in the bar chart.

Some thoughts on the problem from @dehays and others:
* There is a single mongo collection storing all annotations which should contain metaT, metaP, metaG annotations. Understand that table and how to distinguish between the annotation types.
* Annotations should link back to omics_processing ids, which should have a definitive type (metaG, metaT, etc.).
* metaT annotations are run separately from metaG, and there should be data parsed and populated in mongo.
* nmdc_server does not seem to have a separate join path/table for metaT. Investigate why that is and how to add it.

subdavis:
fixes #618


subdavis:
Unnecessary right now, #618 only required a data change.  Can revisit as requirements become more clear.

PR #654, opened 2022-04-25
Updates for submission portal:

subdavis:
* Fixes backend admin checking logic.
* Update submission portal template choice logic.
    * If JGI/EMSL, only enable soil.
    * if user has uploaded metadata, disable template form and prompt to clear form.
    * if they have any non-doe enabled, allow all base packages
* Templates aren't configured properly: air and misc should be enabled.
* Adds 4 new administrators.

### DIFF
--- a/nmdc_server/api.py
+++ b/nmdc_server/api.py
@@ -484,19 +484,16 @@ async def download_zip_file(
     responses=login_required_responses,
     response_model=query.MetadataSubmissionResponse,
 )
-def list_submissions(
+async def list_submissions(
     db: Session = Depends(get_db),
     token: Token = Depends(login_required),
     pagination: Pagination = Depends(),
-    orcid: str = None,
 ):
     query = db.query(SubmissionMetadata)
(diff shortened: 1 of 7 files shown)


pvangay:
@subdavis JGI and/or EMSL only data does not seem to limit the packages to soil only. [See if this link works?](https://data.dev.microbiomedata.org/submission/822e290d-6837-4956-abb9-996dd5f6d8b9/templates)

subdavis:
It's definitely possible I've misunderstood what you wanted here.  With the link you sent me, the page looks like this:

Screenshot from 2022-04-25 17-43-05

Everything except soil is disabled, and you get a message at the bottom explaining why. 

If this isn't what you're seeing, try a cache refresh.  If this is what you're seeing, but it isn't the right behavior, could you clarify how this page should look?  Thanks.

pvangay:
oh geez - yes, I needed to refresh. I swore I did that already. really sorry!! @subdavis

PR #667, opened 2022-05-04
Embedded DataHarmonizer

subdavis:
# Summary

This pull request attempts to embed DataHarmonizer in the submission portal by pulling it in as a library rather than using an `iframe`.

* Installs code as `npm` dependency from `sheets_and_friends`.
* Runs necessary scripts on the page by copying them using `CopyWebpackPlugin` to move the scripts from `node_modules` into the production build and including them from our own `index.html` page.
* Defines globals like `DataHarmonizer` and `TEMPLATES` in typescript def file.
* Imports the HTML template file using a webpack raw loader to include the template as a string then mount it into the page as `innerHTML`.
* Finally, basically recreates `DataHarmonizer/linkml.js` in the Submission portal vue route.

## Discussion

To consume DataHarmonizer as a proper library, it would be nice to have it bundled in a way that modern build processes understand.   In order to keep the requirement that someone can just double-click an HTML file to run the DH app, I think you'd have to use rollup to build both an ESM distribution and a UMD distribution and https://requirejs.org/ to load the umd version in a browser without a server but ship the ESM version to NPM.

Also, it would really be nice to not have to load the HTML template into the page like this, and instead be able to pass the DH library a root dom selector where the library is responsible for populating content.


ssarrafan:
@jeffbaumes @subdavis should I add this issue to the May sprint board?

jeffbaumes:
Yes, this is being actively worked on.

PR #687, opened 2022-05-31
Better help in submission portal

### LINKED ISSUE #671: It is non-intuitive that "double clicking" is required to reveal the help for each term
Add help ? icon similar to portal, or hover over.

mcovalt:
Closes #671 
Blocked by #667 

Currently, a modal appears after double clicking a field name in DataHarmonizer to describe that field. This PR builds off of the work being done to embed DataHarmonizer in order to access the help text of that modal.

So far, a method has been added to the `harmonizerApi` to get the help text. Now, we need to either add a link to the Handsontable to display this help text for each field or some other method to determine if the user would like to view the help-text.


web/src/views/SubmissionPortal/HarmonizerView.vue
```diff
@@ -339,6 +342,7 @@ export default defineComponent({
         </v-icon>
         Go to previous step
       </v-btn>
+      <div v-html="helpHtml" />
```
mcovalt, line 345: We could replicate [the HTML building from DataHarmonizer](https://github.com/cidgoh/DataHarmonizer/blob/b7cfae38b97a063bd1baa321314f72d8fbad0781/script/main.js#L555-L564) rather than setting the HTML directly.

Issue #681, opened 2022-05-24
GSP Feedback: Data Portal: Click study name to access study page


ssarrafan:
Some users clicked on the title name, checkbox, then > (last). Could be that the arrow is on the far right of the screen.

User need - Low
Feasibility - Make study title name clickable and takes the user to the study page

subdavis:
Solved by @DeepikaGhodki in #724

PR #724, opened 2022-06-20
Enabled navigation to study page on click

DeepikaGhodki:
Issue #681 

Clicking on study tile in search results will route to the respective study.

### DIFF
--- a/web/src/views/Search/SearchLayout.vue
+++ b/web/src/views/Search/SearchLayout.vue
@@ -172,7 +172,6 @@ export default defineComponent({
               </v-card-title>
               <SearchResults
                 disable-pagination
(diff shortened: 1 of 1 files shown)


subdavis (changes requested):
Should make only the title clickable, not the full list item row.  This change breaks the study checkbox and omics type chip click behavior -- clicking the checkbox should add a filter condition,  not trigger a navigation event.

I actually don't mind making the full list item clickable as long as it doesn't break any of the existing behaviors.  For example, `@click.prevent` on sub-elements may prevent the selected event from bubbling up.

DeepikaGhodki:
Used `@click.stop`  to prevent parent click events from getting fired on sub elements.

Issue #716, opened 2022-06-16
Avoid inadvertent page navigation on horizontal scrolling


jeffbaumes:
This shows up when using mac trackpad.

Seems there is a potential solution described here:

https://stackoverflow.com/questions/50616221/prevent-page-navigation-on-horizontal-scroll

DeepikaGhodki:
Closed in #727

PR #727, opened 2022-06-22
Horizontal scroll

DeepikaGhodki:
Issue #716 

To avoid inadvertent page navigation on horizontal scrolling in data harmonizer, we disable default browser behavior on reaching to the end of scroll.


subdavis (changes requested):
Please add a comment about why this code exists.

subdavis (approved):
Excellent, thanks!

PR #733, opened 2022-06-29
User table

DeepikaGhodki:
### **Changes**

1. Endpoints to list all users and update them (only admin status update supported as now)
2. A new route `/users` that lists all users present in the system
3. This page is only accessible to an admin user
4. Only admin status of a given user can be changed using the toggle switch (in admin column) 

**To-do**

- [x] Add pagination to list all users
- [x] Client implementation of update admin status

**User Page**

Screenshot from 2022-06-29 14-18-37

### DIFF
--- a/nmdc_server/api.py
+++ b/nmdc_server/api.py
@@ -555,14 +555,15 @@ async def submit_metadata(
 
 
 @router.get(
-    "/users", responses=login_required_responses, response_model=List[schemas.User], tags=["user"]
+    "/users", responses=login_required_responses, response_model=query.UserResponse, tags=["user"]
 )
 async def get_users(
     db: Session = Depends(get_db),
     user: models.User = Depends(admin_required),
+    pagination: Pagination = Depends(),
 ):
-    users = db.query(User).all()
-    return users
+    users = db.query(User)
+    return pagination.response(users)
 
 
 @router.post(
--- a/nmdc_server/api.py
+++ b/nmdc_server/api.py
@@ -561,6 +561,5 @@ async def get_users(
     db: Session = Depends(get_db),
     user: models.User = Depends(admin_required),
 ):
-    print(user.orcid)
     users = db.query(User).all()
     return users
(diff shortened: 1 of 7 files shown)


nmdc_server/api.py
```diff
@@ -552,3 +552,28 @@ async def submit_metadata(
...
+
+
+@router.get(
+    "/users", responses=login_required_responses, response_model=List[schemas.User], tags=["user"]
+)
+async def get_users(
+    db: Session = Depends(get_db),
+    user: models.User = Depends(admin_required),
+):
+    users = db.query(User).all()
```
DeepikaGhodki, line 564: @subdavis FYI - given the volume of users in not high, we're fetching all rows at once as of now. Would you recommend introducing pagination parameters here?

subdavis: Yeah, I think server-side pagination is a good idea here.   I need to add it to the submission list as well.

web/src/data/api.ts
```diff
@@ -309,6 +309,13 @@ export interface BinResponse<T = string | number> {
   facets: number[];
 }
 
+export interface User{
+    id: string,
+    orcid: string;
+    name: string;
+    isAdmin: boolean;
```
subdavis, line 316:
```suggestion
    is_admin: boolean;
```

subdavis (changes requested):
This looks really good!  Aside from a couple comments, please see my pagination adaptation in #734 and do something similar here.

💯

web/src/views/User/UserPage.vue
```diff
@@ -0,0 +1,88 @@
...
+    <v-container>
+      <v-card flat>
+        <v-card-title class="text-h4">
+          Users
+        </v-card-title>
+        <v-card outlined>
+          <v-data-table
+            dense
+            :headers="headers"
+            :items="users.data.results"
```
subdavis, line 57:
```suggestion
            :items="users.data.results.results"
```

subdavis (changes requested):
Small issue

web/src/views/User/UserPage.vue
DeepikaGhodki: Oh thanks!

subdavis (approved):
:100: Awesome

ssarrafan:
completed in July

Issue #723, opened 2022-06-20
Set and test session cookie attributes


mcovalt:
Currently, after the authorization flow with ORCiD is complete, our backend provides the frontend with a session cookie containing the JWT. This cookie is foundational to the rest of our authorization workflow, so it must be secure.

Ensure the following attributes are set on that session cookie:
- HttpOnly
- SameSite: Strict
- Secure

Provide a test on the backend that proves cookie responses set these attributes.

*Note: the above attributes generally prevent XSS and CSRF on modern web browsers. Older browsers that don't work correctly with the above attributes may still pose risks.*

naglepuff:
See #1257 
We no longer use the session cookie for authentication. Closing this issue

PR #735, opened 2022-06-30
Setting cookie headers

DeepikaGhodki:
### Changes (#723 )

**Configuration Changes for Session Middleware**

- Setting the following cookie headers in production environment

    - HttpOnly (set by default) 
    - SameSite: Strict (default setting is lax)
    - Secure (set by setting https_only to true)

**Testing**

- Parameterized create_app() method to take secret_key as an input: to facilitate signature validation for a cookie during testing 
(This change can be reverted for now as the current test is not asserting set-cookie field in request headers)
- Added a new test that checks if the above mentioned headers are set in the cookie

The above mentioned test performs the following:
- encode the token object and sign it using the secret key provided in `configtest.py` => this gives the session cookie value (JWT token) (Ref: [Session Middleware](https://github.com/encode/starlette/blob/master/starlette/middleware/sessions.py#L62-L64))
- make a GET call with the above cookie set in request
- assert if `Set-Cookie` header is set in response

**TODO**
- [ ] Test the cookie headers in staging environment

### DIFF
--- a/nmdc_server/app.py
+++ b/nmdc_server/app.py
@@ -32,12 +32,14 @@ def create_app(env: typing.Mapping[str, str], secret_key: str = settings.secret_
     app.include_router(api.router, prefix="/api")
     app.include_router(auth.router, prefix="")
     same_site_value = "lax"
+    https_only = False
     if settings.environment == "production":
         same_site_value = "Strict"
+        https_only = True
     app.add_middleware(
         SessionMiddleware,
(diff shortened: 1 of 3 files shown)


subdavis (commented):
Adding review comment to remove the task from my queue.

naglepuff:
superseded by #1257 
we no longer use the session cookie for authentication

PR #743, opened 2022-07-12
Reorder merge of `SubmissionMetadata` to after `User`

mcovalt:
`SubmissionMetadata` has a foreign key to `User` that is violated when
ingested out-of-order.


ssarrafan:
Completed in July

PR #788, opened 2022-09-13
Prevent zooming too far when map data changes

### LINKED ISSUE #783: Set minimum map zoom when programmatically updating map bounds
Currently the map zooms very close when the filter updates. Update this to only zoom to something like the state level. Manual interaction should still allow zooming to closer levels to see more detail in sample locations.

naglepuff:
Resolves #783 

Previously, when filtering through data using the sidebar would cause the map bounds to change, the map would zoom in too far if the results of the query were small. Now, when the map bounds change due to updated search queries, there is a maximum zoom level. This allows users to maintain context for where the data is without needing to zoom back out. Users can still zoom in towards the data.

Without max zoom:
without_max_zoom

With max zoom = 5:
with_max_zoom

### DIFF
--- a/web/src/components/ClusterMap.vue
+++ b/web/src/components/ClusterMap.vue
@@ -133,7 +133,11 @@ export default defineComponent({
 
     watchEffect(() => {
       if (mapRef.value && mapCenter.value) {
-        mapRef.value.fitBounds(mapCenter.value, { padding: [20, 20] });
+        const fitBoundsOptions = {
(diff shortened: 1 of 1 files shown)


mcovalt (approved):
Perfect. Folks will be glad this is implemented. After merging this, check out the GitHub Action "deploy / build (client)". Once that completes, the updated behavior should be visible on data.dev.microbiomedata.org.

PR #797, opened 2022-10-27
Migrate ingest to schema version 7

### LINKED ISSUE #790: Migrate data portal to newest version of schema

naglepuff:
Resolves #790 

Summary of update:

- `env_broad_scale`, `env_local_scale`, and `env_medium` had their representation changed in the new schema. Our ingest now accommodates that
- `part_of` slot renamed to `sample_link`
- Add `emsl_`, `igsn_` and `img_identifiers` to `alternative_identifiers`
- Add `depth` to the stored json, since `depth` and `depth2` have been collapsed into one slot
- rename several columns for consistent casing


mcovalt (approved):
These changes look good to me and I can't find an instance where the UI is broken as a result. The database migrations work, but I've created back-ups if things go awry.

Next step is to merge and get the migrations to go on dev. Then we'll change the MongoDB URI to point to the MongoDB instance containing migrated data. At that point, we can run an ingest to make sure all is well.

PR #798, opened 2022-10-27
Fix DOI representation

### LINKED ISSUE #772: Fix DOI presentation on study page
Currently on ingest, the "doi" slot and "publication_dois" slots are [mapped uniformly](https://github.com/microbiomedata/nmdc-server/blob/main/nmdc_server/ingest/study.py#L56-L59) to doi objects, and then the client attempts to [separate them back out](https://github.com/microbiomedata/nmdc-server/blob/main/web/src/views/IndividualResults/StudyPage.vue#L122-L127) by assuming that the dataset DOI (and no others) will have the "dataset" citation type. This causes duplicated and malformed DOI links when that is not the case (see e.g. #606 and the newly created study https://data.microbiomedata.org/details/study/gold:Gs0110119).

We should not be making this assumption, but should instead explicitly keep the provenance of which slot each publication came from and consistently place "doi" slot in the "Dataset Citation" heading and "publication_dois" in "Other Publications" heading.

mcovalt:
Closes #772

A study's DOI information is related to:

- the study's produced dataset
- publications using the dataset

Historically, the type of each of the study's DOIs was indicated via attributes on the DOI information. The schema has since changed  so that each study can indicate separately its dataset's DOI and publication DOIs. The backend supports this schema, but the frontend was not updated to.

This PR:

- updates the frontend to use the explicit DOI form
- normalizes the form of DOI data (previously could be a URL or just the DOI)
- conditionally renders the dataset's DOI to hide this information if not available

### DIFF
--- a/nmdc_server/ingest/study.py
+++ b/nmdc_server/ingest/study.py
@@ -1,3 +1,4 @@
+import re
 from typing import Optional
 
 import requests
@@ -38,6 +39,11 @@ class Study(StudyCreate):
         return extract_extras(cls, values)
 
 
+def transform_doi(doi: str) -> str:
+    matches = re.findall(r"10.\d{4,9}/[-._;()/:a-zA-Z0-9]+$", doi)
+    return matches[0]
+
+
 def load(db: Session, cursor: Cursor):
     for obj in cursor:
         pi_obj = obj.pop("principal_investigator")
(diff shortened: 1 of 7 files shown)


naglepuff (approved):
Overall changes look straightforward and reasonable. I was able to successfully migrate my local database, which cleaned up the entries in the FK columns and DOI `id` column.

I also compared the study detail pages for the study `gold:Gs0154044` to view the changes and verify new behavior.

Once the linting is fixed this would be fine to merge. It would be useful to create a dump of the dev database after merge (and perhaps before) and update the Notion page with the recent dump for project spin-up.

PR #810, opened 2022-12-07
Update python and nmdc-schema

naglepuff:
This change updates the minimum python version required for `nmdc-server` to 3.9.

With #797, our migration script and database supports ingest of data compliant with NMDC schema version 7. This change will allow us to use the most recent release of the `nmdc-schema` package as well, since that required python>3.9.

### DIFF
--- a/.github/workflows/server.yml
+++ b/.github/workflows/server.yml
@@ -26,7 +26,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: [3.9, 3.10]
+        python-version: [3.9, '3.10']
 
     steps:
     - uses: harmon758/postgresql-action@v1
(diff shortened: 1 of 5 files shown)


naglepuff:
I've run a local ingest with these changes, which finished successfully. I'm also not sure how to change the GH actions so that 3.7 and 3.8 are no longer "expected" checks.

mcovalt:
> I've run a local ingest with these changes, which finished successfully. I'm also not sure how to change the GH actions so that 3.7 and 3.8 are no longer "expected" checks.

I've changed the GitHub rules to require 3.9/3.10 tests to pass.

mcovalt (approved):
Looks good! After merging and GitHub Actions finishes deploying this to SPIN, we'll run a full ingest on dev.

PR #816, opened 2022-12-15
Pin requests, nmdc_schema, unpin click

naglepuff:
This resolves dependency conflicts introduced by recent updates to the nmdc_schema package.


mcovalt:
Strange that the `mypy` tests are failing. Two things:
- I think this is because `mypy` is doing dependency check of our dependencies which we do not want. [We can disable checking the `.tox` folder.](https://mypy.readthedocs.io/en/stable/config_file.html#confval-exclude)
- This test is failing for Python 3.10 which we do not use. We can disable that check.

naglepuff:
> Strange that the mypy tests are failing. Two things:

   > I think this is because mypy is doing a type check of our dependencies which we do not want. [We can disable checking the .tox folder.](https://mypy.readthedocs.io/en/stable/config_file.html#confval-exclude)
    This test is failing for Python 3.10 which we do not use. We can disable that check.

Disabling the check for our dependencies makes sense to me. As for the 3.10 test, does it make sense to just get rid of it as an environment altogether? i.e. both from the tox config and the gh actions config?

mcovalt:
Yes, I think so.

mcovalt (approved):

PR #817, opened 2022-12-19
Move documentation endpoint

mcovalt:
Closes https://github.com/microbiomedata/issues/issues/19

This PR moves the API docs from "/docs" to "/api/docs".

Although references to "/docs" have been updated in the README.md, a [301 Moved Permanently](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/301) response is added to the backend in case this is being referenced elsewhere.

Test this by:
- visiting "/api/docs" and ensuring the documentation is visible
- visiting "/docs" and ensuring you're redirected to "/api/docs"

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ NMDC_HOST=http://localhost:8080
 docker-compose up -d
 ```
 
-View main application at `http://localhost:8080/` and the swagger page at `http://localhost:8080/docs`.
+View main application at `http://localhost:8080/` and the swagger page at `http://localhost:8080/api/docs`.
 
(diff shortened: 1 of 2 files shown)


naglepuff (approved):
Nice. Looks like there are also a few links in the wiki like [here](https://github.com/microbiomedata/nmdc-server/wiki/Data-Portal-Management#step-1-ingest-into-the-staging-database) and [here](https://github.com/microbiomedata/nmdc-server/wiki/Search-API-Docs#full-example-searching-for-biosamples), which should probably also be updated, but I'll approve as-is.

Issue #825, opened 2023-01-05
Location of openapi.json incompatible with unified api


naglepuff:
Our unified api is located at api-dev.microbiomedata.org. Requests to that url go to the runtime api, while requests to api-dev.microbiomedata.org/api go to the nmdc-server frontend (data portal). 

We want to be able to navigate to the api documentation via `/docs` for the runtime api, and via `/api/docs` for the data portal api. However, it appears that FastAPI (which is used by both of these apis) redirects to `openapi_url` when the `docs_url` endpoint is hit. Since both these apis use the default value, (`/openapi.json`), both `/docs` and `/api/docs` redirect to `api-dev.microbiomedata.org/openapi.json`, showing the documentation for the runtime api only.

FastAPI allows you to change where the open api docs are served from. Changing that for the data portal API could fix this issue. This issue should be considered fixed if `api-dev.microbiomedata.org/api/docs` correctly served the openapi documention for the data portal api.

naglepuff:
This was solved by #827

PR #827, opened 2023-01-06
Move location of openapi schema

naglepuff:
Solves #825 

This is to improve the data portal's API compatibility with the reverse proxy set up to unify this API with the NMDC runtime API. Due to how FastAPI's documentation endpoint works, we needed to move where the openapi json schema was served from to ensure compatibility with the unified API.

Previously, [api-dev.microbiomedata.org/api/docs](api-dev.microbiomedata.org/api/docs) was serving the runtime API documentation, where it should be serving the data portal API documentation.

To Test:
- visit /api/openapi.json and make sure the schema is served from that endpoint.

### DIFF
--- a/nmdc_server/app.py
+++ b/nmdc_server/app.py
@@ -25,6 +25,7 @@ def create_app(env: typing.Mapping[str, str]) -> FastAPI:
         title="NMDC Dataset API",
         version=__version__,
         docs_url="/api/docs",
+        openapi_url="/api/openapi.json",
     )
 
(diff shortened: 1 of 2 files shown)


jeffbaumes (approved):

jeffbaumes:
@naglepuff looks like there is an api test which assumes the location of `openapi.json` that is failing.

naglepuff:
Fixed failing test with d9d5632

PR #840, opened 2023-01-31
Persist facility, award, and shipping data

naglepuff:
This PR holds the set of changes that persist submission portal data collected from the new context/address form. For more information on how that form integrates into the existing workflow, see [this diagram](https://lucid.app/lucidchart/75516c27-0974-4a77-ae92-73234e3f385e/edit?invitationId=inv_8fe7dca1-e52f-431c-9a0e-f1e6dd1e3a72&page=0_0#).

This PR contains changes to our pydantic models. Since attempting to load old data (without context/address form) would result in errors, a database migration that converts metadata submissions to be compatible with model changes is also included in this set of changes.


jeffbaumes (approved):
Did live code review via video call just now, looks good!

PR #850, opened 2023-02-08
Context and address form validation

naglepuff:
This PR builds some front-end validation into the new context and address forms. Validation added in this PR is as follows:

**Context Form**
1. The "Has data already been generated for your study?" choice is required
2. If data has been generated, the DOI field is required (this field has been moved from the multiomics form)
3. If a facility (EMSL or JGI) is selected from the "Are you submitting metadata for samples that will be sent to a DOE user factility?" an award type is required
4. If the Award type is "Other," the associated text field must have a non-empty string

**Address Form**
1. Only show IRB fields if biosafety level is BSL2

**Additional Changes**
1. New `enums` for `BiosafetyLevels` and `AwardTypes`
2. Sample type drop down (address form) pulls in values from the NMDC schema

### DIFF
--- a/web/src/views/SubmissionPortal/Components/SubmissionContextForm.vue
+++ b/web/src/views/SubmissionPortal/Components/SubmissionContextForm.vue
@@ -21,7 +21,7 @@ export default defineComponent({
     const projectAwardValidationRules = () => [(v: string) => {
       const awardChosen = v === 'MONet' || v === 'FICUS' || v === contextForm.otherAward;
       const valid = awardChosen || contextForm.facilities.length === 0;
-      return valid || 'If submitting to a use facility, this field is required.'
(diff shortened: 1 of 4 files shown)


web/src/views/SubmissionPortal/Components/SubmissionContextForm.vue
```diff
@@ -1,40 +1,66 @@
...
-      showAddressForm,
-      addressFormRef,
-      address,
-      date,
-      datePicker,
-      sampleItems,
-      biosafetyLevels,
+      projectAwardValidationRules,
+      otherAwardValidationRules,
+      reValidate,
```
jeffbaumes, line 63: I'd rename to `revalidate`

jeffbaumes (approved):
Put space before *
Fix "factilities"

PR #854, opened 2023-02-09
Fix bug caused by empty sheet data

pkalita-lbl:
Small bug fix for missing null check in multitab DH work.

### DIFF
--- a/web/src/views/SubmissionPortal/HarmonizerView.vue
+++ b/web/src/views/SubmissionPortal/HarmonizerView.vue
@@ -285,7 +285,7 @@ export default defineComponent({
         // they aren't there already.
         (nextData[environmentKey] || []).forEach((row) => {
           const rowId = row[SCHEMA_ID];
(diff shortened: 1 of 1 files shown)


naglepuff:
What's a good way to reproduce the bug?

pkalita-lbl:
1. Start a _new_ submission (this is important and probably why I didn't catch it sooner -- I often tested with existing submissions). Ensure some of the EMSL or JGI checkboxes are checked on the Multiomics Data screen.
2. Once you reach the DataHarmonizer screen enter some data in the first three columns like:

sample name | globally unique ID | analysis/data type
-- | -- | --
1 | A | metagenomics
2 | B | metagenomics
3 | C | metabolomics

3. Click on the second tab
4. Verify that on the second tab the first three columns are read-only and that it is only showing rows relevant to that tab based on the value of the analysis/data type column

naglepuff (approved):
Looks good to me. Thanks for the testing instructions. Also see #855, which I noticed while testing this.

PR #857, opened 2023-02-13
Upgrade data-harmonizer library version

pkalita-lbl:
The latest version of the `data-harmonizer` package includes a fix to how it parses/formats multivalued slot values to/from delimited strings. This fixes a bug in the submission portal where rows with multiple values in the "analysis/data type" column would not be persisted across multiple DataHarmonizer tabs. 

The fix can be verified by:
1. On any submission portal entry, go to the DataHarmonizer screen. For any row select multiple values for the "analysis/data type" column. It should say, for example, "metabolomics; metagenomics" (note the space after the semicolon).
2. Click the "Validate" button
3. Verify that the "analysis/data type" cell you edited has not changed. It should still read "metabolomics; metagenomics" with a space after the semicolon.

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -25,7 +25,7 @@
     "d3-selection": "^3.0.0",
     "d3-time": "^3.0.0",
     "d3-time-format": "^4.1.0",
-    "data-harmonizer": "1.4.0",
+    "data-harmonizer": "1.4.5",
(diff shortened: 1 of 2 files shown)


jeffbaumes (approved):
Just a library version change that seems fine. Approving and merging now.

PR #860, opened 2023-02-16
Run DataHarmonizer tab data sync before performing export

pkalita-lbl:
Fixes https://github.com/microbiomedata/issues/issues/76

The diff makes it look like more changes than it really is. The core is moving data synchronizing logic that used to exist only in `changeTemplate` to a new function called `synchronizeTabData`. That function can then be called on tab change _and_ on export (`downloadSamples`). I had to shuffle around the order of some functions to keep the linter happy.


jeffbaumes (approved):
Looks good, will merge momentarily

PR #863, opened 2023-02-18
Upgrade to data-harmonizer@1.4.6

pkalita-lbl:
No changes to functionality in this release. It is built in a new way that attempts to reduce the main bundle size.


naglepuff (approved):
Dependency version change looks good to me.

PR #874, opened 2023-02-28
Refine logic for computing templateChoiceDisabled from sampleData

pkalita-lbl:
Fixes https://github.com/microbiomedata/issues/issues/87


naglepuff (approved):
LGTM

PR #876, opened 2023-03-02
Remove "Menus" entry from column help

pkalita-lbl:
Fixes https://github.com/microbiomedata/issues/issues/88

To see the difference, in the DataHarmonizer view of the submission portal double-click on the header of any column that has a drop-down menu. In the resulting modal, there should not be a "Menus" entry.


naglepuff (approved):
LGTM. After updaing DH I was able to run this branch locally and verify that the issue is resolved.

PR #888, opened 2023-03-07
Fix typos in variable names

pkalita-lbl:


naglepuff (approved):
LGTM. I added an additional commit to prevent key errors in one of the other migrations. Tested locally with a recent backup of prod submissions and it seems to work now.

PR #889, opened 2023-03-09
Fix XLSX export bug when entire DH is empty

pkalita-lbl:
Fixes https://github.com/microbiomedata/issues/issues/76

This handles an edge case that I missed in the original fix for that issue. Specifically, these changes ensure that the necessary fields are populated in the `sampleData` object even in the case where _no_ data has been entered in DH interface. To test: create a new submission, once you reach the DH screen click the "Download XLSX" button before any other interaction.


naglepuff (approved):
LGTM.

PR #894, opened 2023-03-17
Implement WordPress-style app bar

pkalita-lbl:
Related to https://github.com/microbiomedata/issues/issues/123.

These changes are mainly aimed at getting the portal's nav header to have similar organization and styling as the WordPress site's header. I don't want to close the issue with these changes because I'm sure there will be a few rounds of iterating on the exact content of the menus. Finally once that's all sorted out, I'll go back and update the WordPress header to ensure the content there matches.

### DIFF
--- a/web/src/App.vue
+++ b/web/src/App.vue
@@ -1,14 +1,13 @@
 <script lang="ts">
 import { defineComponent } from '@vue/composition-api';
 import AppHeader from '@/components/Presentation/AppHeader.vue';
-import { setConditions, stateRefs } from '@/store/';
+import { stateRefs } from '@/store/';
 
 export default defineComponent({
   name: 'App',
   components: { AppHeader },
   setup() {
     return {
-      setConditions,
       stateRefs,
     };
   },
@@ -17,7 +16,7 @@ export default defineComponent({
 
 <template>
   <v-app>
-    <app-header @logoClicked="setConditions([])" />
+    <app-header />
     <keep-alive>
       <router-view />
     </keep-alive>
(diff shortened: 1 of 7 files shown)


web/src/App.vue
```diff
@@ -17,111 +17,7 @@ export default defineComponent({
...
-      <v-spacer />
-
-      <v-btn
-        depressed
-        small
-        href="https://nmdc-documentation.readthedocs.io/en/latest/tutorials/nav_data_portal.html"
-        target="_blank"
-        rel="noopener noreferrer"
-        class="mx-2"
-      >
```
naglepuff, line 45: Is there going to be a place for these data portal tutorial/documentation links? I'm not seeing them anywhere in the new app bar.

naglepuff (commented):
Looks good, I just had one question/concern about some of the links that previously existed that I can't seem to find.

web/src/App.vue
pkalita-lbl: Yeah that's kind of what I meant when I was alluding to iterating on the content of the menus. But actually let's hold off on merging this for right now and I'll see if I can get some engagement team members to provide some input first.

pkalita-lbl:
After talking with Ingrid and Julia, I've made the following updates:
* The header logo now links to the home page proper (instead of the portal "home") in order to be consistent with the WordPress header.
* The user guide, docs, and video tutorial links that were previously in the portal header have been moved into a top-level help menu in the search view itself. It reuses the same icon and placement idioms that already exist on the page, but this one triggers a menu instead of a tooltip. 

The rest of the header content was approved as-is.

web/src/App.vue
pkalita-lbl: See latest comment

naglepuff (approved):
LGTM. I think it would be worth the effort to make the documentation/tutorial/video slightly more discoverable, but I'll capture that concern in an issue so the unification of style between the data portal and the wordpress website can proceed.

PR #896, opened 2023-03-20
Ingest aggregated gene function collections

naglepuff:
This PR updates how we build the gene function aggregation tables during ingest. Previously, we ingested data into a few tables, such as `mga_gene_function` and the various `peptide_` tables, and then used a SQL query to populate `metap_gene_function_aggregation` and `mga_gene_function_aggregation`.

Now we ingest data from mongo collections that contain already aggregated data.


jeffbaumes (approved):

nmdc_server/ingest/pipeline.py
```diff
@@ -78,26 +83,49 @@ def load_mags(db: Session, obj: Dict[str, Any], **kwargs) -> LoadObjectReturn:
...
     pipeline = cast(models.MetaproteomicAnalysis, load_mp_analysis_base(db, obj, **kwargs))
 
-    for p in peptides:
-        if not db.query(models.MGAGeneFunction).filter_by(subject=p["best_protein"]).first():
-            continue
+    annotations: Collection = kwargs["annotations"]
+
+    query = annotations.find(
+        {
+            "metaproteomic_analysis_id": pipeline.id,
```
jeffbaumes, line 92: @naglepuff I believe we want to add a line here to only get ones where best_protien is set to true:
```
"best_protein": True,
```

naglepuff: You're right. Added in c23c4de

Issue #835, opened 2023-01-30
Allow submission of metadata in submission portal


naglepuff:
For now, clicking the "Submit" button should change a submission's status to "Submitted: In Review."

naglepuff:
This should be accompanied by the addition of a "Status" enum to the schema. Initial values could just be "In Progress" and "Submitted: In Review"

mslarae13:
For GSP, in production (& sandbox), the 'Submit' button should be enabled. 
What should the submit button do? : Change the 'Status' of the submission from "In Progress" to "Submitted- Pending Review"

What it looks like?: The submit button is grey. Once metadata passes validation, the button will turn green and can be clicked. A pop up should appear that says "You are about to submit this study and metadata for NMDC review. Would you like to continue? Yes- Submit, Cancel" 

Can a submitted sample set revert back the 'in progress' status: Not by the user, for now. Once we have additional status, yes.

**NOT for GSP**
Addition of the other metadata status values: https://docs.google.com/spreadsheets/d/18tN1_KlQFHq1FFaBTT3zzusEn57tkidAfj-3nC5wDfM/edit?usp=sharing

Submission stats can be changed from "Pending NMDC Review" back to "In Progress" via the main table by a  user. Adding an "Edit this submission" button. However, once the status of the submission has progressed to any of these listed below, the user cannot edit the submission without contacting an NMDC POC
Approved-Held
Pending- Sent to User Facility
Released
Approved- User Facility

Status "Denied- Updates Required" can be re-opened by user.

NOTHING from the submission portal db should go to the Data Portal db until "Released" is the status.

naglepuff:
This is being worked on in #899

mslarae13:
If I start a submission, leave it or the "submission context" page, the EMSL shipping information appears as incomplete, red box... I have to open it, and close it (no changes made) to move forward

pkalita-lbl:
> Why did all the required EMSL and JGI terms become recommended?

Looks like this is an artifact of moving from the old Google sheet to TSV files in the submission schema repo. New issue -> https://github.com/microbiomedata/issues/issues/196

> Why are the envo terms a long _________ now? (________________mediterranean savanna biome [ENVO:01000229])

[...]

PR #899, opened 2023-03-22
Add Submitted- Pending Review status

naglepuff:
Solves #835 

This PR makes some minor updates to the way submitting samples works in the submission portal. 

Namely, it attempts to start creating a set of valid submission statuses based on what will eventually be needed for the system. "In Progress," "Submitted- Pending Review," and "Complete" are the only statuses defined  here so far.

A new dialog window has been added to the workflow, as described in the linked issue. See the video for a short demo. (@mslarae13 take a look if you'd like before this goes into dev, and I can address any feedback you have)

https://user-images.githubusercontent.com/7085625/228962627-e26f2b79-83c8-49de-b60a-ea503ba44d18.mp4

### DIFF
--- a/web/src/views/SubmissionPortal/HarmonizerView.vue
+++ b/web/src/views/SubmissionPortal/HarmonizerView.vue
@@ -464,24 +464,17 @@ export default defineComponent({
             </v-icon>
           </v-btn>
         </label>
-        <div v-if="validationErrorGroups.length === 0">
-          <v-btn
-            color="primary"
-            outlined
-            @click="validate"
-          >
-            2. Validate
-            <v-icon class="pl-2">
-              mdi-refresh
-            </v-icon>
-          </v-btn>
-          <span
-            v-if="!tabsValidated[activeTemplateKey]"
-            class="warning--text"
-          >
-            This tab must be validated before being submitted.
-          </span>
-        </div>
+        <v-btn
+          v-if="validationErrorGroups.length === 0"
+          color="primary"
+          outlined
+          @click="validate"
+        >
+          2. Validate
+          <v-icon class="pl-2">
+            mdi-refresh
+          </v-icon>
+        </v-btn>
         <v-card
           v-if="validationErrorGroups.length"
           color="error"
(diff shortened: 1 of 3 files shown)


web/src/views/SubmissionPortal/HarmonizerView.vue
```diff
@@ -163,6 +173,9 @@ export default defineComponent({
       incrementalSaveRecord(root.$route.params.id);
       if (valid === false) {
         errorClick(0);
+        samplesValid.value = false;
+      } else {
+        samplesValid.value = true;
```
naglepuff, line 178: @pkalita-lbl I noticed that `samplesValid` wasn't getting updated anywhere. Does this make sense as the place to update that value? Are there any other places in the `HarmonizerView` that would need to update the value as well?

pkalita-lbl: I think that's the right place to update `samplesValid` but I think keying off of the `valid` value might not be quite right. The value of `valid` in that function is referring to the _current_ tab. I think it would be safer to check the `invalidCells` object, which holds the validation state for _all_ the tabs.

pkalita-lbl: `harmonizerApi.ts::validate` only validates the current tab. There is currently no way to validate all tabs at once (we can chat about the technical reasons why, but it's probably more detail than what's needed here). 

But tab A is validated when it is the current tab and the user selects tab B. So the validation state of all the non-current tabs _should_ be accurate. The big caveat is the case where you have tabs A, B, and C and the user has never clicked on tab C, for example. I wonder if a bit of state to track which tabs have been viewed is warranted, and that the Submit button becomes enabled if there are no validation errors _and_ all tabs have been viewed.

naglepuff: cd9516b replaces `samplesValid` with a new piece of state called `tabsValidated`, which tracks validation across all tabs. If data is edited in the data harmonizer, it is marked as invalid. Now, in order to enable the submit button, a user has to click validate on all tabs.

naglepuff: c33b381 adds new UI features to indicate what the user needs to do in order to submit.

https://user-images.githubusercontent.com/7085625/228962535-cd5c4af1-708d-47e7-a2d0-fc410de1c3f3.mp4

pkalita-lbl: Looks good to me!

turbomam:
Excellent! Remember, the statuses need to be specified in the schema, at least as an enumeration, but preferably as classes that specify the transitions between them. Let me know how I can help make that happen.

turbomam:
I see this issue has been closed. Nice work.

Looking back on my comment https://github.com/microbiomedata/nmdc-server/pull/899#issuecomment-1492387842, I would rephrase my comment as

> we should make a good faith 80/20 type effort to represent the status states outside of code. Ideally we would do that in a way that the code could take as a specification, otherwise the code and external specification might drift apart from each other.

turbomam:
Just asked [claude code](https://docs.anthropic.com/en/docs/agents-and-tools/claude-code/overview)

> analyse web/src/views/SubmissionPortal/Components/SubmissionList.vue , web/src/views/SubmissionPortal/HarmonizerView.vue and web/src/views/SubmissionPortal/store/index.ts. what does it mean for a submission
   to have a status? are the status levels defined anywhere? how can a submission change from  one status to another?

turbomam:
In the NMDC server's submission system:

  1. Submission status is defined as an enum in store/index.ts (line 41-46):

```ts
  type SubmissionStatus = 'In Progress' | 'Submitted- Pending Review' | 'Complete';
  const submissionStatus: Record<string, SubmissionStatus> = {
    InProgress: 'In Progress',
    SubmittedPendingReview: 'Submitted- Pending Review',
    Complete: 'Complete',
  };
```

  2. The three status levels are:
    - In Progress: Initial state when users are working on their submission
    - Submitted- Pending Review: When a user has submitted their data for NMDC review
    - Complete: Final state after submission has been approved
  3. A submission changes status through:
    - The submit() function in store/index.ts (line 234-239), which accepts a status parameter
    - In HarmonizerView.vue, the submission status changes from InProgress to SubmittedPendingReview when clicking the "Submit" button (lines 270-275)
    - Status is displayed as a chip in SubmissionList.vue with color-coding (lines 157-162)
    - Only submissions with validated data can be submitted (line 230-236 in HarmonizerView.vue)

  Submissions can only be moved to SubmittedPendingReview by users with appropriate permissions (owner status), as indicated by the permission checks in the submit function and UI constraints.

turbomam:
@mslarae13 what additional documentation is there about submission statuses? 

- who can change them
- what criteria they should follow

turbomam:
@samobermiller you sent me a slack message asking what followup steps I was thinking of. At a minimum I think we should ensure that the code and any documentation files make reference to one another. Two years ago I thought it was really important to model the status states and the rules for their transitions in the schema. I am questioning the return on investment for that now. It  should be easy to document the status states in the schema, but modeling the transitions in the schema and getting the nmdc-server code to implement it's behavior on the schema might be much harder.

PR #900, opened 2023-03-22
Fix bad migration

naglepuff:
For some reason, auto-generating a recent migration created a duplicate constraint on a column. This PR fixes that migration.


naglepuff:
Created issue #901 to track this problem and hopefully prevent this from happening again.

PR #903, opened 2023-03-23
Use v1.0.0 submission schema from PyPI artifact

### LINKED ISSUE #897: Get submission schema from new repo

pkalita-lbl:
The main goal of these changes are to remove the local copy of the submission schema and use the version published to PyPI instead. 

* The checked-in copy of the submission schema has been removed along with the script that fetched it. A new dependency has been added in `package.json`. Admittedly it's a little weird to be getting a front-end dependency from PyPI (LinkML schema repositories aren't generally set up to publish to NPM), but I think it's a step in the right direction. 
* This version of the submission schema changes the names of some classes and slots. An alembic migration is included to handle the name changes in previously saved data.
* Previously the `HARMONIZER_TEMPLATES` map in `harmonizerApi.ts` had a key that was used for both internal bookkeeping as well as a display name. The value included the corresponding schema class name in the `default` field. The class name was then mangled into a slot name in various places by appending the `_data` suffix. With the renaming in submission schema v1.0.0 that mangling is no longer as straightforward. So this seemed like a good opportunity to separate out the different concerns. Therefore with these changes the keys of `HARMONIZER_TEMPLATES` remain the same, but in the value `schemaClass` replaces `default`, `displayName` and `sampleDataSlot` have been added. 

Fixes #897


naglepuff (approved):
Sweet. Love seeing a -55k on the diff, even if it's just the generated schema. I tested your migration against the backed up submissions I had on hand from 3/8/2023, and it worked as expected.

PR #912, opened 2023-03-28
Allow null values on reads QC cols

naglepuff:
Fixes microbiomedata/issues/issues/145

A new batch of `read_qc` analyses were added to mongo. These haven't been ingested properly due to `null` values in the following slots:

`input_read_bases`
`input_read_count`
`output_read_bases`
`output_read_count`

Now, the data model allows for null values in the corresponding columns in postgres.


jeffbaumes (approved):
I guess this was more than a 4-line change 😄 . Looks good, this should hopefully allow those records to be ingested.

naglepuff:
Tested this locally. I'm now seeing workflow IDs like `nmdc:wfrqc*`.

PR #920, opened 2023-03-31
Update ingest for new module mapping

naglepuff:
This updates the ko_terms to module mapping used by the data portal.

Note that I did have to modify the file I downloaded from the NERSC CFS, as it contained duplicate mappings. See 9fac1cf for all of the removed rows.

@jeffbaumes I'd appreciate a review before proceeding to merge.

@emileyfadrosh If you'd prefer I wait for a new file from the IMG team please let me know, and I'll refrain from moving forward with this PR until I get a corrected file.


emileyfadrosh (approved):
Looks like we have the right file now, thanks @naglepuff !!

naglepuff:
@emileyfadrosh One more question before this goes in:

In the new file there are some lines like this: `M00614	KO:K00240-K00241-K00242`
Does this represent three different mappings? Is it ok to split lines like this up onto separate lines?

```
M00614	K00240
M00614	K00241
M00614	K00242
```

emileyfadrosh:
for lines where there is not a 1:1 mapping (e.g., M00614	KO:K00240-K00241-K00242), please split into separate lines. While this is not quite the right representation, it would at least capture that those KO terms are part of a module-within-a-module association. For the current search feature in the portal, this will be sufficient. Post-GSP, we can think about another representation and get input from others. Thanks!

jeffbaumes (approved):
Assuming you have tested module 804 and it now gives a nontrivial result, I approve. Please check again (and ping @emileyfadrosh to check) once it's deployed on data-dev.

PR #921, opened 2023-03-31
Add study-level image

naglepuff:
Fixes microbiomedata/issues/issues/143

Now, if a study-level image URL exists, that will be shown on the study page over the PI image.

image


naglepuff:
This is mostly working as of 384a921, however, I will need to go back and change some things before this can be merged. Namely, the front end actually does need to know is the image is for the study or the PI. Study images are bigger and square shaped, so shoving them into `v-avatar` is not the right thing to do.

jeffbaumes (approved):

PR #925, opened 2023-04-03
Display all GOLD links for a study

### LINKED ISSUE #918: NMDC study pages should show all GOLD studies avaliable
NEON (study id nmdc:sty-11-34xj1150) is the first study to have multiple GOLD studies in mongo prod. The data portal is only displaying the first study record. 
The study page, https://data-dev.microbiomedata.org/details/study/nmdc:sty-11-34xj1150, should display both GOLD study records. 
GOLD:Gs0144570 and GOLD:Gs0161344
Screenshot 2023-03-31 at 10 35 39 AM

cc @emileyfadrosh @ssarrafan

naglepuff:
Client-side changes only. #911 added the column to our data model. This supports multiple gold IDs for a single study being displayed as linked on the study detail page.

Fixes #918

### DIFF
--- a/web/src/views/IndividualResults/StudyPage.vue
+++ b/web/src/views/IndividualResults/StudyPage.vue
@@ -24,6 +24,7 @@ import InvestigatorBio from '@/components/InvestigatorBio.vue';
 const CitationOverrides: CitationOverridesType = {
   '10.46936/10.25585/60000017': 'Doktycz, M. (2020) BioScales - Defining plant gene function and its connection to ecosystem nitrogen and carbon cycling [Data set]. DOE Joint Genome Institute. https://doi.org/10.46936/10.25585/60000017',
 };
+const GoldStudyLinkBase = 'https://gold.jgi.doe.gov/study?id=';
 
 export default defineComponent({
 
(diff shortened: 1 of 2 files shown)


web/src/views/IndividualResults/StudyPage.vue
```diff
@@ -64,6 +65,24 @@ export default defineComponent({
       });
     });
 
+    const goldLinks = computed(() => {
+      if (!item.value) {
+        return [];
+      }
+      const links = new Set();
+      if (item.value.open_in_gold) {
+        links.add(item.value.open_in_gold);
```
jeffbaumes, line 74: With the definition of `gold_url` [here](https://github.com/microbiomedata/nmdc-server/blob/cf4015789ce012c7f8f856362e8c08e918b0d846/nmdc_server/models.py#L33-L38), it would seem that when the id does not start with `gold:`, first gold identifier would be added here, then added again from the gold_study_identifiers list below. Are you doing anything to address this so it is not duplicated?

jeffbaumes (changes requested):

web/src/views/IndividualResults/StudyPage.vue
naglepuff: Yes, I'm using a [set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/add), so that an attempt to add a duplicate link would not basically be a no-op. When running locally, I've checked study pages with various `id` and `gold_study_identifiers`, and see no duplicate links.

naglepuff: If its more clear to use an array and an explicit call to `.includes`, I wouldn't mind changing it. I don't think we get much of a performance gain here from using a set, since this list will probably stay pretty small, but it felt like the right data structure here.

jeffbaumes: Ah, I missed that it is a set. Great makes sense we can keep this as is.

jeffbaumes (approved):

PR #927, opened 2023-04-04
Prevent issues building the GOLD classification tree

naglepuff:


naglepuff:
Closing for now. This was resolved upstream in mongo. However, this should be revisited. Our code makes assumptions that the various environment type fields are required by the `nmdc-schema`, but they are not. See #930

PR #929, opened 2023-04-04
Treat EMSL Biosample IDs Separately

### LINKED ISSUE #919: sample identifier change to string not link

naglepuff:
Fixes #919 

The EMSL IDs that biosamples belonging to the 1000 soils study don't currently go anywhere. Previously, we assumed that all `alternate_identifiers` for biosamples could by combined with identifiers.org to form a URL.
image

Since the EMSL IDs don't represent an external resource (yet), it doesn't make sense to render them as URLs. In order to make this happen, EMSL IDs now have their own column on the `biosample` table called `emsl_biosample_identifiers`. This prevents the frontend from having to do any guesswork as to which IDs in `biosample.alternate_identifiers` are EMSL identifiers. These IDs are now rendered  on the search page like:
image

And on individual sample results pages like:
image


naglepuff:
@jeffbaumes I decided to tackle this issue with what I believe is the "right" solution, instead of just a hack. It ended up being fairly straightforward, although I did end up adding some functionality to the `AttributeItem.vue` component, which I think makes it more flexible overall.

jeffbaumes (approved):

Issue #945, opened 2023-04-07
Ontology url links in submission portal help menus not working


aclum:
Links in the help menu in the submission portal to  http://purl.obolibrary.org/obo/ are being truncated. An example field is broad-scale environmental context. Any columns that have links to purl.obolibrary.org (ie soil type, broad-scale environmental context, environmental medium)
![sub port help menu issue](https://user-images.githubusercontent.com/8196591/230678778-d06f4b16-3e88-404f-a397-cea634a2467a.png)

@mslarae13 mentioned this may be a regression from the schema sync and the sheets and friends info was lost.

cc @pkalita-lbl 

Using chrome version 112.0.5615.49

aclum:
duplicate of https://github.com/microbiomedata/nmdc-server/issues/820

aclum:
@naglepuff Is it possible to fix the help menu? @pkalita-lbl said this is different that the pop-up box which is controlled by Data Harmonizer which is what Montana reported in https://github.com/microbiomedata/nmdc-server/issues/820

pkalita-lbl:
I'm working on this now.

PR #949, opened 2023-04-07
Use linkify-it package for more robust matching of URLs

pkalita-lbl:
It seems like the existing regex was mostly getting tripped up on underscores. But instead of handling just that one case I figured it might be better to delegate to a library that is purpose-built for parsing URLs out of strings.

Fixes https://github.com/microbiomedata/nmdc-server/issues/945


naglepuff (approved):
LGTM. Tested this out locally and links with underscores seem to be working as they should. Additionally, the `linkify-it` package looks like it is made by a reputable team and is itself a small library with minimal dependencies.

mslarae13:
The ENVO links are still broken....
Screen Shot 2023-04-13 at 10 06 29 AM

pkalita-lbl:
Working on dev:
image

PR #955, opened 2023-04-21
Unify behavior of study page filters

### LINKED ISSUE #933: study page link to omics procesing set records does not filter studies correctly
From a study page the link to sample count filters correctly but the link to omics processing records doesn't at the study level correctly. Based on how the search is being done the difference is a id [is] $nmdc_study_id vs study id is $nmdc_study_id see attached screenshots. id [is] produces the desired behavior, study id [is] does not produce the desired behavior.
study id vs id from study page
id as filter
study id as filter

Created a new issue to separate concerns from https://github.com/microbiomedata/issues/issues/168

naglepuff:
Fixes #933 

**Previous Behavior:**
When on a study page, if a user clicked the sample count, they would be taken back to the main data portal view, and the study list would be filtered to show only the study they were just looking at. If the user clicked one of the omics processing chips, they would be taken back to the main data portal view, and the study would be selected (indicated by the checkboxes next to the study titles). If a study is further down the list, the visual cue would be missing.
See linked issue for screenshots

**New Behavior**
Now, both actions produce the same result: a returning to the main data portal view with only the study of interest present in the studies list. If the user wants to see all studies, they have to remove the filter from the search sidebar.

### DIFF
--- a/web/src/store/index.ts
+++ b/web/src/store/index.ts
@@ -55,6 +55,23 @@ function setConditions(conditions: Condition[], push = false) {
 }
 
 /**
+ * For each condition, remove all others with a similar table & field.
+ */
+function setUniqueCondition(
+  field: string[],
+  table: string[],
+  conditions: Condition[],
+) {
+  const others = state.conditions.filter((c) => (
+    !field.includes(c.field)) || (!table.includes(c.table)
+  ));
+  setConditions([
+    ...conditions,
+    ...others,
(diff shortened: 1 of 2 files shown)


aclum:
@jeffbaumes  would you please review this pull request? I'm trying to determine what outstanding issues identified by GSP still need work.

jeffbaumes (approved):

PR #972, opened 2023-05-26
Update readme

marySalvi:


docs/development.md
```diff
@@ -0,0 +1,158 @@
+# Development Setup
```
marySalvi, line 1: @naglepuff Do you think this is clear enough or should I add a TOC? Specifically the Docker vs outside docker portion breaks the flow a bit.

naglepuff: I think its fine as is. I wouldn't object to a TOC, but I don't think its necessary at this point.

naglepuff (approved):
Thanks for updating this!

PR #977, opened 2023-06-08
Upgrade to submission schema 7.6.0

pkalita-lbl:
This upgrades to the latest version of the submission schema, 7.6.0 (there was a change in the version numbering scheme to align the version numbers with the main schema, hence the large jump). Previously the `source_mat_id` column was used to link rows across different DH tabs. That slot is no longer required, so `samp_name` -- which is required and values must be unique -- is used instead.


naglepuff (approved):
LGTM

PR #981, opened 2023-06-13
Sync header with WordPress header

### LINKED ISSUE #269: Change text for organic matter characterization
I think there have been a few requests around this, tagging @corilo since I think he's the subject matter expert on how we should refer to this data.

image (1)

Feedback:
Can we just use "organic matter" instead of with the word "characterization" in the label? We can still show the long name as a tool-tip.

pkalita-lbl:
Fixes https://github.com/microbiomedata/issues/issues/269

### DIFF
--- a/web/src/menus.ts
+++ b/web/src/menus.ts
@@ -28,10 +28,6 @@ const Menus = [
         href: 'https://microbiomedata.org/fair/',
       },
       {
-        label: 'Data',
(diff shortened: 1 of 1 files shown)


naglepuff (changes requested):
Code looks good, but could you update the linked issue to point to microbiomedata/issues/issues/269?

naglepuff (approved):

PR #983, opened 2023-06-22
Handle null gold slots

marySalvi:
This removes null values from the gold classification tree. It is important to note that this effects all the tree and as such should be discussed further. We may need a more sophisticated solution instead of this all or nothing approach.


naglepuff (approved):
LGTM

PR #984, opened 2023-06-23
Wrap URLs in InvestigatorBio

### LINKED ISSUE #942: neon study logo is now on top of PI name instead of to the left of it
This is happening in both dev and prod now. As of Wednesday the study logo was showing to the left of the PI name, the same place where profile_image_url (with a pi picture was). for the [NEON study page ](https://data.microbiomedata.org/details/study/nmdc:sty-11-34xj1150) the study logo is above the PI name and the PI name and websites aren't indented as expected.

**expected behavior**
neon study page 20230405

**observed behavior**
neon study page 20230407

naglepuff:
This prevents stacking the PI/study image until the window becomes much narrower. It has the added benefit of unifying text wrapping behavior between web browsers. See the conversation in the linked issue for more details.

Fix #942

### DIFF
--- a/web/src/components/InvestigatorBio.vue
+++ b/web/src/components/InvestigatorBio.vue
@@ -69,7 +69,7 @@ export default defineComponent({
             v-for="site in piWebsites"
             :key="site"
             class="blue--text py-1"
-            style="cursor: pointer; text-decoration: none; display: block;"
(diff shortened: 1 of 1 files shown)


marySalvi:
I actually think that the issue here is that the data citation component spans 5 cols for the whole page for all breakpoints.
If we had this stack at some breakpoint and also not display when empty this should solve this problem and a few other responsive issues.

image

PR #986, opened 2023-07-03
Add some documentation for study page

### LINKED ISSUE #978: Create annotated study page screenshot which maps page elements to schema slots

naglepuff:
Fix #978 

This is a first pass at adding documentation to help explain how the study detail page is populated with data as of July 3, 2023. The upcoming study page refresh squad will likely change some aspects of this page, so the documentation will need to be updated accordingly.


docs/study_page_annotations.md
```diff
@@ -0,0 +1,49 @@
...
+## __External Resources: Additional Data__
+
+![Additional Data, ESS Dive Identifiers](./images/additional_data_ess_dive.png)
+
+![Additional Data, massive identifiers](./images/additional_data_massive.png)
+
+### __Explanations__
+
+- **ESS Dive Dataset** links are derived from the `ess_dive_datasets` slot of `Study`. The frontend of the data portal prepends `https://identifiers.org/` to each ID it finds in this slot.
+- **MassIVE Study** links are derived from the `massive_study_identifiers` slot of `Study`. The frontend of the data portal prepends `https://identifiers.org/` tto each ID it finds in this slot.
```
marySalvi, line 49:
```suggestion
- **MassIVE Study** links are derived from the `massive_study_identifiers` slot of `Study`. The frontend of the data portal prepends `https://identifiers.org/` to each ID it finds in this slot.
```

docs/study_page_annotations.md
```diff
@@ -0,0 +1,49 @@
...
+
+## __Full Study Page__
+
+![Example study page](./images/study_page_1_annotated.png)
+
+### __Explanations__
+
+1. This comes directly from the `title` slot of `study` class instances
+2. `description` slot of `study` class instances
+3.  `profile_image_url` slot of the `principal_investigator` slot (type `PersonValue`) of `study` class instances OR the first entry in the `study_image` slot of `study` class instances. If both are present, priority is given to the `study_image` if both are present. During ingest, the image(s) are downloaded from the specified URI and stored as a `BLOB`s in Postgres. This is for stability reasons. We don't want to risk broken image links.
```
marySalvi, line 13:
```suggestion
3.  `profile_image_url` slot of the `principal_investigator` slot (type `PersonValue`) of `study` class instances OR the first entry in the `study_image` slot of `study` class instances. Priority is given to the `study_image` if both are present. During ingest, the image(s) are downloaded from the specified URI and stored as a `BLOB`s in Postgres. This is for stability reasons. We don't want to risk broken image links.
```

marySalvi (changes requested):
Just a couple of minor changes, otherwise looks good!

jeffbaumes (approved):
Looks great to me, thanks @naglepuff

turbomam:
This is great. I have short term requests directly related to this documentation PR and a related implementation request that can be dealt with or discussed further later on.

## Short term
- Let's use a standard notation for mentioning the slots and classes that must be traversed to get the values that appear on the Study pages. I would generally suggest dotted notation over slots only like 7) `has_credit_associations.applies_to_person.name`. If there are some conditional rules, I guess that might notation might not suffice. If it's important to emphasize the classes that are traversed across those paths, maybe we could do something like `[Study].has_credit_associations.[CreditAssociation].applies_to_person.[PersonValue].name` or `Study--has_credit_associations-->CreditAssociation--applies_to_person-->PersonValue--name`. Note that almost all NMDC slots are in PascalCase. It would be nice if that page used the same case.
- Could we please list an entry point for all URL/DOI modifications in steps like 12)? Like you have already for 14)?

## Implementation
- Let's discuss the advantages and disadvantages of URL/DOI modification being hard coded in nmdc-server versus getting that information from the schema. Here's a link for those [prefix expansions that are defined in nmdc.yaml](https://github.com/microbiomedata/nmdc-schema/blob/d2dcdfaedd7244e5db6836e5a71c2e651ed1834b/src/schema/nmdc.yaml#L40-L62). There are other prefixes defined in other modules, too. You can get a complete report with `gen-prefix-map src/schema/nmdc.yaml`
- If there are errors or missing prefixes, let's fix them in the schema.

turbomam (changes requested):
Thanks. I just entered some notes on the conversation tab. I guess they won't really be associated with this change request.

naglepuff:
> Thanks. I just entered some notes on the conversation tab. I guess they won't really be associated with this change request.

@turbomam Take a look at the most recent commits. I attempted to add something resembling a standard notation. As long as what I have is coherent, I think we should get this PR merged, and we can iterate a notation in the future for subsequent documentation updates that will explain other pages like the sample detail page. I also added a couple more links to the code where we create links.

marySalvi (approved):

turbomam (approved):

PR #987, opened 2023-07-03
Update nmdc-schema package to 7.7.0

naglepuff:
This brings the data portal in line with mongo. Local ingest is completing successfully after changing only the package number. The shape of the data still appears to be compatible with ingest code.


naglepuff:
Today I tried ingesting from dev mongo and got errors regarding certain data objects because they didn't have `file_size_bytes` populated. This isn't required  by the schema, so this PR should be updated to include changing the column to be nullable.

naglepuff:
From Alicia:

> For neon we will have data objects that don’t have a file size, those will be for raw sequencing data and will have a md5sum

naglepuff:
This PR ended up being a bit more involved due to some test data that exposed issues with the `nmdc-server` data model. 

### DataObject.file_size_bytes
Our `DataObject` model required column `file_size_bytes` to not be null. This was incompatible with how the `nmdc-schema` is defined, and with upcoming NEON data objects that won't have this property we needed this to change. This changed some of our query logic, such as coalescing NULL to 0 for aggregations. One drawback is that we can't report the exact size of data objects available to download.

### Biosample.lat_lon
Similarly, the dev database had Biosample documents without `lat_lon`. Our database required these, but they are not required per the `nmdc-schema`. This is to support legacy data that might not have values for latitude and longitude. Now, they are optional for our data model as well.

### OmicsProcessing.omics_type
During ingest, we use a bitstring to encode which omics_types can be associated with each Biosample. We use a custom SQL function (defined in `database.py`) which has hard-coded omics types (e.g. "Metagenome"). New data contains values like "metagenome," which don't match, resulting in failure to populate the `omics_types` for some biosamples. Now, during ingest, we map values from mongo to the values we check for in that SQL function (basically we make all capitalization uniform). See commit 6edb5ea

### Processed samples
#991 was merged into this branch. It changes how we model the relationship between `OmicsProcessing` and `Biosample`. In particular, `OmicsProcessing` objects can now have more than one `Biosample` input. Our search logic and testing suite was updated to account for this change. Front-end changes associated with this will come in a future PR.

nmdc_server/models.py
```diff
@@ -329,8 +343,11 @@ class OmicsProcessing(Base, AnnotatedModel):
 
     add_date = Column(DateTime, nullable=True)
     mod_date = Column(DateTime, nullable=True)
-    biosample_id = Column(String, ForeignKey("biosample.id"), nullable=True)
-    biosample = relationship("Biosample", backref="omics_processing")
+    biosample_inputs = relationship(
+        "Biosample", secondary=biosample_input_association, back_populates="omics_processing"
+    )
+    # biosample_id = Column(String, ForeignKey("biosample.id"), nullable=True)
+    # biosample = relationship("Biosample", backref="omics_processing")
```
jeffbaumes, line 350: Remove dead code

nmdc_server/ingest/biosample.py
```diff
@@ -31,6 +31,10 @@ def extract_extras(cls, values):
                 lat, lon = values.pop("lat_lon")["has_raw_value"].split(" ")
                 values["latitude"] = float(lat)
                 values["longitude"] = float(lon)
+        else:
+            # Workaround/testing code -- not to be included in PR
```
jeffbaumes, line 35: Should this code be in this PR? Either remove the else clause or the comment.

jeffbaumes (changes requested):
Note that this is a code sanity check, I did not test locally apart from the pooling schema changes. A couple questions on code inclusion but otherwise looks good.

nmdc_server/models.py
naglepuff: d65287e

nmdc_server/ingest/biosample.py
naglepuff: d65287e

jeffbaumes (approved):

PR #992, opened 2023-07-14
Upgrade to submission schema 7.7.2

pkalita-lbl:
* Upgrades the dependency
* Add migration to address changed enum values


naglepuff (changes requested):
Overall looks good, but I think you'll need to update your migrations slightly to account for other revisions. The current most recent revision is: https://github.com/microbiomedata/nmdc-server/blob/main/nmdc_server/migrations/versions/af8b2e3c91b2_biosample_omics_association.py

naglepuff (approved):

PR #1000, opened 2023-07-20
Allow parsing less specific dates

### LINKED ISSUE #999: Ingest: validation errors when parsing some datestrings from mongo
Some biosamples have a collection date without a day (e.g. `2018-07`). Ingest fails to parse these as valid `Date`s or `DateTime`s. 

These dates are valid per the schema. One solution is to interpolate days/months when they are missing. `2018-07 -> 2018-07-01`

naglepuff:
Previously, biosample collection dates without a month or year would not be ingested. This interpolates months/days as needed
Fix #999


naglepuff:
Upon further investigation, the first commit to this branch breaks datetime validation for other collection dates. I'm currently investigating further

PR #1005, opened 2023-07-24
Use aliased tables for study search

### LINKED ISSUE #1002: lat/long filter bug in data portal dev
There is a new? bug where the latitude and longitude active query terms does not correctly filter the studies. When I search around Alaska in data portal prod I get 3 studies, when I search this in data portal dev it shows all 18 studies so is not applying any filters. study_set id `gold:Gs0114663` for example does not have any samples from this region. When you add a filter for that study only it shows 0 samples.

product: data portal
browser: chrome Version 112.0.5615.137
env: dev
query:
https://data-dev.microbiomedata.org/?q=Ci0IARAGGAQiJVs1OS4zMTA3Njc5NTYwMzg4NCw3Mi4wNDY4Mzk4OTM3OTM5N10KMQgBEAcYBCIpWy0xODYuNjc5Njg3NTAwMDAwMDMsLTExMC4wMzkwNjI1MDAwMDAwMV0=

Screenshot 2023-07-21 at 4 18 37 PM

correct behavior in prod
Screenshot 2023-07-21 at 4 21 27 PM

It would be good to have this fixed for ESA for the data portal exercise so adding to this sprint.

naglepuff:
Fix #1002 

#991 did not properly update SQLAlchemy code to join `biosamples` through to `omics_processing` records, resulting in some cartesian products which reduced the functionality of aggregating on omics type.

Additionally study filtering was broken due to multiple references to tables in generated SQL code, so aliases are used for `biosample` and `omics_processing` in order to eliminate this error.

Now, searching using the data portal should be functional, and omics types aggregations and studies should update as expected.

### DIFF
--- a/nmdc_server/filters.py
+++ b/nmdc_server/filters.py
@@ -198,10 +198,7 @@ class BiosampleFilter(BaseFilter):
     table = Table.biosample
 
     def join_omics_processing(self, query: Query) -> Query:
-        return self.join_self(
-            query.join(models.OmicsProcessing, models.Biosample.omics_processing),
-            Table.biosample,
-        )
+        return query.join(models.biosample_input_association).join(models.Biosample)
 
     def join_biosample(self, query: Query) -> Query:
         return self.join_self(query, Table.biosample)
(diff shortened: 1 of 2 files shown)


marySalvi (approved):
LGTM!

PR #1006, opened 2023-07-25
Indicate pooled sample inputs on `DataObjectTable`s

### LINKED ISSUE #975: data portal ui updates to support pooled soil samples
Base on the infrastructure meeting we had the following ideas about how the data portal UI could support biosamples which are pooled for further processing for ESA
- [ ] on biosample pages show or link to related samples
- [ ] for WorkflowExecutionActivity records show/link to other contributing samples 

Depends on https://github.com/microbiomedata/issues/issues/260

naglepuff:
Fix #975 (partial)

This PR adds related Biosample inputs to the search results page. That is, if workflows had multiple biosample inputs (e.g. a pooled sample was used as the input), it will be indicated in the data object data for those biosamples:

image

Note that this PR does not update the biosample detail page.

### DIFF
--- a/web/src/components/DataObjectTable.vue
+++ b/web/src/components/DataObjectTable.vue
@@ -6,7 +6,7 @@ import { flattenDeep } from 'lodash';
 
 import { DataTableHeader } from 'vuetify';
 import { humanFileSize } from '@/data/utils';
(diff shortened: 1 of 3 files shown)


marySalvi (approved):
Looks good

PR #1009, opened 2023-07-25
Add related biosample to detail page

### LINKED ISSUE #1008: Show pooled biosample on biosample detail page
Related to #975

Separating out this part into its own issue

naglepuff:
This appears to be a first. We are displaying some data on a detail page that doesn't directly come from the object itself. Because of this I opted to not use `AttributeItem`.

Now, when applicable, a new section appears on the biosample detail page that contains a list of related biosamples. This list is built by looking at related `omics_processing` records, and noting what additional biosamples are used as inputs to those.

Fix #1008 

image

### DIFF
--- a/web/src/components/Presentation/AttributeList.vue
+++ b/web/src/components/Presentation/AttributeList.vue
@@ -58,12 +58,33 @@ export default defineComponent({
       return ret;
     });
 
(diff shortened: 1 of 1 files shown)


marySalvi (approved):
This looks good, I agree with splitting it out from the attributes

Issue #1023, opened 2023-07-27
Investigate ways to indicate future data


naglepuff:
Many UI elements are counts use `omics_processing` records . However, they existence of an `omics_processing` record does not always indicate the existence of real data that can be downloaded. 

We should try and update the UI in order to make it clear to end users that this is the case.

Some ideas:
- Disable certain UI elements (data object tables) if there are no data objects (Small-Medium)
- Keep separate counts of `omics_processing` records with available and unavailable data (Large-XLarge)

aclum:
This came up as high priority in user research with @ljohnson09, some users were confused by the greyed out tabs.

ljohnson09:
@aclum Does this also relate to "Clarify what "sample" numbers means appearing across pages"?

PR #1027, opened 2023-07-28
1023 disable empty data tables

marySalvi:
partially handles #1023, this is admittedly an easy, incomplete fix

the button for expanding the data object table is disabled if there is no processed data.


naglepuff (changes requested):
Could you merge main into this branch/rebase off main?

naglepuff (approved):

PR #1029, opened 2023-08-01
Count disctinct omics IDs

### LINKED ISSUE #1028: Double counting certain omics processing records
A single omics_processing records can be informed by multiple biosamples, so our query code needs to be updated to accommodate that.

naglepuff:
Fix #1028 

Chip counts for omics processing types in the study portion of the data portal were incorrect. For NEON, the number of `omics_processing` records related to that study is 1473, which was reflected properly by the omics bar chart, but not by the chip count in the study section (which reported 4111).

We were double-counting `omics_processing` records that had multiple `biosample` inputs.

For the study queries, we now correctly count *distinct* `omics_processing` IDs when aggregating for the chip counts.

### DIFF
--- a/nmdc_server/query.py
+++ b/nmdc_server/query.py
@@ -570,7 +570,7 @@ class StudyQuerySchema(BaseQuerySchema):
         query = (
             db.query(
                 op_summary_alias.annotations["omics_type"].astext.label("omics_processing_type"),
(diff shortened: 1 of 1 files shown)


marySalvi (approved):
LGTM, did a fresh ingest with the updated records and numbers match as expected

PR #1036, opened 2023-08-14
Add ORCID integration info to login button

### LINKED ISSUE #970: ORCID Integration Information
From ORCID feedback:

> As a part of the authentication login integration screen, include information as to how/why you will be connecting to ORCID and using their information; essentially, provide an explanation of the integration and how/why you are collecting ORCID iDs.

@shreddd  how do you interpret this feedback from the email? It sounds like we need to add an explanation to the oauth login screen, but I'm not sure how to do this. Unless you have a different idea or better understanding of this, I think it would be good to get some clarification from ORCID and maybe some examples of websites that comply with this.

> If you are writing to records, include an explanation of your integration and the benefits of connecting your information to the researcher’s ORCID record, in particular, the auto-update feature if applicable

Currently, we don't write to ORCID records.

>  If you use webhooks (the synchronize feature), include an explanation of your integration and the benefits of synchronization.

We don't use webhooks either.

> If you are reading records, include an explanation of your integration and the benefits of collecting information from the researchers record.

> Specify Scopes used to write on Record: activities/update and/or person/update.

We know why we're collecting information, and we know we want to use the read-limited scope.

From the initial demo meeting with ORCID, I remember discussing adding some kind of FAQ page to the data portal with this information would suffice. I think the next steps would be to get clarification that this is a good approach, and then design and implement something to fill this need, ensuring that it is easily discoverable for users.

naglepuff:
Fix #970 

This PR moves us closer to ORCID integration compliance by adding some information about how we intend to use users' ORCID IDs to the login workflow.

Now, if a user is logged out, hovering over the login button will display a modal that has some information about our use of ORCID.

### DIFF
--- a/web/src/components/Presentation/AuthButton.vue
+++ b/web/src/components/Presentation/AuthButton.vue
@@ -50,20 +50,63 @@ export default defineComponent({
       </v-btn>
     </template>
     <template v-else>
-      <v-btn
-        :text="!nav"
-        :plain="nav"
-        :small="nav"
-        href="/login"
+      <v-menu
+        bottom
+        max-width="500"
+        :open-on-hover="true"
+        transition="fade-transition"
+        offset-y
+        content-class="login-btn-orcid-help"
+        class="login-btn-orcid-help"
       >
-        <img
-          width="28px"
-          class="mr-2"
-          alt="OrcId login"
-          src="https://orcid.org/assets/vectors/orcid.logo.icon.svg"
-        >
-        OrcID Login
-      </v-btn>
+        <template #activator="{ on, attrs }">
+          <v-btn
+            :text="!nav"
+            :plain="nav"
+            :small="nav"
+            href="/login"
+            v-bind="attrs"
+            v-on="on"
+          >
+            <img
+              width="28px"
+              class="mr-2"
+              alt="OrcId login"
+              src="https://orcid.org/assets/vectors/orcid.logo.icon.svg"
+            >
+            OrcID Login
+          </v-btn>
+        </template>
+        <v-card>
+          <v-card-title>ORCID Account Integration</v-card-title>
+          <v-card-text>
+            <p>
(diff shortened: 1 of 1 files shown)


web/src/components/Presentation/AuthButton.vue
```diff
@@ -50,20 +50,63 @@ export default defineComponent({
...
+              already have one, to sign into your ORCID account, then granting permission to get your
+              ORCID ID. We do this to ensure that you are correctly identified and securely connected to
+              your ORCID ID.
+            </p>
+
+            <p>Learn more about <a href="https://orcid.org/blog/2017/02/20/whats-so-special-about-signing">what's so special about signing in?</a></p>
+
+            <p>
+              Additionally, we will use information from your ORCID record, like your name, to connect you with your NMDC submissions.
+            </p>
```
naglepuff, line 100: Here is the first draft of the text. Reviewers, please give feedback on this. If you need more context about this, reach out to me and I can provide that.

marySalvi, line 100:
```suggestion
            <p>
              NMDC requires an ORCID ID to log in. When logged in you have access to features such as downloading 
              files and the ability to create and manage metadata submissions through
              our Submission Portal.
            </p>
            <p>
              Click the "ORCID Login" button, to either register for an ORCID ID or, if you
              already have one, to sign into your ORCID account, then grant permission for NMDC to access your
              ORCID ID. This allows us to verify your identity and securely connect to
              your ORCID ID.
            </p>

            <p>Learn more about <a href="https://orcid.org/blog/2017/02/20/whats-so-special-about-signing">what's so special about signing in?</a></p>

            <p>
              Additionally, we will use information from your ORCID record, such as your name, to associate you with your NMDC submissions.
            </p>
```




Feel free to change/disregard.

shreddd (approved):

PR #1056, opened 2023-09-20
Fix results race condition

### LINKED ISSUE #1034: Issue when copying a data portal URL query from one google chrome tab to another
I discovered what I believe is a separate bug with query results when debugging the load balancing issues today.
When I run a query, say select the neon study in one search window (ie https://data.microbiomedata.org/?q=Ch4IABABGAMiFiJubWRjOnN0eS0xMS0zNHhqMTE1MCI=) I get the expected results in terms of number of samples. Expected number of active search results is 4204.
If I copy the url with the applied filter (https://data.microbiomedata.org/?q=Ch4IABABGAMiFiJubWRjOnN0eS0xMS0zNHhqMTE1MCI=) to a new tab in google chrome and press enter the active query search results starts as 0, this is expected as the search is loading, then it momentarily shows the correct result of 4204, then it updates to 6734 which is the total number of biosamples in the data portal. The incorrect results seems to be limited to number of samples and the active search query results. The upset plot, map, and omics bar chart filter correctly.
Screenshot 2023-08-08 at 10 26 05 PM

Chrome version Version 115.0.5790.170
This is reproducible in safari Version 16.1 (18614.2.9.1.12). The first time I copy the query into a safari tab the results are correct. Subsequent tabs produce the same incorrect behavior as chrome, sometimes skipping over monetarily having the correct 4204 value. 

cc @mslarae13 @pkalita-lbl @shreddd @jeffbaumes

marySalvi:
fixes #1034

### DIFF
--- a/web/src/use/usePaginatedResults.ts
+++ b/web/src/use/usePaginatedResults.ts
@@ -54,7 +54,6 @@ export default function usePaginatedResult<T>(
   if (dataObjectFilter !== undefined) {
     watch(dataObjectFilter, fetchResults, { deep: true });
   }
-  fetchResults();
   // ENDTODO
 
   function setPage(newPage: number) {


naglepuff (approved):
I think this did the trick. To test, I compared this branch with the current dev code. I don't have the same data locally that exists in dev, but I was more interested in looking at network activity anyway, so I don't think it matters. I tested by running a search to get a URL with that search encoded, and then pointing my browser at that URL.

In dev, I see (and I think @jeffbaumes pointed this out in the original issue) two pairs of requests to the `/search` endpoints for biosample and study, which was the cause of the race condition:
image

With this branch, I only see a single pair of requests, and I see the expected results on the data portal:
image

PR #1066, opened 2023-10-12
Rename data portal API

naglepuff:
`NMDC Dataset API` -> `NMDC Data and Submission Portal API`

@aclum @jeffbaumes any objections to this change? This affects the title at the top of the swagger page. I think the proposed title better reflects what the user is actually looking at. I am open to any tweaks as reviewers see fit.

image


aclum:
Fine w/ me.

jeffbaumes (approved):

PR #1067, opened 2023-10-18
Check for pi before instantiating component

marySalvi:


marySalvi:
@naglepuff This is draft because I was unsure about if the `team` slot was dependent on a PI or not. This fix just moves everything up if there is no PI but I can make a more elegant solution in the refresh PR.

naglepuff (approved):
LGTM!

PR #1068, opened 2023-10-18
Allow searching text for study name and description

### LINKED ISSUE #965: Option to search by name and description for study in the data portal
Search based on and slots  `title` (title is what is displayed in the data portal, not `name`, and `description` for `study_set`

### LINKED ISSUE #1071: NMDC custom sql functions don't get created when expected
NMDC ingest and search relies on some custom functions defined [here](https://github.com/microbiomedata/nmdc-server/blob/5072499f0afeac2e1b8fcef76ded7ab20697aefe/nmdc_server/database.py#L38-L43).

These functions appear to be created or updated before the creation of the schema using the SQLAlchemy event system. Since we don't use SQL alchemy to create the schema (instead, doing that in our `prestart` script), these functions don't get created for fresh databases, and they don't get updated for existing ones.

There should be a different, more consistent way to update these functions.

naglepuff:
Fix #965 
Fix #1071 

This PR adds some additional query options when using the text search bar, as well as establishes a framework for including more in the future. For this iteration, we have added plaintext searching for:

`title` of `Study`
`description` of `Study`
`name` of `Study`

In order to accomplish this, our DSL query language has been updated to support using the SQL `ilike` function. This was a straightforward change for properties represented directly as columns on the `study` table, but required some additional work for values stored in the `annotation` column (JSON).

There is also a new CLI command, `nmdc-server create-or-replace-nmdc-functions` which updates the custom NMDC functions used for ingest and search at a whim, rather than relying on the SQLAlchemy event system. See linked issue #1071 for some additional details.

Updating these functions is also folded into the `nmdc-server migrate` command, which is invoked in our prestart script. This should populate the needed function when the app spins up.

### DIFF
--- a/nmdc_server/database.py
+++ b/nmdc_server/database.py
@@ -34,12 +34,8 @@ metadata = MetaData(
 
 Base = declarative_base(metadata=metadata)
 
-# define functions used for comparison between arbitrary types
-listen(
-    metadata,
-    "before_create",
-    DDL(
-        """
+update_nmdc_functions_sql = DDL(
+    """
 create or replace function nmdc_compare (lhs text, op text, rhs text)
 returns boolean as $$
     declare
@@ -55,6 +51,7 @@ returns boolean as $$
                 when op = '>' then (lhs_ > rhs_)
                 when op = '>=' then (lhs_ >= rhs_)
                 when op = '!=' then (lhs_ <> rhs_)
+                when op = 'like' then (lhs_ ILIKE rhs_)
             end into result;
         return result;
     end;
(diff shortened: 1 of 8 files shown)


nmdc_server/database.py
```diff
@@ -128,8 +125,14 @@
...
-"""
-    ),
+    """
+)
+
+# define functions used for comparison between arbitrary types
+listen(
+    metadata,
+    "before_create",
+    update_nmdc_functions_sql,
```
jeffbaumes, line 135: So you are leaving in this listen even if we don't know if it's ever triggered. Sounds reasonable and a safe choice just to make sure we aren't breaking existing functionality. At worst we may now be sometimes updating functions twice but it's a very short and reentrant operation so that's ok.

jeffbaumes (approved):
I left a comment about the existing `listen` statement but this all looks great to me.

nmdc_server/database.py
naglepuff: Yes, that sums up my reasoning as well.

PR #1069, opened 2023-10-19
Show total dataset and selected dataset on chart

### LINKED ISSUE #928: Data portal collection date not filtering when study selected
collection date not filtering when a study is selected. This doesn't work on prod or dev. 

GROW for example there was a bug in GOLD so collection date isn't available for any biosamples.

marySalvi:
Closes #928


naglepuff (approved):
LGTM! This is great

PR #1074, opened 2023-10-21
Fix inaccuracy in API endpoint name

eecavanna:
I put "processings" in an attempt to pluralize the word "processing" (which I think is being used as a noun in this context).


eecavanna:
All good! I haven't been thinking about it.

FYI: I am not authorized to merge the branch (which is OK with me). Screenshot:

image

PR #1084, opened 2023-11-29
Change ingest code to include new DOI Class

### LINKED ISSUE #1010: Support DOI overhaul - multivalued award DOIs
`nmdc-schema` is retiring the DOI slot in favor of more specific, multivalued slots e.g. `award_dois`.

The data portal should be updated to reflect this.

A rough outline of the required changes are:

- [ ] Retire column `doi` of table `study`
- [ ] Model the new relationships with SQLAlchemy
- [ ] Update ingest to populate the new `study_award_doi_association` table
- [ ] Update the web server schemas (`Pydantic` and `typescript`) 
- [ ] Update the study landing page

marySalvi:
Closes #1010

### DIFF
--- a/nmdc_server/ingest/study.py
+++ b/nmdc_server/ingest/study.py
@@ -9,7 +9,7 @@ from sqlalchemy.orm import Session
 from nmdc_server.crud import create_study, get_doi
 from nmdc_server.ingest.common import extract_extras, extract_value
 from nmdc_server.ingest.doi import upsert_doi
-from nmdc_server.models import DOIType, PrincipalInvestigator
+from nmdc_server.models import PrincipalInvestigator
 from nmdc_server.schemas import StudyCreate
 
 
@@ -65,14 +65,13 @@ def load(db: Session, cursor: Cursor):
             obj["principal_investigator_id"] = get_or_create_pi(db, pi_name, pi_url, pi_orcid)
             obj["principal_investigator_websites"] = obj.pop("websites", [])
         obj["image"] = get_study_image_data(obj.pop("study_image", []))
-        dois=obj.pop("associated_dois", None)
+        dois = obj.pop("associated_dois", None)
         if dois:
             for doi in dois:
                 doi["doi_value"] = transform_doi(doi.pop("doi_value"))
 
             for doi in dois:
                 upsert_doi(db, doi["doi_value"], doi["doi_category"])
-
 
         new_study = create_study(db, Study(**obj))
         if dois:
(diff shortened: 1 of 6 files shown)


marySalvi:
Because of the Enum alteration this was especially tricky. Please test vigorously! Thank you!

nmdc_server/models.py
```diff
@@ -273,18 +273,6 @@ def open_in_gold(self) -> Optional[str]:
...
-        return [d for d in self.dois if d.doi_type == DOIType.AWARD]  # type: ignore
-
-    @property
-    def publication_dois(self) -> list[DOIInfo]:
-        return [d for d in self.dois if d.doi_type == DOIType.PUBLICATION]  # type: ignore
-
-    @property
-    def dataset_dois(self) -> list[DOIInfo]:
-        return [d for d in self.dois if d.doi_type == DOIType.DATASET]  # type: ignore
-
```
naglepuff, line 287: What's the reason for getting rid of these properties? I think they'd still work based on the changes you've made. Are you thinking its best to just leave it to consumers of the API to sort out all the DOIs?

nmdc_server/models.py
```diff
@@ -273,18 +273,6 @@ def open_in_gold(self) -> Optional[str]:
...
     @property
     def doi_map(self) -> Dict[str, Any]:
```
naglepuff, line 277: Related to the above changes, wouldn't the construction of the `doi_map` need to be updated now that there aren't `self.<doi_type>_dois` properties?

nmdc_server/ingest/study.py
```diff
@@ -65,27 +65,18 @@ def load(db: Session, cursor: Cursor):
...
 
-        publication_dois = [transform_doi(d) for d in obj.pop("publications", [])] + [
-            transform_doi(d) for d in obj.pop("publication_dois", [])
-        ]
-        award_dois = [transform_doi(doi) for doi in obj.pop("award_dois", [])] + [
-            transform_doi(d) for d in obj.pop("emsl_project_dois", [])
-        ]
-        dataset_dois = [transform_doi(doi) for doi in obj.pop("dataset_dois", [])]
+            for doi in dois:
+                upsert_doi(db, doi["doi_value"], doi["doi_category"])
```
naglepuff, line 74:
```suggestion
                upsert_doi(db, doi["doi_value"], DOIType(doi["doi_category"]))
```

Since `upsert_doi` expects a `DOIType` as the last argument, I'd recommend either changing the signature of that method to `str` or just doing the enum value construction here.

naglepuff (changes requested):
Just a couple of thoughts on the code changes so far. Nothing too crazy. I'll start doing a local ingest test in the meantime.

nmdc_server/models.py
marySalvi: Yes, that was my thought

nmdc_server/models.py
marySalvi: Yes, you are right. I though I had updated it I think I might have lost that on rebase. I'll either recover or re-write it

marySalvi: fixed in 2a7c9385d07f653b2cd5c2d902529fac30c6f2f0

naglepuff:
Also if this PR will resolve an issue please link it @marySalvi

PR #1091, opened 2023-12-04
Fix DOI links and move essdive and massive dois back

marySalvi:


naglepuff (approved):
LGTM! I see ESS and MassIVE DOIs back under additional data, and the DOI links are now working in the gray right hand area.

PR #1097, opened 2023-12-07
Fix XLSX import when columns are reordered or missing

### LINKED ISSUE #1047: submission portal columns map incorrectly when importing a spreadsheet
Someone reported this to @emileyfadrosh in the workshop this morning and I was able to reproduce the problem in chrome using the 'soil' DH package. 
I filled out some metadata, clicked the download xlsx button, moved the gold ecosystem pathway columns to be before the environmental context terms, filled out some of those fields, then clicked import xlsx file and it imports the data without taking into account the column headers. Example file for import is attached. The correct behavior would be for the fields to be populated based on matching up the column headers, so `broad-scale environmental context `should be populated with
  `terrestrial biome [ENVO:00000446]` not `test`

[nmdc_sample_export test_mixed up rows.xlsx](https://github.com/microbiomedata/nmdc-server/files/12434108/nmdc_sample_export.test_mixed.up.rows.xlsx)

cc @pkalita-lbl

pkalita-lbl:
Fixes #1047 (also related to https://github.com/microbiomedata/issues/issues/543)

### DIFF
--- a/web/src/views/SubmissionPortal/HarmonizerView.vue
+++ b/web/src/views/SubmissionPortal/HarmonizerView.vue
@@ -361,13 +361,13 @@ export default defineComponent({
           // IDs. This code reads the worksheet data and remaps the keys from slot names to IDs.
           const slotIdToNameMap = harmonizerApi.getHeaderRow(template.schemaClass);
           const slotNameToIdMap = Object.fromEntries(Object.entries(slotIdToNameMap).map(([k, v]) => [v, k]));
-          const worksheet_data: Record<string, string>[] = utils.sheet_to_json(worksheet);
-          const remapped_data = worksheet_data.map((row) => Object.fromEntries(Object.entries(row)
-            .filter(([column]) => slotNameToIdMap[column] !== undefined)
-            .map(([column, value]) => [slotNameToIdMap[column], value])));
+          const worksheetData: Record<string, string>[] = utils.sheet_to_json(worksheet);
+          const remappedData = worksheetData.map((row) => Object.fromEntries(Object.entries(row)
+            .filter(([slotName]) => slotNameToIdMap[slotName] !== undefined)
+            .map(([slotName, value]) => [slotNameToIdMap[slotName], value])));
 
           imported[template.sampleDataSlot] = harmonizerApi.unflattenArrayValues(
-            remapped_data,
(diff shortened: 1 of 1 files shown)


web/src/views/SubmissionPortal/HarmonizerView.vue
```diff
@@ -355,11 +355,19 @@ export default defineComponent({
...
           }
+
+          // The spreadsheet has slot names as the header row. So `sheet_to_json` will produce array
+          // of objects with slot names as keys. But we want the imported data to be keyed on slot
+          // IDs. This code reads the worksheet data and remaps the keys from slot names to IDs.
+          const slotIdToNameMap = harmonizerApi.getHeaderRow(template.schemaClass);
+          const slotNameToIdMap = Object.fromEntries(Object.entries(slotIdToNameMap).map(([k, v]) => [v, k]));
+          const worksheet_data: Record<string, string>[] = utils.sheet_to_json(worksheet);
+          const remapped_data = worksheet_data.map((row) => Object.fromEntries(Object.entries(row)
+            .filter(([column]) => slotNameToIdMap[column] !== undefined)
```
eecavanna, line 366: A question arose for me while reading this code in isolation, and without being familiar with the rest of the file: I infer, from the previous line, that `row` is not a scalar. Is `column` here a column (group of "cells") or a column name (scalar)? If the latter, I think I'd prefer the variable name `columnName`.

eecavanna (approved):
I left one comment about a variable name. I don't consider it a deal-breaker (i.e. I'm OK with this being merged in as is).

web/src/views/SubmissionPortal/HarmonizerView.vue
eecavanna: Fixed in https://github.com/microbiomedata/nmdc-server/pull/1097/commits/50ca68278db99acd755c82d3584137f82c866365

pkalita-lbl: Thanks, good points. Even using "column" at all here is a little ambiguous here, so I've updated to use `slotName`. I also updated `worksheet_data` and `remapped_data` to be camel case. Whoops. Forgot I wasn't in Python anymore 🤦🏻

Issue #1079, opened 2023-11-03
Add cron jobs to do automatic ingest


pkalita-lbl:
As proposed in the Release Management Squad, it would be useful to do periodic, automatic ingests from Mongo. This would have a few benefits:

* On dev (assuming ingest is done from the dev Mongo deployment -- which is not currently the case!) a failing daily ingest would give us an earlier heads-up if the ingest code needs to be updated to accommodate recent runtime changes.
* On production (assuming ingest is done from the production Mongo deployment which is more stable see https://github.com/microbiomedata/nmdc-runtime/issues/357) a weekly ingest would alleviate the need for team members to make and respond to one-off requests for ingest. 

Most of the ingest is already pretty well-automated (either via the API or CLI). The manual steps currently are doing the A/B database swap and restarting the service in Rancher. I assume those can be automated via the Rancher CLI.

eecavanna:
I ran into some User ID-related issues when trying to set up a CronJob recently. Here's a link to my notes about it: https://github.com/microbiomedata/infra-admin/issues/27#issuecomment-1863740052

ssarrafan:
Sprint over, no updates, removing from sprint and adding backlog label.  If still active please move to new sprint.

PR #1103, opened 2024-01-03
Update ingest CLI to support scheduled ingest

pkalita-lbl:
Re: https://github.com/microbiomedata/nmdc-server/issues/1079

These changes are in support of having a cron job in Rancher for periodic automated ingests. The idea is that cron job will use the existing `server` Docker image to do an ingest via the existing `nmdc-server` CLI. Typically when we initiate an ingest manually we do so via the API with a `POST /api/jobs/ingest` request. I'm proposing _not_ using the API for the automated ingest because 1) the `/api/jobs/ingest` endpoint requires authentication and that would be challenging without persistent API authentication tokens and 2) we would probably need to add an additional endpoint to check on the job status after is has been finished in order to poll for success or failure. 

The changes here move the guts of the `ingest` celery task to a separate plain function (`do_ingest`) that can be called by both the celery task and the CLI to ensure they do the same operations. It also adds a `--swap-rancher-secrets` to the ingest CLI. If used, it will use the Rancher API to do the A/B database swap and redeploy the server and worker workloads. A number of environment variables need to be set in order to use that flag.

### DIFF
--- a/nmdc_server/cli.py
+++ b/nmdc_server/cli.py
@@ -89,23 +89,26 @@ def ingest(verbose, function_limit, skip_annotation, swap_rancher_secrets):
             click.echo(id)
 
     if swap_rancher_secrets:
-        if not settings.rancher_api_base_url:
-            raise ValueError("rancher_api_base_url must be set to use --swap-rancher-secrets")
-        if not settings.rancher_api_auth_token:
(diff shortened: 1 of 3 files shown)


naglepuff (approved):
LGTM!

PR #1105, opened 2024-01-05
Display git-based version in swagger docs

### LINKED ISSUE #1092: Update Swagger UI to display application version identifier
For reference, `nmdc-runtime`'s Swagger UI shows this:

- Production environment: https://api.microbiomedata.org/docs

    image

- Development environment: https://api-dev.microbiomedata.org/docs

    image

I want `nmdc-server`'s Swagger UI to show something analogous (I think even showing a commit hash would be useful).

I think `nmdc-server`'s Swagger UI (https://data.microbiomedata.org/api/docs) always shows `0.1` there.

https://github.com/microbiomedata/nmdc-server/blob/910c92d431b75fe67b26d3de0768f5a47cacd121/setup.py#L5

https://github.com/microbiomedata/nmdc-server/blob/910c92d431b75fe67b26d3de0768f5a47cacd121/nmdc_server/app.py#L26

image

pkalita-lbl:
Fixes #1092 

Previously the `nmdc-server` package's version was hardcoded in `setup.py` at `0.1`. With these changes the version is determined automatically using [`setuptools_scm`](https://setuptools-scm.readthedocs.io/en/latest/). Using `setuptools_scm` requires migrating from the legacy `setup.py`-based build to using `pyproject.toml`. Since we weren't really doing anything particularly dynamic in `setup.py` that conversion was pretty straightforward. Since the build process now relies on git state, the `.git` directory is mounted during the Docker build.

I also took the opportunity to get rid of `setup.cfg` as well. It was only holding configuration that could be moved elsewhere:
* `isort` config is now in `pyproject.toml`. Going forward I think it's good practice to put as much tooling configuration as possible in `pyproject.toml`.
* `flake8` was already being configured in _both_ `setup.cfg` and `tox.ini`, so that's been consolidated into `tox.ini`. Sadly they sort of refuse to support `pyproject.toml`; might be a good motiviation to switch to [ruff](https://docs.astral.sh/ruff/) at some point.
* `mypy` says they support `pyproject.toml` but I found it to be buggy so I've moved that config to a separate `mypy.ini` file for now.


naglepuff (approved):
LGTM. Welcome to the future, `nmdc-server`!

eecavanna:
Awesome! Thanks for implementing this! (I just noticed this PR)

PR #1106, opened 2024-01-06
Reconfigure deploy workflow to use main/latest/version number Docker tags

pkalita-lbl:
Fixes https://github.com/microbiomedata/issues/issues/554

These changes update the `deploy.yml` workflow file to:

* Tag Docker image with `main` for pushes to the `main` branch. Tag Docker image with `prod` and `X.Y.Z` for pushes to tags like `vX.Y.Z`. This is accomplished by using the `docker/metadata-action` action.
* Redeploy the Rancher `nmdc-dev` workloads on pushes to `main` and `nmdc` workloads on pushes to version tags. Previously only `nmdc-dev` workloads were redeployed on pushes to main. 

Once these changes are in place I'll update Rancher to point the `nmdc` workloads to the new `prod` tagged images. Then when a GitHub Release is created with a new version tag the deployment will happen automatically with no need to update anything in Rancher. This also brings this repository inline with what `nmdc-runtime` does.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,88 +1,87 @@
 name: deploy
 
-on: 
+on:
   push:
     branches:
       - main
-  release:
-      types: [published]
+    tags:
+      - 'v*'
+
+env:
+  # We don't want to do certain steps if this is running in a fork
+  IS_ORIGINAL_REPO: ${{ github.repository == 'microbiomedata/nmdc-server' }}
+
+  # Used to determine if we add the special `prod` Docker image tag
+  IS_PROD_RELEASE: ${{ startsWith(github.ref, 'refs/tags/v') }}
+
+  # Used when sending redeploy action requests to Rancher
+  RANCHER_NAMESPACE: ${{ startsWith(github.ref, 'refs/tags/v') && 'nmdc' || 'nmdc-dev' }}
 
 jobs:
   build:
     runs-on: ubuntu-latest
+
     strategy:
       matrix:
         image: [server, client, worker]
 
     steps:
-      -
-        name: Checkout code
-        uses: actions/checkout@v2
+      - name: Checkout code
+        uses: actions/checkout@v4
         with:
-          submodules: 'recursive'
-      -
-        name: Login to GitHub Container Registry
-        uses: docker/login-action@v1
+          # history for all branches and tags is needed for setuptools-scm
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -1,88 +1,87 @@
...
+        deployment: [ backend, frontend, worker ]
+
+    steps:
+      - name: Redeploy ${{ env.RANCHER_NAMESPACE }}:${{ env.IS_PROD_RELEASE && '' || 'portal-' }}${{ matrix.deployment }}
+        if: ${{ env.IS_ORIGINAL_REPO }}
+        uses: fjogeleit/http-request-action@v1
         with:
-          url: 'https://rancher2.spin.nersc.gov/v3/project/c-tmq7p:p-bkv45/workloads/deployment:nmdc-dev:portal-worker?action=redeploy'
-          method: 'POST'
+          url: https://rancher2.spin.nersc.gov/v3/project/c-tmq7p:p-bkv45/workloads/deployment:${{ env.RANCHER_NAMESPACE }}:${{ env.IS_PROD_RELEASE && '' || 'portal-' }}${{ matrix.deployment }}?action=redeploy
```
eecavanna, line 84: I see that this is where the prefix of `portal-` would be applied when needed (i.e. in the `nmdc-dev` namespace).

eecavanna (approved):

.github/workflows/deploy.yml
pkalita-lbl: Cool thanks. It'll be nice when they're all consistent, but no particular rush. We'll just have to remember to update this workflow again once that issue is resolved.

.github/workflows/deploy.yml
```diff
@@ -1,88 +1,87 @@
...
-        name: Login to GitHub Container Registry
-        uses: docker/login-action@v1
+          # history for all branches and tags is needed for setuptools-scm
+          fetch-depth: 0
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
```
naglepuff, line 39: What are these actions giving us? It doesn't look like the previous version of this action used these steps.

naglepuff (approved):
I just have one comment about the changes, which is more for educational purposes than anything else.

PR #1107, opened 2024-01-09
Used namespacedSecrets endpoint to get/update rancher secrets in ingest

pkalita-lbl:
Just a quick revision to #1103. I had been testing that locally against the "development" project (which we don't really use anymore). This updates the ingest CLI to use the correct endpoint for the "production" project where we used namespaced secrets.

### DIFF
--- a/nmdc_server/cli.py
+++ b/nmdc_server/cli.py
@@ -107,7 +107,7 @@ def ingest(verbose, function_limit, skip_annotation, swap_rancher_secrets):
         secret_url = (
             f"{settings.rancher_api_base_url}"
             f"/project/{settings.rancher_project_id}"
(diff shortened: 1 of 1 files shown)


naglepuff (approved):
LGTM, but I do have one question:

What are the odds we switch away from using namespaced secrets in the future? Coming back here to edit that seems like it could be tedious. Do you think its worth using an environment variable to handle the "secrets" vs. "namespacedSecrets" segment of the URL?

pkalita-lbl:
Fair question. I don't _think_ that's likely, but that's just based on vibes not any Rancher expertise. Maybe if it needs to be updated in the future we could switch to passing in full endpoints (e.g. `/project/xxxxx/someObject/yyyyy`) in the environment variables.

PR #1110, opened 2024-01-10
Ensure clean git working directory to get correct version number

pkalita-lbl:
This is a follow-up to #1105. `setuptools-scm` uses the git working directory state as part of its [default versioning scheme](https://setuptools-scm.readthedocs.io/en/latest/usage/#default-versioning-scheme). So to get the right version number we need to ensure that the working directory is clean before `pip install`-ing the `nmdc-server` package. Two things were making it unclean:

* The base image (`tiangolo/uvicorn-gunicorn:python3.9-2023-06-05`) includes an `/app/main.py` file. It's not used in any way; it just had some placeholder hello world code. That file needed to be removed.
* I needed to copy _all_ the checked-in files into the `/app` folder. Since that includes the `.git` directory itself it no longer needs to be mounted. Unfortunately I think this will make the image size larger. I don't know if that will be a problem in practice or not. If it does we might need to explore ways of customizing `seuptools-scm`'s behavior.

Also as a follow-on to #1106 I updated the image tags in `docker-compose.yml`

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,6 +1,9 @@
+.mypy_cache
 .tox
 .coverage
 .cache
+.venv
+.vscode
 env
 *.egg-info
 node_modules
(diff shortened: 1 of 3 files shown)


naglepuff (approved):
LGTM!

PR #1112, opened 2024-01-11
Model Metadata Submission roles

naglepuff:
This creates the infrastructure needed to enforce access control for metadata submissions. It is a home grown attempt to achieve this. I looked at packages like [oso](https://github.com/osohq/oso) and [casbin](https://github.com/casbin/casbin/), but the overhead to learn those systems and be efficient developing with them was not worth it at this time. It also appears like we could use those in the future for policy enforcement, so we aren't strictly locked out of third part tools in the future. We also have some special needs due to our storage of the submission metadata as json. Our permission model actually applies the contents of a single column in our database, and it is not simply row-based permissions.

Once again, I found the Sanbox Orcid environment useful for testing changes. I could create multiple users, and do some functional testing of the permissions by adding some users as PIs for other user's submissions, ensuring that those submissions appeared on the list, etc. I am happy to help set that up for anyone who would like to try.

### Summary of Changes
The changes here are strictly in backend code. Several files were touched for this PR. Some of the code added is demonstrative of intent, such as `crud.py::can_read`, which isn't used anywhere but shows how the permissions model can be used to implement additional permission checks at the API level.

#### New Model
To track permissions, the class/model/table `models.py::SubmissionRole` has been added. This table links a submission (via FK) to a user (via ORCID string). This way, permissions can be added for users who don't yet have a record in our `user_logins` table. The tradeoff is that this can lead to permissions that point to nonsense users (fake/wrong ORCID, etc). The permission levels are defined by the `SubmissionEditorRole` enumeration. See the permissions tab of [this spreadsheet](https://docs.google.com/spreadsheets/d/18tN1_KlQFHq1FFaBTT3zzusEn57tkidAfj-3nC5wDfM/edit#gid=235041042) for definitions.

#### New Database Operations
Several new functions have been added to `crud.py` to be used to enforce the permissions policy defined by the roles. In the future, this kind of check could be used as a wrapper function for api endpoints, or dependencies to be injected via FastAPI.

#### Some routes have changed slightly

- `GET /api/metadata_submission`: The "get all my submissions" endpoint now used a new function which combines the check for authorship with a check for any submissions that the logged in user has a role for. Admin access to all submissions regardless of role is unaffected.
- `POST /api/metadata_submission`: The endpoint for creating a new submission now has the added responsibility of creating an `owner` role for the creator of the submission.
- `GET /api/metadata_submission/{id}`: The endpoint for retrieving a single submission uses the `can_edit_all` check. If that check fails and the user is not an admin, a `403` is returned. This can and should be modified as additional roles are added.
- `PATCH /api/metadata_submission/{id}`: Updating a submission is gated by the `can_edit_add` check as well. Additionally, change to the `piOrcid` will be detected and create an `owner` role for that ORCID.


nmdc_server/crud.py
```diff
@@ -488,3 +489,90 @@ def release_submission_lock(db: Session, submission_id: str):
...
+    submission: Optional[models.SubmissionMetadata] = db.query(models.SubmissionMetadata).get(
+        submission_id
+    )
+    if submission is None:
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail="Submission not found")
+    is_creator = submission.author_orcid == user_orcid
+    return (is_creator or (role and models.SubmissionEditorRole(role.role) in read_roles)) is True
+
+
+def can_edit_all(db: Session, submission_id: str, user_orcid: str) -> Optional[bool]:
```
jeffbaumes, line 540: I'd either change the function name to be longer (probably preferred if possible) or add a comment describing the function intent. "can edit all" I think means they can edit the entire submission, maybe "can_edit_entire_submission"?

nmdc_server/crud.py
```diff
@@ -488,3 +489,90 @@ def release_submission_lock(db: Session, submission_id: str):
...
+    if user.is_admin:
+        return all_submissions
+
+    # Use a left outer join to include submissions with no roles for the user,
+    # so we can filter on author orcid
+    permitted_submissions = all_submissions.outerjoin(models.SubmissionRole)
+    permitted_submissions = permitted_submissions.filter(
+        or_(
+            models.SubmissionRole.user_orcid == user.orcid,
+            models.SubmissionMetadata.author_orcid == user.orcid,
```
jeffbaumes, line 575: So a particular role (author) is stored not in the roles table, but in the submission metadata itself. I can see the positive here of maintaining 1-1 naturally between authors and submissions by it being right in the submissions table. Some possible negatives are more complex logic like this, and people who can edit the entire submission can potentially also edit the author field (maybe in a hacked client?).

tests/test_submission.py
```diff
@@ -54,3 +65,106 @@ def test_try_edit_locked_by_current_user_submission(
...
+        )
+    else:
+        submission = fakes.MetadataSubmissionFactory()
+    db.commit()
+    response = client.request(method="get", url=f"/api/metadata_submission/{submission.id}")
+    assert response.status_code == code
+
+
+@pytest.mark.parametrize(
+    "role,code", [(SubmissionEditorRole.owner, 200), ("author", 200), (None, 403)]
```
jeffbaumes, line 119: Another consequence (I think) of author directly in the metadata is that the "author" role is not in the enum since it's not to be used in the role table, which is confusing. I wondered a while why one of these was an enum and one a string.

jeffbaumes (changes requested):
I have a couple of thoughts to consider.

nmdc_server/crud.py
naglepuff: Author isn't really a role -- it implies the `owner` role.

Its a consequence of existing submissions not having any roles associated with them now. I could possibly modify the migration script to include a data migration that creates an `owner` role for the author of each submission it finds.

I don't think there's a concern about editing the author field from the client, since that information is stored in a column on the `submission_metadata` table and the [endpoint that updates submissions](https://github.com/microbiomedata/nmdc-server/blob/069b63224a3abb732b95ef2f65293f968a660574/nmdc_server/api.py#L548-L577) only ever modifies the `metadata_submission` (jsonb) column.

tests/test_submission.py
naglepuff: If we can ensure that every submission has an `owner` role for its author, we could prevent the scenario that is being tested here (implicit ownership for authors without explicit roles). I agree that this is confusing. Minimally I could add a comment describing the meaning of the test case, but this could also potentially be fixed with a data migration.

nmdc_server/crud.py
jeffbaumes: > Author isn't really a role -- it implies the owner role.

Ah ok that makes more sense.

> I could possibly modify the migration script to include a data migration that creates an owner role for the author of each submission it finds.

If you did this, would it make sense to also remove the `SubmissionMetadata.author_orcid` field entirely in the migration?

tests/test_submission.py
jeffbaumes: Yeah, I have the gut feeling that implicit ownership seems less ideal than explicit ownership from a potential bug and maintenance point of view. It's not a huge problem but may make sense to clean it up now rather than later.

nmdc_server/crud.py
naglepuff: I updated this function name and `can_read` above in commit 92950c9

nmdc_server/crud.py
naglepuff: It might. I didn't add that field so I don't know if it used anywhere else. I can take a look at that as more development is done for more submission roles. This PR is already getting to be pretty large so I'm hesitant to remove that right away.

tests/test_submission.py
naglepuff: Tests were fixed up in d7b99f4.

I removed the "author" nonsense, but had to fix some earlier tests by creating explicit roles for the `logged_in_user`. I also updated the `admin_required` dependency (ba8f181) because the tests need to NOT bypass that check (that check was previously only actually done in production environments)

jeffbaumes (approved):

mslarae13:
@naglepuff  just confirming this is for edit permissions for the PI?
I just had @bmeluch & @brynnz22 test this out & Bea made Brynn the PI & she was able to view and edit the submission! Success!

naglepuff:
Correct!

PR #1116, opened 2024-01-18
Specify join for envo term when faceting omics_processing

### LINKED ISSUE #1115: Omics type bucketed facet bugged
Screenshot 2024-01-12 at 1 09 32 PM

The search from the above screenshot contains unexpected results in the omics bar chart. Running the search locally and looking at the logs reveals some warnings from `SQLAlchemy`:

```
backend_1  | /app/./nmdc_server/pagination.py:73: SAWarning: SELECT statement has a cartesian product between FROM element(s) "envo_term_2" and FROM element "biosample".  Apply join condition(s) between each element to resolve.                                                                                                                                 
backend_1  |   count = query.count()  
backend_1  | /app/./nmdc_server/pagination.py:73: SAWarning: SELECT statement has a cartesian product between FROM element(s) "envo_term_1" and FROM element "biosample".  Apply join condition(s) between each element to resolve.                                                                                                                                 
backend_1  |   count = query.count()  
```

Looking at the SQL generated by the search, and looking for the `envo_term_n` segments, I found the following section of the query:

```
SELECT
    distinct(omics_processing.id) AS id
FROM
    envo_term AS envo_term_1,
    omics_processing
    JOIN biosample_input_association ON omics_processing.id = biosample_input_association.omics_processing_id
    JOIN biosample ON biosample.id = biosample_input_association.biosample_id
WHERE
    envo_term_1.label = %(label_1) s
INTERSECT
SELECT
    distinct(omics_processing.id) AS id
FROM
    envo_term AS envo_term_2,
    omics_processing
    JOIN biosample_input_association ON omics_processing.id = biosample_input_association.omics_processing_id
    JOIN biosample ON biosample.id = biosample_input_association.biosample_id
WHERE
    envo_term_2.label = %(label_2) s
```

So this could be caused by the introduction of the `biosample_input_association` created in PR #991.

The addition of this extra table between `omics_processing` and `envo_term` might be the source of the issue.

Further investigation is needed to determine exactly where this part of the query comes from (easy) and to determine a solution (unknown).

naglepuff:
Fix #1115 

This PR changes how subclasses of `BiosampleFilter` join to the `omics_procesing` table.

Previously, they didn't specify at all, so when a filter on an envo term was applied to a search on `omics_processing` records, there was a Cartesian product and no filtering was done.

Now, those subclasses that represent `<EnvoTerm>Filter`s use the `join_envo` function of the `BaseFilter` class to perform the join and correctly filter omics processing results.

### DIFF
--- a/nmdc_server/filters.py
+++ b/nmdc_server/filters.py
@@ -218,15 +218,21 @@ class BiosampleFilter(BaseFilter):
         return self.join_omics_processing(query)
 
 
-class EnvBroadScaleFilter(BiosampleFilter):
+class EnvoTermFilter(BiosampleFilter):
+
+    def join_omics_processing(self, query: Query) -> Query:
+        join_with_biosample_query = super().join_omics_processing(query)
+        return self.join_envo(self.table, join_with_biosample_query)
+
+class EnvBroadScaleFilter(EnvoTermFilter):
     table = Table.env_broad_scale
 
 
-class EnvLocalScaleFilter(BiosampleFilter):
+class EnvLocalScaleFilter(EnvoTermFilter):
     table = Table.env_local_scale
 
 
-class EnvMediumFilter(BiosampleFilter):
+class EnvMediumFilter(EnvoTermFilter):
     table = Table.env_medium
 
 
--- a/nmdc_server/filters.py
+++ b/nmdc_server/filters.py
@@ -219,10 +219,10 @@ class BiosampleFilter(BaseFilter):
 
 
 class EnvoTermFilter(BiosampleFilter):
-
     def join_omics_processing(self, query: Query) -> Query:
         join_with_biosample_query = super().join_omics_processing(query)
         return self.join_envo(self.table, join_with_biosample_query)
+
 
 class EnvBroadScaleFilter(EnvoTermFilter):
     table = Table.env_broad_scale
--- a/nmdc_server/filters.py
+++ b/nmdc_server/filters.py
@@ -198,7 +198,10 @@ class BiosampleFilter(BaseFilter):
     table = Table.biosample
 
     def join_omics_processing(self, query: Query) -> Query:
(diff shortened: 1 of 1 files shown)


nmdc_server/filters.py
```diff
@@ -218,15 +218,21 @@ def join_data_object(self, query: Query) -> Query:
         return self.join_omics_processing(query)
 
 
-class EnvBroadScaleFilter(BiosampleFilter):
+class EnvoTermFilter(BiosampleFilter):
+    def join_omics_processing(self, query: Query) -> Query:
```
jeffbaumes, line 222: Currently `join_self` has a condition in `BiosampleFilter` based on the table but `join_omics_processing` is differentiated by a subclass. For consistency could you either put a `if self.table in envo_tables` into the parent `join_omics_processing` and remove `EnvoTermFilter` or specialize the implementation of `join_self` here and remove the `if self.table in envo_tables` from `BiosampleFilter`?

jeffbaumes (changes requested):
This fix makes a lot of sense to me and your analysis on the parent issue confirms it's working as expected. One thing unclear to me is whether this worked before #991 or if the omics bar chart never worked properly when querying by envo terms, but it is probably not necessary for find out.

I have one consistency-based suggested change.

nmdc_server/filters.py
naglepuff: Good catch. I think that's the cleanest way to do it, and it fact, `join_omics_processing` was leveraging `join_self` prior to 688ae80.

naglepuff: db5f617 removes the intermediate class that was added by a previous commit in favor of consistency with how the other filter class functions work, and consistency with the code prior to commit 688ae80

jeffbaumes (approved):
Nice, now a one-line change (ignoring formatting), `join_omics_processing` logic looks exactly parallel to the logic in `join_biosample` and `join_study`. Perfect IMO!

PR #1119, opened 2024-01-25
Upgrade to submission-schema 10.0.0

pkalita-lbl:


eecavanna:
The changes look straightforward to me (in isolation, not knowing their implications):
- Updating a dependency to a new version, which impacts its integrity "check value"
- Importing an object from a package instead of a first-party file
- Deleting the first-party file from which that object used to be imported

naglepuff (approved):
LGTM! Love to see a good +4, -20k

PR #1127, opened 2024-02-01
Fix how Rancher workloads are redeployed in GitHub workflow

### LINKED ISSUE #1121: Fix logic to generate workload names in `deploy.yml`
We ran into a small issue while deploying Release 2024.1 to production. The [step](https://github.com/microbiomedata/nmdc-server/blob/21773ed821d3700a4258b7eb1895ba1d1c7ff429/.github/workflows/deploy.yml#L74) in `deploy.yml` where it attempts to redeploy the Rancher workloads [failed](https://github.com/microbiomedata/nmdc-server/actions/runs/7702673031/job/20991537308#step:2:16). It failed because it was attempting to restart workloads with the `portal-` prefix, which only exists in the dev namespace (i.e. `portal-backend` on dev, just `backend` on prod). The workflow file has this logic to attempt to come up with the right workload name:
```
${{ env.IS_PROD_RELEASE && '' || 'portal-' }}
```
However this doesn't work as intended because if `env.IS_PROD_RELEASE` is true then `env.IS_PROD_RELEASE && ''` evaluates to an empty string and the therefore `'' || 'portal-'` still evaluates to `portal-`. 

Since the empty string is causing the problem I think we need to flip the logic around to:
```
${{ !env.IS_PROD_RELEASE && 'portal-' }}
```
This issue will also become irrelevant if https://github.com/microbiomedata/infra-admin/issues/29 is fixed first.

### LINKED ISSUE #1122: Update `deploy.yml` to redeploy sandbox workloads for production releases
After deploying the 2024.1 release to production, I manually updated the workloads in the `nmdc-sandbox` namespace to point to the new `prod`-tagged images (previously they were using the `0.16.0`-tagged images). In order to automate the updating of sandbox we just need `deploy.yml` to also issue redeploy requests for the `nmdc-sandbox` workloads when doing a production release.

pkalita-lbl:
Previously there was logic to try to resolve the differences between workload names in the dev and prod Rancher namespaces (e.g. `portal-backend` on dev and `backend` on prod). This logic was faulty and was a little opaque. These changes remove that logic and split the dev redeploy and prod redeploy into separate steps with appropriate `if` clauses. It's a little more verbose now, but at least it's more clear what it's doing. It also makes it simpler to add an additional step to redeploy the sandbox workloads as part of a production release.

Fixes #1121
Fixes #1122

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -14,9 +14,6 @@ env:
   # Used to determine if we add the special `prod` Docker image tag
   IS_PROD_RELEASE: ${{ startsWith(github.ref, 'refs/tags/v') }}
 
-  # Used when sending redeploy action requests to Rancher
-  RANCHER_NAMESPACE: ${{ startsWith(github.ref, 'refs/tags/v') && 'nmdc' || 'nmdc-dev' }}
-
 jobs:
   build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


naglepuff (approved):
LGTM

PR #1130, opened 2024-02-06
Allow different post-ORCID login behaviors

pkalita-lbl:
These changes add a `behavior` query parameter to the existing `/login` and `/token` endpoints in order to change the behavior of what happens to the token after authenticating with ORCID. The default, `web`, retains the current behavior: store the token in a session cookie and redirect to the home page. The new ones are:

* `jwt` which just returns a plain string response of the token
* `app` which redirects to the app's token collection page (the hostname is configurable with a new variable here) with the token in a query parameter

At Shreyas's suggestion I did look to see if we could avoid re-encoding the token after the `authorize_access_token` function automatically decodes it, but nothing really jumped out to me. As far as I can tell the re-encoding isn't hurting anything so I'm inclined to just let it be.


nmdc_server/auth.py
```diff
@@ -51,11 +60,14 @@ class Token(BaseModel):
 
 
 async def get_token(
-    request: Request, access_token: Dict[str, Any] = Security(oauth2_scheme)
```
naglepuff, line 54: I'm hesitant to remove this line. Do you know what `Security(oauth2_schema)` does here?

pkalita-lbl: Thanks for point that out. Some of this is still a little opaque to me, but after some additional testing it seems like the additional scheme that I added in my original changes wasn't necessary. So now the only change to this line is correcting the type annotation.

nmdc_server/auth.py
```diff
@@ -95,27 +107,49 @@ async def admin_required(user: models.User = Depends(login_required)) -> models.
...
 
+def create_local_token(data: dict):
+    encoded_jwt = jwt.encode(data, settings.secret_key, algorithm="HS256")
+    return encoded_jwt
+
+
 router = APIRouter()
 
 
+class LoginBehavior(str, Enum):
```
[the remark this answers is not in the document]
pkalita-lbl: It does do that for sure. I often just do it out of habit. It _may_ also be required for using it as a type annotation on a query parameter with FastAPI. Well, anyway, their docs recommend it: https://fastapi.tiangolo.com/tutorial/path-params/#working-with-python-enumerations (that's about path parameters, but it works the same for query parameters)

PR #1133, opened 2024-02-08
Show depth as range or with units, if possible

eecavanna:
### Summary of changes

On the biosample details page, the depth is now displayed according to the following rules:
1. If the depth annotation contains a minimum, a maximum, and a unit → display `{min} - {max} {unit}`
1. If the depth annotation contains a minimum and a maximum (but no unit) → display `{min} - {max}`
1. If the depth annotation contains a magnitude and a unit → display `{depth} {unit}`
1. Otherwise → display whatever is in the top-level `depth` property

#### Scenario 1:

image

> Local URL: http://localhost:8081/details/sample/gold:Gb0110680

#### Scenario 2:

image

> Local URL: http://localhost:8081/details/sample/nmdc:bsm-11-q5nxet10

#### Scenario 3:

image

> Local URL: http://localhost:8081/details/sample/gold:Gb0115859

#### Scenario 4:

image

> Local URL: http://localhost:8081/details/sample/igsn:IEWFS000I

<details>
  <summary>Developer notes for myself (click to show/hide)</summary>

Here's an example of a SQL query I used to find biosamples meeting certain criteria:

```sql
SELECT id
FROM biosample
WHERE
    depth > 0
    AND annotations->'depth'->>'has_unit' IS NULL
    AND annotations->'depth'->>'has_maximum_numeric_value' IS NULL
    AND annotations->'depth'->>'has_minimum_numeric_value' IS NULL;
```

</details>

### DIFF
--- a/web/src/components/Presentation/AttributeItem.vue
+++ b/web/src/components/Presentation/AttributeItem.vue
@@ -43,7 +43,6 @@ const buildStrFromDepthAnnotation = (depthAnnotation: JSONValue, format: Format)
         break;
       }
       default: {
-        console.error('Invalid format specifier.');
         break;
       }
     }
--- a/web/src/components/Presentation/AttributeItem.vue
+++ b/web/src/components/Presentation/AttributeItem.vue
@@ -110,10 +110,10 @@ export default defineComponent({
             if (typeof rangeStr === 'string') {
               result = rangeStr;
             } else {
-              // Check whether the raw depth is non-null and the depth annotation contains a unit;
+              // Check whether the raw depth is numeric (i.e. not null) and the depth annotation contains a unit;
               // and, if so, use a concatenation of the two as the result.
               const unitStr = buildStrFromDepthAnnotation(props.item.annotations.depth, Mode.Unit);
-              if (rawDepth !== null && typeof unitStr === 'string') {
+              if (typeof rawDepth === 'number' && typeof unitStr === 'string') {
                 result = `${rawDepth} ${unitStr}`;
               }
             }
--- a/web/src/components/Presentation/AttributeItem.vue
+++ b/web/src/components/Presentation/AttributeItem.vue
@@ -17,7 +17,7 @@ export enum Format {
 const buildStrFromDepthAnnotation = (depthAnnotation: JSONValue, format: Format): string | undefined => {
   let str: string | undefined;
 
-  // Regardless of the specified mode, check whether the depth annotation is a "non-null, non-array" object.
+  // Regardless of the specified format, check whether the depth annotation is a "non-null, non-array" object.
   if (typeof depthAnnotation === 'object' && depthAnnotation !== null && !Array.isArray(depthAnnotation)) {
     // Extract data according to the specified format.
     switch (format) {
--- a/web/src/components/Presentation/AttributeItem.vue
+++ b/web/src/components/Presentation/AttributeItem.vue
@@ -104,21 +104,22 @@ export default defineComponent({
 
         // Check whether there is a depth annotation.
         if ('depth' in props.item.annotations) {
+          const depthAnnotation = props.item.annotations.depth as JSONValue;
           // Check whether the depth annotation describes a range with a unit;
           // and, if so, use that as the result.
-          const rangeWithUnitStr = buildStrFromDepthAnnotation(props.item.annotations.depth, Format.RangeWithUnit);
+          const rangeWithUnitStr = buildStrFromDepthAnnotation(depthAnnotation, Format.RangeWithUnit);
           if (typeof rangeWithUnitStr === 'string') {
             result = rangeWithUnitStr;
           } else {
             // Check whether the depth annotation describes a range;
             // and, if so, use that as the result.
-            const rangeStr = buildStrFromDepthAnnotation(props.item.annotations.depth, Format.Range);
+            const rangeStr = buildStrFromDepthAnnotation(depthAnnotation, Format.Range);
             if (typeof rangeStr === 'string') {
               result = rangeStr;
             } else {
               // Check whether the top-level depth value is numeric (not null) and the depth annotation contains a unit;
               // and, if so, use a concatenation of the two as the result.
-              const unitStr = buildStrFromDepthAnnotation(props.item.annotations.depth, Format.Unit);
+              const unitStr = buildStrFromDepthAnnotation(depthAnnotation, Format.Unit);
               if (typeof topLevelDepth === 'number' && typeof unitStr === 'string') {
                 result = `${topLevelDepth} ${unitStr}`;
               }
--- a/web/src/components/Presentation/AttributeItem.vue
+++ b/web/src/components/Presentation/AttributeItem.vue
@@ -4,41 +4,38 @@ import { getField } from '@/encoding';
 import { fieldDisplayName, valueDisplayName } from '@/util';
 import { BaseSearchResult, BiosampleSearchResult, JSONValue } from '@/data/api';
 
-enum Mode {
+export enum Format {
   RangeWithUnit = 'RangeWithUnit',
   Range = 'Range',
   Unit = 'Unit',
 }
 
 /**
- * Helper function that checks whether the specific depth annotation adequately describes something and, if so,
- * returns a string representing that thing; otherwise, it returns `undefined`.
- *
- * @param depthAnnotation {JSONValue} The `depth` annotation containing data from which you want to build a string
- * @param mode {Mode} Keyword indicating the type of string you want to build
+ * Helper function that returns a string representation of the depth in the specified format, if the specified
+ * depth annotation contains adequate information to do so. Otherwise, the function returns `undefined`.
  */
-const buildStrFromDepthAnnotation = (depthAnnotation: JSONValue, mode: Mode): string | undefined => {
+const buildStrFromDepthAnnotation = (depthAnnotation: JSONValue, format: Format): string | undefined => {
   let str: string | undefined;
 
   // Regardless of the specified mode, check whether the depth annotation is a "non-null, non-array" object.
   if (typeof depthAnnotation === 'object' && depthAnnotation !== null && !Array.isArray(depthAnnotation)) {
-    // Extract data according to the specified mode.
-    switch (mode) {
-      case Mode.RangeWithUnit: {
+    // Extract data according to the specified format.
+    switch (format) {
+      case Format.RangeWithUnit: {
         const { has_minimum_numeric_value, has_maximum_numeric_value, has_unit } = depthAnnotation;
         if (typeof has_minimum_numeric_value === 'number' && typeof has_maximum_numeric_value === 'number' && typeof has_unit === 'string') {
           str = `${has_minimum_numeric_value} - ${has_maximum_numeric_value} ${has_unit}`;
         }
         break;
       }
-      case Mode.Range: {
+      case Format.Range: {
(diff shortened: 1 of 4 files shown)


eecavanna:
### Messages to reviewers

- To @aclum: Will you review the written rules and the corresponding screenshots in the PR description, and tell me whether those reflect what you want? I don't expect you to review the source code, although you can if you want.
- To @naglepuff and/or @marySalvi: This is my first PR on the client side and my first time working with Vue.js (I'm familiar with React, though). I seem to have unearthed some TypeScript errors. I suspect they are due to the fact that I updated the type "annotation" for the `BaseSearchResult.annotation` property (which is annotated in `api.ts`), so it would accommodate an arbitrary JSON object (instead of just a string or array of strings). I think the new type more closely reflects reality; but maybe there's a more `nmdc-server`-friendly way of specifying it. I will look into the errors now. I am curious whether you have any guidance for me about them, specifically; in addition to the changes, overall.

eecavanna:
I think I found another preexisting type annotation may not account for all possible values. The `depth` property has type `number`, whereas I have seen the client receive a biosample object from the server where `depth` contained `null`.

https://github.com/microbiomedata/nmdc-server/blob/c0b7b03089df19c47657e041eb770229e9f8f1c8/web/src/data/api.ts#L86

I'm wondering whether I'd be going down a "rabbit hole" by trying to reconcile the type annotations; whereas I could, alternatively, maybe sprinkle in some `any`s and kick the can down the road. 🤷

eecavanna:
I opted to revert the changes I made to the `BaseSearchResult.annotations` type annotation, and cast things to `JSONValue` downstream so as not to impact the rest of the code base. I consider this branch done and I am ready for people to review it (and merge it, if they are comfortable doing so).

I'll create a separate issue about the type annotations I mentioned above, which I think account for only a subset of the values the SQL schema allows.

aclum (approved):
Behavior looks correct to me based on screenshots and descriptive text of desired behavior.

web/src/components/Presentation/AttributeItem.vue
```diff
@@ -2,7 +2,63 @@
...
+  Range = 'Range',
+  Unit = 'Unit',
+}
+
+/**
+ * Helper function that returns a string representation of the depth in the specified format, if the specified
+ * depth annotation contains adequate information to do so. Otherwise, the function returns `undefined`.
+ */
+const buildStrFromDepthAnnotation = (depthAnnotation: JSONValue, format: Format): string | undefined => {
+  let str: string | undefined;
```
naglepuff, line 18: This function is a little mysterious to me. My understanding is that it will generate a depth string for a given depth annotation according to the specified format. If the specified format is unsupported by the amount of data in the annotation, `undefined` is returned. But when `format === Format.Unit`, it only returns the unit, with no scalar value and the calling function is then responsible for picking the right scalar value for the unit.

Do we have a use case for getting the depth string using a format that doesn't include all possible information from the `Biosample` object? If not, then what if this function always returned the most detailed possible depth string?

If this function took the additional parameter of `topLevelDepth`, could it be used to generate the full string for every case?

I think I would have expected the function using this helper to do `depthString = buildStrFromDepthAnnotation(...)` just and have the string ready to go.

It might need to be renamed (since the current name only implies that it uses the annotation), but this change could clean up the changes in `AttributeList.vue`.

web/src/components/Presentation/AttributeItem.vue
```diff
@@ -2,7 +2,63 @@
...
+
+  // Regardless of the specified format, check whether the depth annotation is a "non-null, non-array" object.
+  if (typeof depthAnnotation === 'object' && depthAnnotation !== null && !Array.isArray(depthAnnotation)) {
+    // Extract data according to the specified format.
+    switch (format) {
+      case Format.RangeWithUnit: {
+        const { has_minimum_numeric_value, has_maximum_numeric_value, has_unit } = depthAnnotation;
+        if (typeof has_minimum_numeric_value === 'number' && typeof has_maximum_numeric_value === 'number' && typeof has_unit === 'string') {
```
naglepuff, line 26:
```suggestion
        if (has_minimum_numeric_value && has_maximum_numeric_value && has_unit) {
```

Could existence checks be sufficient here?

naglepuff (changes requested):
Overall looks solid. I had a couple of suggestions for some of the helper functions.

eecavanna:
Thanks, @naglepuff! The feedback you left makes sense to me. 

I'm OOO today and will revise the branch with one or more additional commits once I'm back in.

Regarding checking the type of something versus whether it is "truthy," I tend to do that in contexts where I don't know the possible range of types that the value could have. I'll elaborate when I address each feedback comment, once I'm back in the office.

web/src/components/Presentation/AttributeItem.vue
```diff
@@ -2,7 +2,63 @@
...
+};
+
+/**
+ * Helper function that checks whether the specified depth annotation describes a range.
+ */
+export const doesDepthAnnotationDescribeRange = (depthAnnotation: JSONValue): boolean => {
+  const rangeWithUnit = buildStrFromDepthAnnotation(depthAnnotation, Format.RangeWithUnit);
+  const range = buildStrFromDepthAnnotation(depthAnnotation, Format.Range);
+  return typeof rangeWithUnit === 'string' || typeof range === 'string';
+};
```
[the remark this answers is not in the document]
eecavanna: I don't have any objections to moving these to `util.js`. I will move these functions to that file.

web/src/components/Presentation/AttributeItem.vue
eecavanna: Thanks for bringing this up. I don't know.

The reason I did a type check is that I only wanted to use the value if it had the type I specified; and I didn't know what possible types the value might have. I tried to find that out by looking at the SQLAlchemy migrations and TypeScript type definitions, but failed to find type information for the `depth` annotation.

- The [SQL migrations](https://github.com/microbiomedata/nmdc-server/blob/e301cf26c00be161f0ee31ff1087f86c743161a5/nmdc_server/migrations/versions/780ab86a4f27_squashed_initial.py#L193) describe the `annotations` object as arbitrary JSON.
- The [TypeScript type definitions](https://github.com/microbiomedata/nmdc-server/blob/e301cf26c00be161f0ee31ff1087f86c743161a5/web/src/data/api.ts#L47) describe the `annotations` object as one whose properties contain either strings or arrays of strings (which I thought was incorrect, as the `depth` annotation is neither a string nor an array of strings).

For that reason, I thought checking for truthiness might be insufficient, since the property might contain something like an array of strings.

Also, I didn't want to give special treatment to the number `0` (I assumed `0` was a valid depth), and JavaScript treats `0` as falsy.

web/src/components/Presentation/AttributeItem.vue
eecavanna: Thanks for giving me detailed feedback about this function. The summary you wrote in the first paragraph is accurate.

> Do we have a use case for getting the depth string using a format that doesn't include all possible information from the `Biosample` object?

I don't know of such a use case.

> If not, then what if this function always returned the most detailed possible depth string?
>
> If this function... (the rest of your message)

I think that would be a more intuitive solution than the one I implemented. I'll refactor the code to do that.

web/src/components/Presentation/AttributeItem.vue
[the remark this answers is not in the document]
eecavanna: Resolved. I moved this function in a subsequent commit.

web/src/components/Presentation/AttributeItem.vue
eecavanna: Resolved. In a subsequent commit, I documented my rationale for checking types.

In case that's insufficient, please add a new comment (I don't want to add any code you aren't comfortable with having in the repo).

eecavanna:
Hi @naglepuff, I'm ready for this to be reviewed again.

web/src/data/api.ts
```diff
@@ -39,6 +39,15 @@ export type entityType = 'biosample'
...
 
+/**
+ * Define a TypeScript type alias that accounts for all possible JSON values.
+ *
+ * References:
+ * - https://www.json.org/json-en.html
+ * - https://devblogs.microsoft.com/typescript/announcing-typescript-3-7/#more-recursive-type-aliases
+ */
+export type JSONValue = string | number | boolean | null | { [k: string]: JSONValue } | Array<JSONValue>;
+
```
naglepuff, line 50: Is this being used anywhere? If not should it be removed (at least for now?)

web/src/util.js
```diff
@@ -141,3 +141,50 @@ export function saveAs(filename, text) {
...
+ * - https://microbiomedata.github.io/nmdc-schema/QuantityValue/
+ *
+ * @param depthAnnotation {object | null} Value of the biosample's `annotations.depth` property
+ * @param depthAnnotation.has_minimum_numeric_value {number?} Lower magnitude of the quantity's range
+ * @param depthAnnotation.has_maximum_numeric_value {number?} Upper magnitude of the quantity's range
+ * @param depthAnnotation.has_numeric_value {number?} Magnitude of the quantity
+ * @param depthAnnotation.has_unit {string?} Unit of the quantity
+ * @param depth {number | null} Value of the biosample's top-level `depth` property
+ * @return {string | null} Either a string describing the depth of the biosample, or `null`
+ */
```
naglepuff, line 171: I appreciate the detailed docstring here!

naglepuff (changes requested):
Thanks for addressing my previous issues. I only have one more suggestion, but overall I think this is a cleaner approach.

web/src/util.js
```diff
@@ -141,3 +141,50 @@ export function saveAs(filename, text) {
...
+export const formatBiosampleDepth = (depthAnnotation, depth) => {
```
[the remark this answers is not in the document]
eecavanna: Yes, fixed in 385fb74cbaf808bb298206bc4d409d3cd6b22236

web/src/data/api.ts
eecavanna: It is not being used anywhere. I will remove it. I think it will be useful as a more accurate type for the annotations property, but I didn't implement it there yet because that would open a "can of worms" downstream (where some code is written under the assumption that the current type is comprehensive, which I don't think is consistent with reality). I'll remove it from this PR branch and file an issue about my type concerns.

eecavanna: Fixed (removed) in eab88f26484907bf0c7ee94e8903381404a99c79

web/src/util.js
eecavanna: I'm glad! Thanks for the feedback. (This is the first time I've used the https://jsdoc.app/tags-param#parameters-with-properties feature.)

eecavanna:
Hi @naglepuff, thanks again for leaving those direct and actionable review comments. I have made all the changes I think you suggested and am ready for a follow-up review/merge.

naglepuff (approved):
LGTM!

eecavanna:
Thanks, @naglepuff! I'm comfortable with this being merged in. This is my first time working on a user-facing aspect of this project; and I don't know whether there's any special protocol around merging vs. waiting to merge user-facing changes. I'll wait until the workweek starts, to ask on Slack. You're welcome to merge this in in the meantime.

eecavanna:
I'll go ahead and merge this so people can start kicking its tires in the dev environment, ahead of next Monday's [release](https://github.com/microbiomedata/issues/issues/597).

PR #1136, opened 2024-02-12
Make certain JGI columns always read-only in DataHarmonizer

pkalita-lbl:
Re: https://github.com/microbiomedata/issues/issues/412 (will close once it has been tested on dev)

These changes make a set of JGI-related columns read-only in DataHarmonizer. This does _not_ prevent the columns from being populated by importing an XLSX file. This is a feature, not a bug.

To test:
* Go to any submission with a JGI MG and/or JGI MT tab
* Enter data in the first tab to ensure that some rows will display in the JGI tab(s)
* On the JGI tab(s) ensure that the indicated columns cannot be edited


naglepuff (approved):
LGTM!

mslarae13:
@pkalita-lbl Looks great!!
I think when we have the API, we will have to look into not including these columns in the sheet download. Cuz if I edit it in excel, and then import it does change the values. But I think that's a different issue. I think that is part of the API development. 

I like that the cursor turns into a 🚫 when hovering over those columns.

PR #1139, opened 2024-02-14
Allow adding CORS middleware based on env settings

### LINKED ISSUE #1135: Allow CORS requests from Field Notes app
Currently `nmdc-server` does not allow CORS requests. In order to enable the Field Notes app to make `nmdc-server` API requests, it will need to accept request from certain origins. The docs for the framework we're using to build the app suggests a set of origins here: https://ionicframework.com/docs/troubleshooting/cors#a-enabling-cors-in-a-server-you-control. We should additionally allow the web version of the app deployed on https://fieldnotes.microbiomedata.org.

One open question is where to add the configuration. I know that portal traffic is handled by an nginx server which proxies `/api` request to the backend ([ref](https://github.com/microbiomedata/nmdc-server/blob/ec84c0126cfcbdb16c86b330d72507322af44f83/web/nginx.conf.template)). I think it would be possibly more straightforward to add the CORS setup directly to the backend's FastAPI application. But maybe there are advantages to adding it at the nginx layer? I'm not sure.

pkalita-lbl:
Fixes #1135

### DIFF
--- a/nmdc_server/config.py
+++ b/nmdc_server/config.py
@@ -62,6 +62,9 @@ class Settings(BaseSettings):
     rancher_backend_workload_id: Optional[str] = None
     rancher_worker_workload_id: Optional[str] = None
 
+    # CORS settings necessary for allowing request from Field Notes app
+    cors_allow_origins: Optional[str] = None
+
     @property
     def current_db_uri(self) -> str:
         if self.environment == "testing":
--- a/nmdc_server/app.py
+++ b/nmdc_server/app.py
@@ -2,6 +2,7 @@ import typing
 
 import sentry_sdk
 from fastapi import FastAPI
+from fastapi.middleware.cors import CORSMiddleware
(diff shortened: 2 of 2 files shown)


nmdc_server/config.py
```diff
@@ -62,6 +62,9 @@ class Settings(BaseSettings):
     rancher_backend_workload_id: Optional[str] = None
     rancher_worker_workload_id: Optional[str] = None
 
+    # CORS settings necessary for allowing request from Field Notes app
+    cors_allow_origins: Optional[str] = None  # comma separated list of allowed origins
```
eecavanna, line 66: Assuming the comma-delimited string can be defined in an `.env` file, I'd recommend adding an example to [.env.example](https://github.com/microbiomedata/nmdc-server/blob/main/.env.example) because I think people might not realize the string can be defined in that file (as opposed to editing this Python script).

naglepuff (approved):

PR #1149, opened 2024-02-22
Add debounce to fetch results when conditions change

### LINKED ISSUE #1034: Issue when copying a data portal URL query from one google chrome tab to another

marySalvi:
Closes #1034 
Adding a debouce to `fetchResults` when conditions change ensures that the desired query when copying a url is fired second and eliminates the race condition.

### DIFF
--- a/web/src/use/usePaginatedResults.ts
+++ b/web/src/use/usePaginatedResults.ts
@@ -41,22 +41,24 @@ export default function usePaginatedResult<T>(
     });
   }
 
+  const debouncedFetchResults = debounce(fetchResults, 500);
+
   watch([
     toRef(data, 'limit'),
     toRef(data, 'offset'),
-  ], fetchResults);
+  ], debouncedFetchResults);
 
-  watch([conditions], debounce(() => {
+  watch([conditions], () => {
     const doFetch = data.offset === 0;
     data.offset = 0;
     data.limit = limit;
-    if (doFetch) fetchResults();
-  }, 500));
+    if (doFetch) debouncedFetchResults();
+  });
 
   if (dataObjectFilter !== undefined) {
-    watch(dataObjectFilter, fetchResults, { deep: true });
+    watch(dataObjectFilter, debouncedFetchResults, { deep: true });
   }
-  fetchResults();
+  debouncedFetchResults();
   // ENDTODO
 
   function setPage(newPage: number) {
--- a/web/src/use/usePaginatedResults.ts
+++ b/web/src/use/usePaginatedResults.ts
@@ -1,6 +1,7 @@
 import {
   watch, Ref, computed, toRef, shallowReactive,
 } from '@vue/composition-api';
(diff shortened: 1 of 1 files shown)


naglepuff (changes requested):
I'm still seeing 2 requests on reload. (http://localhost:8081/?q=ChsIABACGAIiEyJNZXRhdHJhbnNjcmlwdG9tZSIKHwgAEAAYASIXIktFR0cuT1JUSE9MT0dZOkswMDI5MiI=)

This is likely because the `fetchResults` call on line 59 (outside of the watcher) is not debounced.

I think creating a debounced version of `fetchResults` and using it in both places would do the trick.

marySalvi:
> I'm still seeing 2 requests on reload. (http://localhost:8081/?q=ChsIABACGAIiEyJNZXRhdHJhbnNjcmlwdG9tZSIKHwgAEAAYASIXIktFR0cuT1JUSE9MT0dZOkswMDI5MiI=)
> 
> This is likely because the `fetchResults` call on line 59 (outside of the watcher) is not debounced.
> 
> I think creating a debounced version of `fetchResults` and using it in both places would do the trick.

Good call. Fixed in e1d41fa1675df8bfb6f45de82f07038c671c13bb

naglepuff (approved):
LGTM!

I used the network tab in devtools to verify that only:

- 1 call was made to `/api/biosample/search`
- 2 calls were made to `/api/study/search` (1 for studies, 1 for consortia),

Where before 2 and 4 calls were made respectively.

I also checked http://localhost:8081/users and http://localhost:8081/submission/home to make sure that the results weren't suppressed (which was the reason the original fix for this problem had to be rolled back).

Issue #1085, opened 2023-11-29
EMP 500 study is showing up in study results with map filter without showing any samples


aclum:
I believe there is a bug in the lat/long filter. If I use the map feature to select an area around Puerto Rico EMP500 shows up in the results, however when add that study to the facet search it shows zero samples.
Initial filter
latitude is between 2.589309996397766 and 32.359222618551655
longitude is between -87.0042084501952 and -45.871395950195186

then add
study id is gold:Gs0154244

There are no samples from gold:Gs0154244 between these geo coordinates so the expected behavior is that gold:Gs0154244  (EMP500) should not show up on the study list.

![Screenshot 2023-11-29 at 11 42 26 AM](https://github.com/microbiomedata/nmdc-server/assets/8196591/0c8d6539-e211-4bc5-9c20-14209ffa0621)

naglepuff:
I'm able to confirm this behavior locally with the following search:
![image](https://github.com/microbiomedata/nmdc-server/assets/7085625/e748b4f3-6f0b-46ed-831e-7fb0252cf1b5)

I suspect that the logic that is being applied to the studies is something like:
```
Include STUDY if 
(there exists >=1 BIOSAMPLE that is part_of STUDY) 
and (>=1 BIOSAMPLES that are part_of STUDY have latitude between [y0, y1]) 
and (>=1 BIOSAMPLES that are part_of STUDY have longitude between [x0, x1])
```

This needs to be confirmed by checking the code and the SQL that is generated by the ORM, but the current behavior of the data portal does indicate that this is the logic being applied. For example, if I modify my example query above to remove the latitude filter, I get biosample results:
![image](https://github.com/microbiomedata/nmdc-server/assets/7085625/dcec268b-1d67-4cef-8dba-57a11c49ecb3)

Similarly, removing the longitude filter looks like:
![image](https://github.com/microbiomedata/nmdc-server/assets/7085625/3b8f82ce-3941-4dd3-bccb-b877fcbe1bac)

aclum:
I get the same results as you  Mike when only intersecting on latitude and study id OR longitude and study.

aclum:
Confirmed fixed in dev.

PR #1153, opened 2024-02-26
Fix multi-tiered search

### LINKED ISSUE #1072: KEGG search results should only display products from which the KEGG hit comes from
**Is your feature request related to a problem? Please describe.**
When searching by a KEGG search term it should only products where it makes sense to have KEGG results (metagenome, metatranscriptome, and proteomics) 
Example (https://data-dev.microbiomedata.org/?q=Ch8IABAAGAEiFyJLRUdHLk9SVEhPTE9HWTpLMDAwMDki)

**Describe the solution you'd like**
A metagenome sample has a hit to the KEGG term and also has a metatranscriptome with no KEGG hit and metabolomics the sample results should only show the metagenome product. The count of metagenomes at the study level should be the number of metagenomes in that study that have the KEGG term, not the total number of metagenome projects in the entire study. 
Minimum requirement for GSP would be to hid metabolomics and natural organic matter results for studies and samples

**Describe alternatives you've considered**
Minimum requirement for GSP would be to hid metabolomics and natural organic matter results for studies and samples

**Acceptance Criteria**
Search by KEGG term  shows the number of projects within a study that have hits and only those with products. At the sample level only show products with the KEGG hit.
Screenshot 2023-03-30 at 7 47 23 PM

naglepuff:
Fix #1072 

The NMDC Data Portal allows adding filters on multiple properties across different levels of the NMDC schema hierarchy. For example, one can add a search filter at the biosample level (depth) and see results from the study and data object level.

Prior to this change, searching across multiple hierarchies could introduce some unexpected results due to how our queries were being constructed. Examples of this inconsistency are documented fairly well in #1072 and #1085.

The fix for #1085 was the first step toward fixing this larger problem, which this PR aims to expand to be a fully fledged fix.

### DIFF
--- a/nmdc_server/query.py
+++ b/nmdc_server/query.py
@@ -656,13 +656,27 @@ class StudyQuerySchema(BaseQuerySchema):
 
     def query(self, db: Session):
         study_query = super().query(db)
-        if any([condition.table == Table.biosample for condition in self.conditions]):
+        biosample_condition_exists = any(
+            [condition.table == Table.biosample for condition in self.conditions]
+        )
+        omics_condition_exists = any(
+            [condition.table == Table.omics_processing for condition in self.conditions]
+        )
+        if biosample_condition_exists:
             sample_query = BiosampleQuerySchema(conditions=self.conditions).query(db)
             studies_from_sample_query = sample_query.with_entities(
                 models.Biosample.study_id
             ).distinct()
             study_query = study_query.where(  # type: ignore
                 self.table.model.id.in_(studies_from_sample_query)  # type: ignore
+            )
+        elif omics_condition_exists:
+            omics_query = OmicsProcessingQuerySchema(conditions=self.conditions).query(db)
+            studies_from_omics_query = omics_query.with_entities(
+                models.OmicsProcessing.study_id
+            ).distinct()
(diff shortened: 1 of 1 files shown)


naglepuff:
Borrowing a [screenshot](https://github.com/microbiomedata/nmdc-server/issues/1072#issuecomment-1791604544) from #1072:
image

There is a study and biosamples showing up in the results that should not be. This is because both of the active search terms use `omics_processing` records as the basis for search. Since building the SQL queries works from the top down, the queries for `biosample` and `study` effectively "or" these conditions together since they are lower on the `study -> sample -> omics` hierarchy.

The change in this PR will "pre-filter" records that are lower on the hierarchy, to avoid the pseudo-or behavior previously seen. In my local environment, the search above yields:
image

naglepuff:
@aclum besides the query mentioned above from #1072, are there any other specific sets of filters that you've noticed with unexpected results?

jeffbaumes:
@naglepuff at a brief glance the logic looks sound. Study search needs to "chain down" to sample (done in the previous fix) and omics processsing, and sample search needs to "chain down" to omics processing.

jeffbaumes (commented):
I'm retroactively approving :) Looks good.

PR #1161, opened 2024-02-29
Grow Study fixes

marySalvi:
Use annotation title for child studies if available.
Children studies whose parent isn't in the search results should only be included if they don't appear as a child in any other part study present in the search results.
You can check both issues by expanding `River Corridors Science Focus Area`. The correct title for the grow child study is `Creating the GROW...`. Also because grow appears here it should not appear as a top level study in the study results.


naglepuff (approved):
LGTM! Verified that both issues are fixed by comparing local with dev.

PR #1164, opened 2024-02-29
Study page fixes

marySalvi:
- Limits Team section to three lines and overflow scrolls
- Does not show PI and ORCID on consortia
- Study_count on Parent studies is a sum of its count  and its childrens' counts
- Fixes bug where consortia children were not disabled when the parent was selected.


naglepuff (approved):
LGTM! If you can figure out the 3 dots thing feel free to fix here, otherwise I say merge it

PR #1168, opened 2024-03-01
Remove flat option for study api

marySalvi:
At this point not getting children results would be unexpected behavior


naglepuff (approved):
LGTM

PR #1171, opened 2024-03-04
Submission Permission fixes

### LINKED ISSUE #939: EMSL Shipping Validation
> If I start a submission, leave it or the "submission context" page, the EMSL shipping information appears as incomplete, red box... I have to open it, and close it (no changes made) to move forward

_Originally posted by @mslarae13 in https://github.com/microbiomedata/nmdc-server/issues/835#issuecomment-1499474207_

### LINKED ISSUE #1162: Tooltip/documentation for submission editor roles and permissions
Additional Updates 

- [x] We should add a tool tip that explains the different permission levels.
   - Editor: Editors of a submission have full permission to edit every aspect of the submission with the exception of permission levels.
   - Owner (Author?): Submission author and Principal Investigator have full permission to edit every aspect of the submission.
   - Viewer: A viewer can see all components of a submission, but cannot edit.
   - Metadata Contributor: A metadata contributor can view all components of a submission and can only edit the sample metadata information on the last step of the submission process.
- [x]  New bug: If someone is a metadata contributor, they can't validate the EMSL shipping to move on. This relates to the EMSL shipping bug here: https://github.com/microbiomedata/nmdc-server/issues/939 
- [ ]  Add a "No shipping information required" check box.
   - [ ] https://github.com/microbiomedata/nmdc-server/issues/1083 
- [x]  Add a banner for "Metadata Contributor" that says "Permission level access does not allow editing of this page. Contact the submission Author to request a change in permission status"
- [x]  If I assign someone as PI, AND make them viewer in the Contributors section, the Contributor permission supersedes PI. Make this the other way around.

naglepuff:
Fix #1162 
Fix #939 ~~(temporary/non-ideal fix)~~

This PR addresses additional issues with assigning permissions from the UI and adds some UX enhancements like banners and tooltips.

#### Don't Override PI Owner Role
If the same ORCID iD was used for the PI ORCID and as a contributor ORCID iD, then the permission level set at the contributor level had priority over the ideal PI level (`owner`). Now, priority is given to the `owner` role when building the request body to update the submission with new permissions.

#### Validate address without rendering form
~~This is a "fix" for #939. The address form now validates by investigating data in addition to through Vuetify's form validation. Vuetify's form validation was insufficient because the address form cannot be validated by it until it is opened. The previous workaround for this was to open the form, which would be validated upon render by Vuetify, and then close the form. Metadata Contributors are unable to do this, and would be stuck at that screen. The only workaround for those users would be to replace `/context` with `/samples` in the URL.~~

~~The reason this fix is not ideal is because it duplicates validation, so changes to validation rules for this form need to be applied in two different places.~~

The address form dialog now uses the [eager prop](https://vuetifyjs.com/en/api/v-dialog/#props-eager), which forces it to render and thus validate when it mounts. (thanks @marySalvi)

#### Add Permission Level Descriptions
A tooltip has been added to the Study form describing each permission level. 
image

#### Add Banner for read-only views
If a user is presented with a read-only view due to insufficient permissions to edit the current form, a banner now appears informing the user of this.
image

#### Prevent PI from changing their own ORCID
If the current user is the PI, don't allow them to edit the PI ORCID field on the study form and remove their permission.

### DIFF
--- a/web/src/views/SubmissionPortal/Components/SubmissionPermissionBanner.vue
+++ b/web/src/views/SubmissionPortal/Components/SubmissionPermissionBanner.vue
@@ -1,13 +1,3 @@
-<script lang="ts">
-import { defineComponent } from '@vue/composition-api';
-
-export default defineComponent({
-  setup() {
-    return {};
-  },
-});
-</script>
-
 <template>
   <v-alert
     type="info"
--- a/web/src/views/SubmissionPortal/Components/SubmissionPermissionBanner.vue
+++ b/web/src/views/SubmissionPortal/Components/SubmissionPermissionBanner.vue
@@ -1,0 +1,17 @@
+<script lang="ts">
+import { defineComponent } from '@vue/composition-api';
+
+export default defineComponent({
+  setup() {
+    return {};
+  },
+});
+</script>
+
+<template>
+  <v-alert
+    type="info"
+  >
+    Your current permission level for this submission does not allow editing of this page. Contact the submission author to request a change in permission level.
+  </v-alert>
+</template>
--- a/web/src/views/SubmissionPortal/Components/MultiOmicsDataForm.vue
+++ b/web/src/views/SubmissionPortal/Components/MultiOmicsDataForm.vue
@@ -5,9 +5,10 @@ import {
   multiOmicsForm, multiOmicsFormValid, multiOmicsAssociations, templateChoiceDisabled, contextForm, canEditSubmissionMetadata,
 } from '../store';
 import SubmissionDocsLink from './SubmissionDocsLink.vue';
+import SubmissionPermissionBanner from './SubmissionPermissionBanner.vue';
 
 export default defineComponent({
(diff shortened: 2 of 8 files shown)


web/src/views/SubmissionPortal/Components/SubmissionPermissionBanner.vue
```diff
@@ -0,0 +1,17 @@
+<script lang="ts">
+import { defineComponent } from '@vue/composition-api';
+
+export default defineComponent({
+  setup() {
+    return {};
+  },
+});
```
marySalvi, line 8: Remove if not using.

marySalvi (approved):
One minor change but otherwise looks good!

web/src/views/SubmissionPortal/Components/SubmissionPermissionBanner.vue
naglepuff: Removed in ef19f02

PR #1172, opened 2024-03-06
Serve sumission schema files from backend

### LINKED ISSUE #1166: Add endpoint for retriving sample metadata schema
The Field Notes app needs to collect sample metadata in a way that's consistent with the main submission portal. The submission portal relies on a JSON-formatted version of the `submission-schema` which is then provided to a DataHarmonizer instance: https://github.com/microbiomedata/nmdc-server/blob/ccd899bf07baae6abac7049a20202d6908c72c60/web/src/views/SubmissionPortal/harmonizerApi.ts#L195. In order to keep the app and the backend in sync we need to _not_ bundle a copy of the schema into the app; instead it should fetch it dynamically from the backend. To support that we should add an API endpoint that returns the schema.

pkalita-lbl:
Fixes #1166 

### Summary

This is partially to support the Field Notes app, but I think it also has benefits for the main submission portal as well. The gist is to allow API clients (whether that's the submission portal or the app) to get submission schema artifacts from the API instead of bundling them in the client application. For the submission portal this hasn't been a major problem (other than the weirdness of having a PyPI artifact as a frontend dependency) since the frontend and backend are part of the same codebase and released at the same time. That won't necessarily be true for the Field Notes app so we want to be able to dynamically fetch the schema.

### Details

* A new module `nmdc_server/static_files.py` defines a path to hold static files and a function which knows how to introspect files provided by the `nmdc_submission_schema` package and write JSON files into the static file directory. In particular it produces a JSON file representing the whole submission schema (used by the submission portal), files representing individual classes of the schema (likely to be used by the Field Notes app), and the GOLD ecosystem tree (used by the submission portal and likely to be used by the Field Notes app).
* When the backend app starts up, the function to generate the static files is called and the static file directory is mounted under the `/static` path.
* The frontend app fetches the necessary JSON files from the `/static` routes instead of importing packaged files.
* Include `nmdc-submission-schema` as a backend dependency, remove it as a frontend dependency.
* Increment the `nmdc-schema` dependencies to the latest version.

### Comments

I decided on the static file approach because FastAPI will serve the right headers (I think mayble Etag) in the responses to allow browsers to aggressively cache the responses. 

In the future we could consider whether we can piggyback on this approach to also remove the `nmdc-server` frontend dependency.


nmdc_server/static_files.py
```diff
@@ -0,0 +1,43 @@
...
+
+from linkml_runtime import SchemaView
+from linkml_runtime.dumpers import json_dumper
+
+STATIC_PATH = Path("static")
+try:
+    shutil.rmtree(STATIC_PATH)
+except FileNotFoundError:
+    pass
+STATIC_PATH.mkdir(parents=True, exist_ok=True)
```
naglepuff, line 13: Should this be wrapped in a function, and called explicitly when needed? That way we have more control over when we blow up the static files directory. Not an issue now but if this module ends up being imported elsewhere, it might cause headaches.

naglepuff (changes requested):
I think this is a pretty elegant solution. I had one comment on the code, and I have additional questions about deployment - is there anything that needs to be done on the spin side to make sure there's no problem creating the static file directory?

nmdc_server/static_files.py
pkalita-lbl: Yeah good call. Let's not set ourselves up for foot-shooting in the future.

pkalita-lbl:
> is there anything that needs to be done on the spin side to make sure there's no problem creating the static file directory?

Not that I'm aware of, but we should find out as soon as this goes to dev if there is.

naglepuff (approved):
Looks good. I tested this locally by visiting some of the `/static` urls and confirming that I got some neat json files as a response, and checking that the DH works correctly (i.e. expected columns, validation). As far as I could tell, things appear to be in working order.

PR #1177, opened 2024-03-08
Create alert to show success upon validation

JamesTessmer:
The end goal of this PR is to have an alert that appears when a user clicks validate and the validation is successful, prompting them to click submit. We want to be minimally intrusive and match the existing style.

Completed work:
- Added snackbar component
   - Appears at bottom of the page for 3 seconds and doesn't block editing or other interaction
   - changed color to 'green' from 'success' to match the colors around a little better
- Updated Submit button to change to be green when validation is succesful

Current Appearance:

https://github.com/microbiomedata/nmdc-server/assets/47541313/8346bc00-49e4-4c46-abbe-c8952d3da875

Here's the new green color

Screenshot 2024-03-29 at 9 02 54 AM


JamesTessmer:
@mslarae13 @naglepuff I've updated the banner to an alert (thank you for pointing out that option Mike) that appears at the top of the page when validation succeeds. 
I'm very open to feedback on anything with the positioning and size but I think this is an improvement over the banner.
I added an image example of what that looks like to the PR description.

@faiza-a I was also told you might be able to help with the design Faiza so I've tagged you here too if that's ok.

faiza-a:
Here is one option. Let me know if this works. I can come up with other options if needed
Alert

mslarae13:
I like what Faiza has provided. I have one concern which is "I've submitted all the required information, but have more to add". Can we add something in that box that says 'I want to add more metadata'.. so basically the little x but clear that you're not done. Also, if I just click anywhere on the page, I'd like the box to disappear. & allow me to keep editing.

JamesTessmer:
That design looks good to me. @naglepuff and I had talked some about the level of intrusiveness, so I just want to double check that as well.
I think Montana’s point is good too, so if we update the text (That’s on me I think) to say something like “validation was successful, you can now submit or continue editing and adding data”.

JamesTessmer:
@faiza-a I think after a quick discussion we all liked the design you showed so we will use that. Is there vue code for the design? I haven't done much UI work so I'm not sure if I need to recreate it.
Thank you!

faiza-a:
@JamesTessmer  Figma can export code but it will be plain HTML and CSS (not the Vue components) Also i am not sure how neat the code would be. Here is the[ Figma link](https://www.figma.com/file/KUtfSqxhYCn334xkoh2CBr/Submission-Portal?node-id=1939%3A35&mode=dev), you can hover over components and see their info (colors, sizing, spacing etc.). But I think you will have to pretty much write Vue code from scratch.

JamesTessmer:
Ah ok, thanks for sharing the Figma link, I think that will be helpful!

JamesTessmer:
@mslarae13 @naglepuff 
I've updated the alert to be a snackbar that pops up for 3 seconds and then disappears (image above).
I tried right aligning it to be above the button to match Faiza's design but I couldn't ever get it to display above the button and it would always display on top and block the submit button while it was on screen.

It does seem like most of the notification/alert/snackbar type components don't support specific placement, so aligning above the button like in Faiza's design might be very difficult but I can keep looking at that if needed. Or if I'm missing something obvious let me know and I can fix the alignment.

Otherwise I think the snackbar is the best iteration of this so far and I think (if getting the alignment correct is possible) it captures the spirit of Faiza's design.

JamesTessmer:
Leah suggested replacing the snackbar notification with updating the existing box that shows up around the validate button when there are errors (shown below for clarity of what I'm talking about) to also show up when validation is successful and instead of having the error text it would just say what the snackbar notification says now.
I looked into doing that and I like the idea, but it's a little beyond my Vue knowledge at the moment. There's a lot of components that go into the error box. I think this might be worth revisiting in the future though.

image

JamesTessmer:
@mslarae13 @naglepuff I've marked this as ready for review now!

mslarae13:
I'm good with the green & can provide testing when it's in dev. As for code review, I'm mostly useless :)

web/src/views/SubmissionPortal/HarmonizerView.vue
```diff
@@ -889,7 +896,7 @@ export default defineComponent({
             v-on="on"
           >
             <v-btn
-              color="primary"
+              v-bind:color="type==='canSubmit' ? 'primary' : 'green'"
```
naglepuff, line 899: What is `type` here meant to be? I see some errors in my console on this branch related to it not being available here. I think this could just be

```suggestion
              color="green"
```
Side note: there is a [shorthand](https://vuejs.org/api/built-in-directives.html#v-bind) for `v-bind`, which is what we opt for in this code base.

web/src/views/SubmissionPortal/HarmonizerView.vue
```diff
@@ -544,6 +547,11 @@ export default defineComponent({
             mdi-refresh
           </v-icon>
         </v-btn>
+        <v-snackbar v-model="snackbar"
```
naglepuff, line 550: I'm seeing some console errors due to the fact that `snackbar` is a `ComputedRef` and is therefor read-only (unless we write a [setter](https://vuejs.org/guide/essentials/computed.html#writable-computed) for it.

Generally, I'm not a fan of writeable computed properties, since it kind of breaks the whole idea of them being derived from other state.

I think the best course of action would be to use a non-computed ref (i.e. `const snackbar = ref(false);`) and update that to `true` in the `validate` function if ALL tabs are valid. That function would also have to set `snackbar` to `false` in the event that it finds validation errors.

web/src/views/SubmissionPortal/HarmonizerView.vue
naglepuff, line 1: Instead of `'green'` I think we should use the color `'success'`. It is more informative as a name. We don't explicity define a "success" color in [colors.js](https://github.com/microbiomedata/nmdc-server/blob/8092b6cda37fceda6a624d9a0c38a18825871bea/web/src/colors.js), so I think it just uses the Vuetify default (but we could define one at some point).

naglepuff (changes requested):
Hey @JamesTessmer, I have some suggestions for this PR. Apologies for the delay in getting back to you. Don't hesitate to reach out if you have any questions.

web/src/views/SubmissionPortal/HarmonizerView.vue
JamesTessmer: I'm not actually sure what the `type` variable is there either, I took the syntax from a tutorial for changing the color based on a variable. 
The intent for this line was to swap the color to green once a user is able to click the submit button. Does changing the color to green without the v-bind still accomplish that?

web/src/views/SubmissionPortal/HarmonizerView.vue
JamesTessmer: I have gotten mixed opinions on this one, I had initially used the `'success'` color but I got feedback during the infrastructure meeting that the color didn't match the rest of the page and to use `'green'`. I personally prefer '`success`' because like you said it is more informative so I'll swap it back to that.

naglepuff: I think if you add 
```
{
...
  success: green,
...
}
```
to the exported object in `colors.js`, you would use the color that people seemed to like more while also using a more informative name.

web/src/views/SubmissionPortal/HarmonizerView.vue
naglepuff: My suggestion will make the Submit button always appear as the "green" color. I'm not sure we'd need to have a different color if the user was unable to submit. We also use the "disabled" prop which should add additional styling to indicate that the user can't submit.

If you wanted to change color depending on the value of a variable, and the variable is already a boolean, you can just use that in the ternary operator:
```
:color="canSubmit ? 'green' : 'primary'"
```
Note the use of the shorthand for `v-bind` here. Vue will evaluate the expression in `""` and use that value for `color`. If `canSubmit` is `true`, the color will evaluate to `green`, else `primary`.

JamesTessmer: Ah ok that makes sense. I think with the "disabled" prop it's fine to leave it green like you said. I'll make that change. Thanks for explaining everything, it's very helpful to know the rationale.

JamesTessmer:
@naglepuff I've made the changes we talked about above, everything builds locally and passes tests. I tested the validation button and the snackbar and everything behaves as expected so I don't think I broke anything.

web/src/views/SubmissionPortal/HarmonizerView.vue
```diff
@@ -544,6 +554,12 @@ export default defineComponent({
             mdi-refresh
           </v-icon>
         </v-btn>
+        <v-snackbar
+          v-model="snackbar"
+          color="success"
+          timeout="3000">
+            Validation Passed! You can now submit or continue editing.
```
naglepuff, line 561: Almost there: the `>` on line 560 should be on its own line, at the same indentation as the corresponding `<`, and the text ("Validation Passed!...") is indented one level (2 spaces) too far.

web/src/views/SubmissionPortal/HarmonizerView.vue
```diff
@@ -213,6 +215,13 @@ export default defineComponent({
       if (!valid && !sidebarOpen.value) {
         sidebarOpen.value = true;
       }
+
+      if (valid) {
+        snackbar.value = true;
+      } else {
+        snackbar.value = false;
+      }
```
naglepuff, line 223: If I validate the current data harmonizer tab, I see the snackbar even if there is another tab with validation errors. This logic should look at all tabs and only trigger the snackbar if all tabs are valid.

naglepuff (changes requested):

web/src/views/SubmissionPortal/HarmonizerView.vue
JamesTessmer: Ah, that's a good point. I have only been testing with 1 tab.

JamesTessmer: @naglepuff I believe I've addressed this, I lifted the logic from `canSubmit` and applied to `snackbar`. I tested with some multi-tab data and it should behave properly.

web/src/views/SubmissionPortal/HarmonizerView.vue
JamesTessmer: This is fixed in the newest commit now

web/src/views/SubmissionPortal/HarmonizerView.vue
```diff
@@ -225,6 +228,12 @@ export default defineComponent({
         ...tabsValidated.value,
         [activeTemplateKey.value]: valid,
       };
+
+      let allTabsValid = true;
+      Object.values(tabsValidated.value).forEach((value) => {
+        allTabsValid = allTabsValid && value;
+      });
+      snackbar.value = allTabsValid;
```
naglepuff, line 236: This would be a good place to use [Array.prototype.every](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every)
```suggestion
      snackbar.value = Object.values(tabsValidated.value).every((value) => value);
```
Feel free to ignore/modify if you don't think this is more readable.

web/src/views/SubmissionPortal/HarmonizerView.vue
```diff
@@ -544,6 +554,13 @@ export default defineComponent({
             mdi-refresh
           </v-icon>
         </v-btn>
+        <v-snackbar
+          v-model="snackbar"
+          color="success"
+          timeout="3000"
+        >
+            Validation Passed! You can now submit or continue editing.
+        </v-snackbar>
```
naglepuff, line 563:
```suggestion
        >
          Validation Passed! You can now submit or continue editing.
        </v-snackbar>
```

Last one, I swear. The text is over-indented by 2 spaces

naglepuff (changes requested):

web/src/views/SubmissionPortal/HarmonizerView.vue
JamesTessmer: I think that's a reasonable fix. I think it's definitely less cluttered and still readable.

JamesTessmer: This is implemented now

naglepuff:
@JamesTessmer could you do a `git merge main` or `git rebase main`?

JamesTessmer:
> @JamesTessmer could you do a `git merge main` or `git rebase main`?

Done now!

naglepuff (approved):
LGTM!

JamesTessmer:
> LGTM!

@naglepuff Awesome! I don't actually have permission to merge, so if you can hit that (or if I there's someone I can talk to get permission) we can get everything closed out!

mslarae13:
I tested and it looks good and is ready for release to prod. I discovered no issues.

PR #1191, opened 2024-03-18
Add metagenomics long read option and logic to UI

pkalita-lbl:
Fixes https://github.com/microbiomedata/submission-schema/issues/168

These changes add a new "Metagenome (Long Read)" checkbox to the Multiomics Data screen in the submission portal. If selected, it enables a new corresponding tab in the DataHarmonizer screen. It all works analogously to the existing checkboxes and tabs.

It turned out that once that was implemented, switching between the Metagenome tab and the Metagenome (Long Read) tab surfaced a bug in the version of Handsontable used by DataHarmonizer@1.4.7 related to column headers not being updated correctly. This bug was fixed in the version of Handsontable used by the latest version of DataHarmonizer. 

Upgrading from DataHarmonizer 1.4.7 to 1.6.5 actually corresponds to quite a large jump in Handsontable (7.4.2 to 13.1.0). Despite that, things seems to generally work fine. The one issue I found was that in certain situations columns were not sized correctly based on their header content. This was resolved by:
1. Moving the import of DataHarmoinzer's CSS (which includes Handsontable's CSS) up to the root level
2. Changing the order of loading data and updating column settings when changing tabs.

---

Update!

Further testing of this branch revealed two issues which have been fixed.

1. Validation was not working because `DataHarmonizer.validate()` became an `async` function and the call to it in `HarmonizerApi.validate()` needed to be awaited. Debugging this also revealed an extraneous watcher on `activeInvalidCells` that was unnecessary.
2. Some interactions with the DataHarmonizer grid caused the `afterSelection` hook callback to throw an error due to passing in a `col` value of `-1`. I'm not 100% sure if this was a new bug caused by the DataHarmonizer upgrade or an existing one we just didn't notice before. Either way the fix was trivial.

### DIFF
--- a/web/src/views/SubmissionPortal/HarmonizerView.vue
+++ b/web/src/views/SubmissionPortal/HarmonizerView.vue
@@ -838,6 +838,17 @@ export default defineComponent({
               </v-icon>
             </v-btn>
           </div>
+          <div v-else>
+            <div class="mx-2">
+              <div class="text-h6 mt-3 font-weight-bold d-flex align-center">
+                Column Help
+                <v-spacer />
+              </div>
+              <p class="my-2 text--disabled">
+                Click on a cell or column to view help
+              </p>
+            </div>
+          </div>
         </v-navigation-drawer>
       </div>
     </div>
--- a/web/src/views/SubmissionPortal/HarmonizerView.vue
+++ b/web/src/views/SubmissionPortal/HarmonizerView.vue
@@ -10,7 +10,14 @@ import { api } from '@/data/api';
 import { urlify } from '@/data/utils';
 import useRequest from '@/use/useRequest';
 
-import { HarmonizerApi, HARMONIZER_TEMPLATES } from './harmonizerApi';
+import {
+  HarmonizerApi,
+  HARMONIZER_TEMPLATES,
+  EMSL,
+  JGI_MG,
+  JGT_MT,
+  JGI_MG_LR,
+} from './harmonizerApi';
 import {
   packageName,
   sampleData,
@@ -83,11 +90,6 @@ const ALWAYS_READ_ONLY_COLUMNS = [
   'dna_seq_project_name',
 ];
 
-// TODO: can this be imported from elsewhere?
-const EMSL = 'emsl';
-const JGI_MG = 'jgi_mg';
-const JGT_MT = 'jgi_mt';
-
 export default defineComponent({
   components: {
     FindReplace,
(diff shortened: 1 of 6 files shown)


pkalita-lbl:
Moving this back into draft. Discussed with @naglepuff and @mslarae13 and decided to not try and sneak this in before the 2023.3 release feature freeze. So we'll just let this PR chill until after that release has been made.

pkalita-lbl:
I think there may be some issues with validation caused by the DH upgrade. This is just a note to my future self to investigate that before marking this ready-to-review again.

pkalita-lbl:
@naglepuff Now that Release 2024.3 is out and I've tested this branch a little more locally, I think this is ready for review and merge.

mslarae13:
@ssarrafan @naglepuff said he could review on Monday. Can we add to this sprint?

ssarrafan:
> @ssarrafan @naglepuff said he could review on Monday. Can we add to this sprint?

@naglepuff are you still planning to review this this week?

naglepuff:
@ssarrafan Yes, I'm working on this now. I should be able to get this back to @pkalita-lbl by EOD, sorry for the delay

web/src/views/SubmissionPortal/HarmonizerView.vue
```diff
@@ -941,6 +955,10 @@ export default defineComponent({
 </template>
 
 <style lang="scss">
+// Handsontable attaches hidden elements to <body> in order to measure text widths. Therefore this
+// cannot be nested inside .harmonizer-style-container or else the measurements will be off.
+@import '~data-harmonizer/lib/dist/es/index';
```
naglepuff, line 960: This is probably the only thing here that gives me pause. I'd worry about collisions or weird behavior with Vuetify styles. I took a glance at this particular file and it look like everything is pretty well scoped using IDs and classes though, and local testing looked fine as well.

web/src/views/SubmissionPortal/HarmonizerView.vue
naglepuff, line 1: I think I might be out of the loop here, but what is going on with:
image

It looks like many columns under the JGI-Metagenomics header are both
1) Invalid
2) Read-only

Is this expected? The column help text indicates that this is intended behavior, I just don't know what the workflow is to resolve these validation errors.

web/src/views/SubmissionPortal/HarmonizerView.vue
pkalita-lbl: I know. I don't like it either which is why I tried to explain why it is apparently necessary in the comment.

web/src/views/SubmissionPortal/HarmonizerView.vue
pkalita-lbl: Yes it is expected. The current process is that users who use the JGI tabs will get a spreadsheet from Montana with those values filled in and they'll use the Import XLSX feature to populate those columns. In the future they will be populated by data pulled from an API provided by JGI.

naglepuff (approved):

PR #1192, opened 2024-03-19
Add homepage_website to consortia pages

marySalvi:
Adds homepage website to consortia pages above team/funding sources


web/src/components/TeamInfo.vue
```diff
@@ -82,6 +82,12 @@ export default defineComponent({
               />
             </span>
           </div>
+          <div
+            v-if="item.homepage_website"
```
naglepuff, line 86:
```suggestion
            v-if="item.homepage_website && item.homepage_website[0]"
```

Should we add this check since empty arrays in javascript are truthy?

naglepuff (changes requested):
One (probably paranoid) suggestion, but otherwise looks good.

web/src/components/TeamInfo.vue
marySalvi: good catch!

Issue #1195, opened 2024-03-28
All "Resources" navigation menu items point to incorrect URLs


eecavanna:
All of the URLs to which the items in the "Resources" menu in the Data Portal's navigation menu point, are incorrect.

![image](https://github.com/microbiomedata/nmdc-server/assets/134325062/91757a83-ca3d-4b1f-87da-8fc10c396885)

For example:

![image](https://github.com/microbiomedata/nmdc-server/assets/134325062/11cfaf8c-157d-47f8-9b99-10a2e8daffac)

For reference: The last time the `menus.ts` file was updated was 3 weeks ago, by me, via https://github.com/microbiomedata/nmdc-server/pull/1178.

Looks to me like something is treating something as a relative path instead of an absolute URL.

I'm labeling this as `priority:high` since GSP is next week.

eecavanna:
I reproduced this in my local development environment

![image](https://github.com/microbiomedata/nmdc-server/assets/134325062/361fcebb-9e3e-4967-9c60-1dc3cff0ba84)

eecavanna:
I think the root cause is that, in `menus.ts`, the links in that menu are specified using the `to` property instead of the `href` property:

https://github.com/microbiomedata/nmdc-server/blob/d1a0bb532aeec8781511c4af6ffcc7ee2bbb7773/web/src/menus.ts#L50-L73

#### For comparison

I think `href` is "for" absolute URLs:

https://github.com/microbiomedata/nmdc-server/blob/d1a0bb532aeec8781511c4af6ffcc7ee2bbb7773/web/src/menus.ts#L28

I think `to` is "for" relative paths:

https://github.com/microbiomedata/nmdc-server/blob/d1a0bb532aeec8781511c4af6ffcc7ee2bbb7773/web/src/menus.ts#L41

PR #1196, opened 2024-03-28
Fix navigation link URLs in "Resources" section

eecavanna:
In this branch, I switched from using the `to` property (which appears to me to have been designed for client-side routing only) to the `href` property (which appears to me to have been designed for server-side routing only). I also added some documentation about that difference in an attempt to make it less likely that someone make the same mistake in the future.

Fixes https://github.com/microbiomedata/nmdc-server/issues/1195


eecavanna:
FYI @pkalita-lbl. No review needed (I initially requested a review).

pkalita-lbl:
Oh, we should probably get this fix into production too huh? 😬 I can work on getting a hotfix (a true hotfix this time) branch going today with this change cherry picked into it.

PR #1208, opened 2024-04-10
Create github issue when user submits metadata

JamesTessmer:
The goal of this PR is to add automated github issue creation. 
Issues will contain:
- the submission ID
- the submitter
- the PI
- If data has already been generated
- the status of the submission
- the data types present in the submission
- the sample type
- the number of samples
- a section for notes on the submission

Current example of the output of the automation:

If data has been generated:

Screenshot 2024-05-03 at 2 09 40 PM

If no data has been generated:
Screenshot 2024-05-03 at 2 16 07 PM


nmdc_server/api.py
```diff
@@ -746,6 +751,33 @@ async def update_submission(
...
+    omicsprocessingtypes = multiomicsform['omicsProcessingTypes']
+    sampletype = submission.metadata_submission['templates']
+    sampledata = submission.metadata_submission['sampleData']
+    numsamples = 0
+    for key in sampledata:
+        numsamples = numsamples + len(sampledata[key])
+
+    payload = '{\n "title":'+f'"NMDC Submission: {submission.id}", \n "body":"Submitter: {user.orcid} \\n Submission ID: {submission.id} \\n Has data been generated: {datagenerated} \\n PI: {pi} {piorcid} \\n Status: Submitted -Pending Review \\n Data types: {omicsprocessingtypes} \\n Sample type: {sampletype} \\n Number of samples: {numsamples} \\n Note:", \n "assignees": ["JamesTessmer"], \n "labels":["testing"]'+'}'
+    
+    res = requests.post(url,cookies=cookies,data=payload,headers=headers)
```
pkalita-lbl, line 778: We should check the status code of the response and at least log something if it's not successful

nmdc_server/api.py
```diff
@@ -728,7 +730,10 @@ async def update_submission(
             status_code=400,
             detail="This submission is currently being edited by a different user.",
         )
-
+    #Create Github issue when metadata is being submitted
+    if submission.status == 'in-progress' and body_dict.get("status", None) == "Submitted- Pending Review":
+        create_github_issue(submission,user)
+    return
```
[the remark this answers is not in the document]
JamesTessmer: Ah yes that's correct I was using it for testing so I didn't have to make a new submission every time I wanted to test.
I will remove that for review.

nmdc_server/api.py
```diff
@@ -746,6 +751,33 @@ async def update_submission(
     crud.update_submission_lock(db, submission.id)
     return submission
 
+def create_github_issue(submission,user):
+
+    url = "https://api.github.com/repos/JamesTessmer/Issue-bot-testing/issues?state=all"
```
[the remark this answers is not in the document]
JamesTessmer: That's a good idea. I was trying to figure out the best way to set both of those because they definitely shouldn't be hard coded in. I'll implement this and the check.

nmdc_server/api.py
```diff
@@ -746,6 +751,33 @@ async def update_submission(
...
+
+    cookies = {'logged_in':'no'}
```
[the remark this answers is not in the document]
JamesTessmer: It was in the example I found for issue creation and I didn't think to check if it was necessary. I'll remove and if everything still works we can leave it out.

nmdc_server/api.py
```diff
@@ -746,6 +751,33 @@ async def update_submission(
...
+    multiomicsform = submission.metadata_submission['multiOmicsForm']
+    pi = studyform['piName']
+    piorcid = studyform['piOrcid']
+    datagenerated = contextform['dataGenerated']
+    omicsprocessingtypes = multiomicsform['omicsProcessingTypes']
+    sampletype = submission.metadata_submission['templates']
+    sampledata = submission.metadata_submission['sampleData']
+    numsamples = 0
+    for key in sampledata:
+        numsamples = numsamples + len(sampledata[key])
```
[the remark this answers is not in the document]
JamesTessmer: Ah ok, thanks for pointing that out, I didn't realize the samples could be represented multiple times across tabs. I'll add this fix.

nmdc_server/api.py
```diff
@@ -746,6 +751,33 @@ async def update_submission(
...
+
+    payload = '{\n "title":'+f'"NMDC Submission: {submission.id}", \n "body":"Submitter: {user.orcid} \\n Submission ID: {submission.id} \\n Has data been generated: {datagenerated} \\n PI: {pi} {piorcid} \\n Status: Submitted -Pending Review \\n Data types: {omicsprocessingtypes} \\n Sample type: {sampletype} \\n Number of samples: {numsamples} \\n Note:", \n "assignees": ["JamesTessmer"], \n "labels":["testing"]'+'}'
```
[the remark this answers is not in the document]
JamesTessmer: Those are good ideas. I was focused on getting everything working, so now I can make it readable. I haven't done much with json in python so I actually didn't know that existed but looking at the docs it does seem like it would fit here really well.

nmdc_server/api.py
JamesTessmer: Good call, I'll put that on the to do list for this.

nmdc_server/api.py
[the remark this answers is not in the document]
JamesTessmer: Added this fix today.

nmdc_server/api.py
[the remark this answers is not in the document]
JamesTessmer: Removed this line and. everything still functions as expected.

nmdc_server/api.py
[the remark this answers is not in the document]
JamesTessmer: Added this change today.

nmdc_server/api.py
JamesTessmer: I added some logging for failures to record the error code given and the reason for failure

nmdc_server/api.py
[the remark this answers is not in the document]
JamesTessmer: I've added this suggestion. Everything looks far more readable now!

nmdc_server/api.py
```diff
@@ -747,6 +755,58 @@ async def update_submission(
...
+        f"Sample type:{sampletype}",
+        f"Number of samples:{numsamples}",
+        "Note:",
+    ]
+    body_string = " \n ".join(body_lis)
+    payload_dict = {
+        "title": f"NMDC Submission{submission.id}",
+        "body": body_string,
+        "assignees": ["JamesTessmer"],
+        "labels": ["testing"],
```
pkalita-lbl, line 795: Just wanted to confirm that you're intentionally signing up to have all these assigned to you?

pkalita-lbl (approved):
The changes look great!

JamesTessmer:
> The changes look great!

Great!
There are probably 1-2 minor touch ups to fix in terms of the content or formatting that I want to double check with Montana but those shouldn't affect anything.

nmdc_server/api.py
JamesTessmer: That is a good point, that's one of the last things that needs to be done. 
I need to talk to Montana about who (if anyone) these should be assigned to and any labels they need.

JamesTessmer:
@mslarae13 @pkalita-lbl 
I've pushed up some new things and I think this will be the last of it.
After talking to Montana I added a new function that posts the created issue to a project board. I tried to anticipate similar fixes I had to make to the issue creation function so this one should (hopefully) be pretty close to done.
I also added 2 new variables to the .env and config to support posting an issue to a project board. 

There are is also a new block of code in the issue creation to pull some IDs based on whether data has or has not been created for that submission. 

I don't anticipate either of you looking at this until next week, but when you do hopefully everything is good or only requires minor fixes that can be done super quick.

nmdc_server/api.py
```diff
@@ -747,6 +754,116 @@ async def update_submission(
...
+        f"PI orcid: {piorcid}",
+        "Status: Submitted -Pending Review",
+        f"Data types: {omicsprocessingtypes}",
+        f"Sample type:{sampletype}",
+        f"Number of samples:{numsamples}",
+        data_ids[0],
+        data_ids[1],
+        data_ids[2],
+        "Note:",
+    ]
```
pkalita-lbl, line 811: I would recommend instead:

```python
    body_lis = [
        f"Submitter: {user.name}, {user.orcid}",
        f"Submission ID: {submission.id}",
        f"Has data been generated: {datagenerated}",
        f"PI name: {pi}",
        f"PI orcid: {piorcid}",
        "Status: Submitted -Pending Review",
        f"Data types: {omicsprocessingtypes}",
        f"Sample type:{sampletype}",
        f"Number of samples:{numsamples}",
    ] + data_ids
```

Then it doesn't depend on a specific length of `data_ids`. That also drops the "Note:" line which never appears to be populated with anything.

nmdc_server/api.py
```diff
@@ -747,6 +754,116 @@ async def update_submission(
...
+    body_string = " \n ".join(body_lis)
+    payload_dict = {
+        "title": f"NMDC Submission: {submission.id}",
+        "body": body_string,
+        "assignees": ["mslarae13"],
```
pkalita-lbl, line 816: Can the assignee come from `Settings` as well? I'm just thinking that if someone wants to test this locally they probably don't want to spam Montana with issue assignments.

nmdc_server/api.py
```diff
@@ -747,6 +754,116 @@ async def update_submission(
...
+    return res
+
+
+def github_issue_to_project(issue_node_id: str, settings):
+    gh_project_token = settings.gh_project_token
+    gh_project_id = settings.gh_project_id
+
+    # Same as github issue, if we're missing the settings then we return.
+    if gh_project_token == None or gh_project_id == None:
+        logging.error("Posting issue to projec board failed. Missing token or id. See config")
```
pkalita-lbl, line 843:
```diff
-projec
+project
```

nmdc_server/api.py
```diff
@@ -747,6 +754,116 @@ async def update_submission(
...
+        # if issue creation is successful we want to put the issue on a project board
+        # if details for that are supplied
+        issue_node_id = res.json()["node_id"]
+        github_issue_to_project(issue_node_id, settings)
+
+    return res
+
+
+def github_issue_to_project(issue_node_id: str, settings):
+    gh_project_token = settings.gh_project_token
```
pkalita-lbl, line 838: This step requires a different token than `settings.github_authentication_token`?

pkalita-lbl (commented):
A couple of comments/questions based on the latest changes. Also looks like there's a conflict to resolve now, in case you hadn't seen that.

nmdc_server/api.py
JamesTessmer: Frustratingly it does seem to required a different token.
Fine grain access tokens don't seem to support the GraphQL api calls needed (found an issue [here](https://github.com/actions/add-to-project/issues/289)) unless they are organization accounts.
And I couldn't find a setting on the classic access token that allowed posting issues. The closest I saw was writing to team conversations but that wouldn't work for this.

I would be happy to be wrong and see that PAT do allow posting issues and then we could just use one token.

nmdc_server/api.py
JamesTessmer: That's a good point. I'll add that in.

nmdc_server/api.py
JamesTessmer: Good idea, I will implement that.

nmdc_server/api.py
JamesTessmer: Fixed this typo now.

JamesTessmer:
@pkalita-lbl I've removed all the project posting related code now and it looks like everything has passed checks. 
If everything looks good can you hit merge? I don't have permission to merge anything in this repo.

PR #1212, opened 2024-04-18
Refrain from reporting "0 results" and "0 samples" while loading data

eecavanna:
#### Old behavior

Previously, while sample data was being loaded, the upper left of the page would say "Found 0 results". Similarly, the top of the "Samples" panel would say "0 Samples".

image

#### New behavior

In this PR, I made it so that, while sample data is being loaded, the message in the upper left says "Loading results..." and the top of the "Samples" panel doesn't report a number.

image

#### Side note

> Note: The horizontal misalignment of the "Samples" (or "x Samples") heading with respect to the "Consortia" and "Studies" headings is not part of this PR (neither introduced in this PR nor fixed in this PR).

### DIFF
--- a/web/src/views/Search/SearchLayout.vue
+++ b/web/src/views/Search/SearchLayout.vue
@@ -203,7 +203,10 @@ export default defineComponent({
 
 <template>
   <div>
-    <SearchSidebar :results-count="biosample.data.results.count" />
+    <SearchSidebar
+      :results-count="biosample.data.results.count"
+      :is-loading="biosample.loading.value"
+    />
     <v-main>
       <v-progress-linear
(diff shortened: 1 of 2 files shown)


eecavanna:
Hi @aclum, are you OK with this change being merged into `main` (and then subsequently deployed to production as part of the next Release)? I'm asking you because I saw you are listed as the data portal product owner (on the NMDC website) and I'm looking for someone appropriate to sign off on the aesthetic aspect of this change.

aclum:
@eecavanna  yes, this looks good.

eecavanna:
Merged — thanks!

Issue #771, opened 2022-08-18
Reset scroll when you visit details page for study/sample


jeffbaumes:
To reproduce the behavior
1) click on any nmdc study
2) scroll to the bottom of the study page
3) use the back button on the study page (not browser back)
4) select another study or the same study
5) study page displays with the scroll bar all the way to the bottom.

aclum:
@jeffbaumes  do you have more info or a screen/video of the problem? is this still a problem?

naglepuff:
Still an issue

https://github.com/microbiomedata/nmdc-server/assets/7085625/0da17f7a-6108-430a-a141-a9db5bafb66e

aclum:
@eecavanna do you have time to work on this next sprint?

ssarrafan:
@JamesTessmer is going to take this one. Adding him and putting on the next sprint board.

JamesTessmer:
@eecavanna Thanks for all that work! I think that will be a really good starting point to jump into this.

JamesTessmer:
Here is the [pull request](https://github.com/microbiomedata/nmdc-server/pull/1225) for this issue.
After adding the fix the issue seems to be resolved.
Eric did most of the work finding the code and docs above, there was just a little bit of extra work to be done on where to place that line of code.

JamesTessmer:
@ssarrafan The branch has been merged in, so I think we can mark this as done!

PR #1225, opened 2024-04-23
Fix for bug where studies would not scroll to return to the top of the page when revisited

JamesTessmer:
Addressing [this issue](https://github.com/microbiomedata/nmdc-server/issues/771)

We've added a scroll behavior line of code to return to the top of the study page when revisiting that page.

### DIFF
--- a/web/src/plugins/router.ts
+++ b/web/src/plugins/router.ts
@@ -104,6 +104,7 @@ const router = new VueRouter({
       component: UserPage,
     },
   ],
+  scrollBehavior: () => ({ x: 0, y: 0 }),
   parseQuery,
   stringifyQuery,
 });


eecavanna (approved):

JamesTessmer:
@eecavanna Can you hit the merge button, I don't have permission to merge in this repo 😞

eecavanna:
On second thought, I was already told that (from a collaboration standpoint) the Kitware team members are generally OK with me merging once I see that green "Merge" button (which I do). I'll go ahead and merge this in. If people notice scroll behavior that they don't like, they can create an issue about it.

JamesTessmer:
@eecavanna Thank you!

PR #1229, opened 2024-04-29
Add nmdc-server CLI subcommand for loading local database with production data

### LINKED ISSUE #1228: Add ability to load Postgres database from nightly production backup on NERSC
Currently the [developer docs](https://github.com/microbiomedata/nmdc-server/blob/main/docs/development.md#postgres-import-and-export) on dumping and restoring the Postgres database that are a bit out of date. Specifically we should take advantage of the fact that we now have daily backups of the production Postgres database. With those we should be able to write something that fully automates reloading a local database from the production backup.

pkalita-lbl:
Fixes #1228 

### Summary

* The main addition is the new `click` subcommand named `load-db` in `nmdc_server/cli.py`. This subcommand is designed to 1) get a list of available backups on NERSC in the `/global/cfs/cdirs/m3408/pgdump` directory, 2) make a local copy of the latest one (or a specified one), 3) run `pg_restore` on the _local_ database using the copied backup.
* I wanted to make sure that this workflow worked:
  1. Completely destroy entire local Postgres database (or perhaps this is the first time ever running `nmdc-server`)
  2. Run the new `load-db` command to populate the database
  3. Run `docker compose up` to start the whole stack

  The issue I ran into was that the `nmdc_a` and `nmdc_b` databases are currently being created in the script (`prestart.sh`) that runs just before the web server application starts. This means that they would not be created yet at step 2 above. Additionally, restoring from the production backup requires the `nmdc_data_reader` role to be present; currently that has to be created manually. 

  In these changes both issues are resolved by adding `scripts/docker/init.sql` and mounting it into the `/docker-entrypoint-initdb.d` directory of the Postgres container. (See also: https://github.com/docker-library/docs/tree/master/postgres#initialization-scripts). This ensures that the database is properly initialized, even when started for the first time outside of the rest of the stack. 
* I updated `docs/development.md` to document the new subcommand and remove the old outdated information.

### DIFF
--- a/nmdc_server/cli.py
+++ b/nmdc_server/cli.py
@@ -238,12 +238,13 @@ def load_db(key_file, user, host, list_backups, backup_file):
 
         backup_file = dumps[-1]
 
-    click.echo(f"Downloading {backup_file}...")
-    subprocess.run(
-        ["scp", "-i", key_file, f"{user}@{host}:{pgdump_dir}/{backup_file}", "."],
-        check=True,
-        encoding="utf-8",
-    )
+    if not (Path(os.getcwd()) / backup_file).exists():
+        click.echo(f"Downloading {backup_file}...")
+        subprocess.run(
+            ["scp", "-i", key_file, f"{user}@{host}:{pgdump_dir}/{backup_file}", "."],
+            check=True,
+            encoding="utf-8",
+        )
 
     click.echo(f"Restoring from {backup_file}...")
     settings = Settings()
--- a/nmdc_server/cli.py
+++ b/nmdc_server/cli.py
@@ -189,8 +189,10 @@ def shell(print_sql: bool, script: Optional[Path]):
 @click.option(
     "-f",
     "--backup-file",
-    help=("Filename in NERSC's backup directory to load. "
-          "If not provided, the latest backup will be loaded."),
+    help=(
+        "Filename in NERSC's backup directory to load. "
+        "If not provided, the latest backup will be loaded."
+    ),
 )
 @click.option(
     "-k",
--- a/nmdc_server/cli.py
+++ b/nmdc_server/cli.py
@@ -1,4 +1,7 @@
 import logging
+import os
+import subprocess
+import sys
 from pathlib import Path
 from typing import Optional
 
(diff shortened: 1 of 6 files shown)


prestart.sh
```diff
@@ -1,17 +1,15 @@
...
 while ! PGDATABASE=postgres psql --quiet -c "select * from user" ; do
     sleep 1;
 done
 
-echo 'Creating and migrating database'
-# in spin the database already exists
-PGDATABASE=postgres psql -c "create database nmdc_a;" || true
-
-
-PGDATABASE=postgres psql -c "create database nmdc_b;" || true
```
naglepuff, line 15: Does removing this add an extra step if setting up a new environment in spin? I know that doesn't happen often but I don't see a real reason to remove these lines if they might still have utility outside of the development `docker-compose` workflow.

nmdc_server/cli.py
```diff
@@ -177,3 +180,91 @@ def shell(print_sql: bool, script: Optional[Path]):
...
+            sys.exit(1)
+
+        backup_file = dumps[-1]
+
+    click.echo(f"Downloading {backup_file}...")
+    subprocess.run(
+        ["scp", "-i", key_file, f"{user}@{host}:{pgdump_dir}/{backup_file}", "."],
+        check=True,
+        encoding="utf-8",
+    )
```
naglepuff, line 246: Do you think its worth checking to see if the most recent backup already exists in the container first?

scripts/docker/init.sql
```diff
@@ -0,0 +1,4 @@
+create role nmdc_data_reader;
+
+create database nmdc_a;
+create database nmdc_b;
```
naglepuff, line 4: FYI I did need to do a `docker compose down -v` in order for this script to run. Logs suggest that this script is skipped if there is any DB already present. Might be worth mentioning in the docs that existing users will see an error if the `nmdc_data_reader` role doesn't exist locally for them.

naglepuff (commented):
This is a great idea. So much faster and sturdier than doing a local ingest. This will be a good tool to have for a lot of cases. I had a few comments, but nothing major that would necessarily require a change.

prestart.sh
pkalita-lbl: I waffled on removing those lines. On one hand I don't think it's really the _right_ place to do database initialization, but on the other hand I guess it's maybe a useful failsafe measure. And they don't really hurt anything other than adding a bit of noise to the application logs at startup. Okay, I guess I've convinced myself to put them back in.

PR #1236, opened 2024-05-03
Only show AppBanner component on submission portal Home page

### LINKED ISSUE #1235: DataHarmonizer screen has unwanted whitespace when AppBanner component is displayed
Something about the presence of the `<AppBanner>` component is throwing off the layout of the DataHarmonizer screen in the submission portal

image

This isn't terribly surprising given that DataHarmonizer relies on a lot of `vh` units and `calc`s with hard-coded pixel values to determine its position and sizing. 

After discussing on Slack with @mslarae13 and @naglepuff I think we're going to make a hotfix to only show the banner on the Home screen of the submission portal. Then later we can try and identify and resolve the real root cause of the issue.

pkalita-lbl:
Fixes #1235 

This is ready for review, but I'm leaving it as a draft until I've made a new release tag on this branch.

### DIFF
--- a/web/src/views/SubmissionPortal/SubmissionView.vue
+++ b/web/src/views/SubmissionPortal/SubmissionView.vue
@@ -1,5 +1,6 @@
 <script lang="ts">
 import {
+  computed,
   defineComponent, PropType, toRef, watch,
 } from '@vue/composition-api';
 import { stateRefs } from '@/store';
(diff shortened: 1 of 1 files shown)


web/src/views/SubmissionPortal/SubmissionView.vue
```diff
@@ -1,5 +1,6 @@
 <script lang="ts">
 import {
+  computed,
   defineComponent, PropType, toRef, watch,
```
naglepuff, line 4:
```suggestion
  computed,
  defineComponent, 
  PropType, 
  toRef, 
  watch,
```

If you could just make the import statement's style more consistent I'd appreciate that.

PR #1253, opened 2024-05-15
Allow Github issue creation to track all IDs

JamesTessmer:
This PR is to update the github issue creation. 
Previously I had misunderstood that the IDs needed would be either JGI/EMSL OR Gold/NCBI but we want to track all of them.

This change checks to see if these ID fields are populated and if they are it adds them to the issue creation.

### DIFF
--- a/nmdc_server/api.py
+++ b/nmdc_server/api.py
@@ -796,20 +796,15 @@ def create_github_issue(submission, user):
         numsamples = max(numsamples, len(sampledata[key]))
 
     # some variable data to supply depending on if data has been generated or not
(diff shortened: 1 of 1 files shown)


JamesTessmer:
@pkalita-lbl This is just a small update to the content of the github issue. 
I had misunderstood what Montana wanted with the IDs so there's some extra logic to be more modular with what IDs are shown on the issue.

I've tested it and everything works as it should with IDs being added to the issue if they're in the omics form at all.

pkalita-lbl (approved):

PR #1262, opened 2024-05-21
Guard against request errors for PI images

marySalvi:


naglepuff (approved):
Changes look good. If you get the chance before merging could you open an issue describing the reason for the change and link it to the PR? I did a quick search in open issues for "PI" and "image" and I couldn't find anything.

eecavanna:
> If you get the chance before merging could you open an issue describing the reason for the change and link it to the PR?

Today, I created such an issue at https://github.com/microbiomedata/nmdc-server/issues/1397.

nmdc_server/ingest/study.py
```diff
@@ -23,11 +23,13 @@ def get_or_create_pi(db: Session, name: str, url: Optional[str], orcid: Optional
...
     if url:
-        r = requests.get(url)
-        if r.ok:
-            image_data = r.content
+        try:
+            r = requests.get(url)
+            r.raise_for_status()
+        except Exception as e:
+            logger.error(f"Failed to download image for {name} from {url} : {e}")
         else:
```
eecavanna, line 31: I was wondering why `else` was being used here instead of adding the additional instruction to the `try` block. I see an explanation in the Python docs:

> The use of the else clause is better than adding additional code to the [try](https://docs.python.org/3/reference/compound_stmts.html#try) clause because it avoids accidentally catching an exception that wasn’t raised by the code being protected by the try … except statement.

That makes sense to me. Noting it here for others' reference.

naglepuff (approved):

PR #1263, opened 2024-05-21
Enable tracing in sentry

### LINKED ISSUE #1264: Enable tracing in sentry
This will require updating some of our dependencies.

naglepuff:
Fix #1264 

Attempt to [capture some sentry transactions](https://github.com/microbiomedata/nmdc-server/pull/580) for analysis. Note how in additional to turning tracing on, several dependencies were updated. According to [sentry documentation](https://docs.sentry.io/platforms/python/integrations/fastapi/#supported-versions), we did need to upgrade our FastAPI version. Since I couldn't determine a concrete reason for the [last pin of FastAPI](https://github.com/microbiomedata/nmdc-server/pull/580), I decided to use the latest release. 

There were some other dependencies that needed a bump to accommodate this, and I also decided to undo a pin of `dnspython` (https://github.com/microbiomedata/nmdc-server/pull/1015/files) which was done as a band-aid fix to get a build to work. `dnspython` is past the version that caused build problems. This update will allow us to close some dependabot PRs as well.

If reviewers feel like this is too big of a jump I can back off down to the minimum version as required by `sentry-sdk[fastapi]`.


jeffbaumes:
Looks like there are some failing tox checks.

But I don't see any fundamental issues in moving forward. If we can get the server working well after bumping things up to the latest versions, we should go ahead and do so.

naglepuff:
Closing in favor of #1420

PR #1267, opened 2024-05-28
Fix DataHarmonizer scrolling issue

pkalita-lbl:
Fixes https://github.com/microbiomedata/issues/issues/701

### Testing instructions

* Create a new submission portal entry or open an existing one that you can edit, just ensure that in Step 4 (Multi-omics Data) the Metagenome checkbox under JGI is selected
* On Step 6 (the DataHarmonizer screen), ensure there are at least ~100 rows are filled out with at least an ID and that they all have "metagenomics" filled out in the "analysis/data type" column
* Click on the "JGI MG" tab
* Verify that you can scroll to the bottom and see all rows

### Details

Previously we set a fixed height directly on the element (`#harmonizer-root`) where the DataHarmonizer instance is mounted. This was a bit janky because the `362px` offset is a bit too brittle to changes in surrounding elements _and_ it doesn't interact well with DataHarmonizer's (sort of odd) choice to have the [height](https://github.com/cidgoh/DataHarmonizer/blob/cc21dc8f7dff3e7fbb9d9cdb2e560ef6f29707fa/lib/DataHarmonizer.js#L504) of its Handsontable instance be `75vh`. So these changes:

* Override DataHarmoinzer's `75vh` choice by telling the Handsontable instance to have a width and height setting of `auto`.
* Make the entire DataHarmonizer screen have a fixed heigh based on the height of the site's `v-app-bar` -- both controlled by the same variable.
* Set various flexbox and overflow settings to let the actual DataHarmonizer container shrink/grow to take up as much space as it can within the bounds of the fixed height element.

This also fixes the minor annoyance that when the red validation result box appears it doesn't push the row with the color key and submit button slightly off the bottom of the page.

### DIFF
--- a/web/src/views/SubmissionPortal/HarmonizerView.vue
+++ b/web/src/views/SubmissionPortal/HarmonizerView.vue
@@ -1027,7 +1027,6 @@ html {
 #harmonizer-root {
   width: 100%;
   height: 100%;
-  overflow: auto;
 
   .secondary-header-cell:hover {
     cursor: pointer;
--- a/web/src/views/SubmissionPortal/HarmonizerView.vue
+++ b/web/src/views/SubmissionPortal/HarmonizerView.vue
@@ -769,9 +769,7 @@ export default defineComponent({
     <div
       class="harmonizer-style-container harmonizer-and-sidebar"
     >
-      <div class="harmonizer-container">
-        <div id="harmonizer-root" />
-      </div>
+      <div id="harmonizer-root" />
 
       <div
         :style="{
@@ -1024,14 +1022,11 @@ html {
   overflow: auto;
 }
 
-.harmonizer-container {
-  flex-grow: 1;
-  overflow: auto;
-}
-
 /* Grid */
 #harmonizer-root {
+  width: 100%;
   height: 100%;
+  overflow: auto;
 
   .secondary-header-cell:hover {
     cursor: pointer;
--- a/web/src/views/SubmissionPortal/HarmonizerView.vue
+++ b/web/src/views/SubmissionPortal/HarmonizerView.vue
@@ -36,6 +36,7 @@ import FindReplace from './Components/FindReplace.vue';
 import SubmissionStepper from './Components/SubmissionStepper.vue';
 import SubmissionDocsLink from './Components/SubmissionDocsLink.vue';
 import SubmissionPermissionBanner from './Components/SubmissionPermissionBanner.vue';
+import { APP_HEADER_HEIGHT } from '@/components/Presentation/AppHeader.vue';
 
 interface ValidationErrors {
   [error: string]: [number, number][],
@@ -456,6 +457,7 @@ export default defineComponent({
     }
 
     return {
+      APP_HEADER_HEIGHT,
       ColorKey,
       HARMONIZER_TEMPLATES,
       columnVisibility,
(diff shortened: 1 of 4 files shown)


web/src/views/SubmissionPortal/HarmonizerView.vue
```diff
@@ -1017,13 +1014,20 @@ html {
...
+
 /* Grid */
 #harmonizer-root {
-  overflow: hidden;
   width: 100%;
-  height: calc(100vh - 362px) !important;
-  float: left;
-  margin-top: 8px;
+  height: 100%;
+  overflow: auto;
```
marySalvi, line 1030: In my opinion removing the `overflow: auto` here makes a for a cleaner scroll
[Screencast from 06-05-2024 04:03:05 PM.webm](https://github.com/microbiomedata/nmdc-server/assets/40494088/634ff2e8-f1c9-44ef-92be-34b407e01c76)

web/src/views/SubmissionPortal/harmonizerApi.ts
```diff
@@ -209,9 +209,16 @@ export class HarmonizerApi {
...
+      // we use our own custom help sidebar, so turn off DataHarmonizer's built-in one
+      helpSidebar: {
+        enabled: false,
+      },
     });
     this.footer = new Footer(document.querySelector('#harmonizer-footer-root'), this.dh);
     this.dh.useSchema(this.schema, [], templateName);
+    // Let the Handsontable element take up as much space as it wants, it will be visually
+    // constrained by its containing elements
+    this.dh.hot.updateSettings({ height: 'auto', width: 'auto' });
```
marySalvi, line 221: I think you might want `100%` here instead of `auto`. `auto` will make the height and width dependent on the children and `100%` will make it dependent on it's parent. Disregard if I misunderstood the comment.

marySalvi (changes requested):
A couple small suggestions but otherwise looks good.

web/src/views/SubmissionPortal/HarmonizerView.vue
pkalita-lbl: Whoa, what OS is that? That looks _way_ worse than on my Mac. The nested scrolling areas is exactly what we're trying to avoid here. I might need to pull out a VM and test on other platforms if that's the case.

pkalita-lbl: Sorry I had to look at that video a second time. I got so distracted by the visual scrollbars (and the fact that there were two sets of them) that I think I missed the real point that with my changes the header rows aren't staying visible. That's definitely not right and I'll definitely need to address that before going forward. But was there anything else you meant by "a cleaner scroll"?

pkalita-lbl:
Okay addressing @marySalvi's comments:

* Yes, `100%` is a better option than `auto`. That was the reason that the header row was being scrolled out of view in the video you posted. I guess I missed that because...
* The `height` and `width` settings on the Handsontable instance need to be set after every template update. I forgot that DataHarmonizer resets a bunch of Handsontable properties -- including `height` -- after you change templates. That's why the header _would_ stay stuck at the top after changing tabs.
* With those changes made, the `overflow: auto` on `#harmonizer-root` is actually superfluous so it can be removed.

Just to make the intention of these changes more clear, the green box should be the _only_ thing on the page that scrolls. Everything else on the page should be fixed. If any other element scrolls or if you ever get more than one set of scrollbars something is wrong.

image

marySalvi (approved):
Looks good to me!

mslarae13:
Tested in dev. Looks fixed! 
When should I expect to see the fix in prod?

pkalita-lbl:
[June 24](https://docs.google.com/document/d/1Lp4lDGlBzEhwq-CcjMPBu_4Hri27A6SkC5WUPG-fakg/edit#heading=h.bg5wdtzfc8u)

PR #1269, opened 2024-05-29
Show save progress

marySalvi:
closes https://github.com/microbiomedata/user_research/issues/18

This adds a `beforeChangeHook` the fires when a user enters new data, initiating a spinning progress component. When a 200 is returned from the patch a green check mark appears with text indicating successful save. If the patch does not return a 200 a red 'x' is shown with corresponding text.

### DIFF
--- a/web/src/views/SubmissionPortal/HarmonizerView.vue
+++ b/web/src/views/SubmissionPortal/HarmonizerView.vue
@@ -164,14 +164,20 @@ export default defineComponent({
       ])),
     ));
 
-    const onDataChange = () => {
+    const isSaving = ref();
+    const saveSuccess = ref();
+    const onBeforeChange = () => {
+      isSaving.value = true;
+    };
+    const onDataChange = async () => {
       hasChanged.value += 1;
       const data = harmonizerApi.exportJson();
       mergeSampleData(activeTemplate.value.sampleDataSlot, data);
-      incrementalSaveRecord(root.$route.params.id);
+      const httpStatus = await incrementalSaveRecord(root.$route.params.id);
       tabsValidated.value[activeTemplateKey.value] = false;
+      saveSuccess.value = httpStatus === 200;
+      isSaving.value = false;
     };
-
     const { request: schemaRequest, loading: schemaLoading } = useRequest();
     onMounted(async () => {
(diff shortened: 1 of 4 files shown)


web/src/views/SubmissionPortal/store/index.ts
```diff
@@ -275,7 +275,9 @@ async function incrementalSaveRecord(id: string) {
   }
 
   if (hasChanged.value) {
-    await api.updateRecord(id, payload, undefined, permissions);
+    const response = await api.updateRecord(id, payload, undefined, permissions);
+    // eslint-disable-next-line consistent-return
+    return response.httpStatus;
```
naglepuff, line 280: Instead of disabling linting for this line, what are your thoughts on updating the function signature to include a return type (I think it would be `Promise<number | undefined>`?

Feel free to ignore if that change would make the linter even angrier or if you disagree.

web/src/views/SubmissionPortal/HarmonizerView.vue
```diff
@@ -183,6 +189,7 @@ export default defineComponent({
         await harmonizerApi.init(r, schema, activeTemplate.value.schemaClass, goldEcosystemTree);
         await nextTick();
         harmonizerApi.loadData(activeTemplateData.value);
+        harmonizerApi.addBeforeChangeHook(onBeforeChange);
```
naglepuff, line 192: I think you might also need to add this in the `changeTemplate` function. It looks to me like if I change tabs in the data harmonizer step I stop seeing the new element change between loading and verifying.

naglepuff (changes requested):

web/src/views/SubmissionPortal/HarmonizerView.vue
marySalvi: Good catch!

web/src/views/SubmissionPortal/store/index.ts
marySalvi: In addition to adding this signature I had to add returns of resolved promises for all paths. 272638b26c79673dad9cbeadb16f846adbe3f9db

naglepuff (approved):
LGTM!

PR #1273, opened 2024-06-05
Update development.md

shreddd:
Add current redirect URIs to docs.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -27,6 +27,18 @@ NMDC_ORCID_CLIENT_ID=changeme
 NMDC_ORCID_CLIENT_SECRET=changeme
 ```
 
+### Production Redirect URIs
+
+These are the set of redirect URIs enabled in ORCiD. We need to file a support ticket with ORCID to modify these.
+
+- https://data.microbiomedata.org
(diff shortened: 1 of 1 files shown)


eecavanna (approved):
Thanks for documenting this. Given that the list includes some non-nmdc-server URLs, I'm wondering whether this list would fit better in the `infra-admin` repo long term. Anyway, I still think it's an enhancement to the current documentation here. It can be moved to `infra-admin` later.

naglepuff (approved):
I agree with @eecavanna's assessment that this documentation might serve us better by being elsewhere (like the infra-admin repo) but there's no harm in having it here.

shreddd:
OK - thanks - I agree with making a separate (and more complete) ORCID doc in the infra-admin repo. 

However, the issue of how nmdc-server uses the REDIRECT_URI in ORCID has come up a few times, and since we do have a section on ORCID auth here, I think it is useful to mention the valid URIs.

eecavanna:
I created an issue in the `infra-admin` repo just now, for transferring this list to that repo. Here's a link to the issue: https://github.com/microbiomedata/infra-admin/issues/87

PR #1275, opened 2024-06-13
Redirect users to submission portal if logging in from submission portal

naglepuff:
Fix microbiomedata/user_research#19

### Changes
Enable passing a state parameter to `initiateOrcidLogin()`, which is encoded in the login request and persisted by the server during the login process.

This handles two workflows where the user would need to be redirected:

1. When clicking the normal login button (the one in the app bar) while navigated to any route containing `submission`
2. When clicking the "login" button that appears on the submission homepage

In those 2 cases, the string `submission` is passed to `initiateOrcidLogin()`.

When we return from the server during the login process, we can expect a `state` query parameter if we sent one up at the start of login. The `init` function now takes this as an optional parameter and handles it accordingly. The only value for `state` that is handled differently as part of these changes is `submission`, in which case the `submission/home` route is pushed to the Vue router.

### To test
Make sure logging in from the submission home screen redirects you back to the submision home screen. Test for regressions by making sure that the filter state is persisted during normal (data portal) login.

### DIFF
--- a/web/src/components/Presentation/AuthButton.vue
+++ b/web/src/components/Presentation/AuthButton.vue
@@ -15,7 +15,18 @@ export default defineComponent({
     const router = useRouter();
 
     function handleLogin() {
-      api.initiateOrcidLogin();
+      if (!router) {
+        api.initiateOrcidLogin();
+        return;
+      }
+
+      const submissionState = 'submission';
+      const submissionRegex = new RegExp(submissionState);
+      let state = '';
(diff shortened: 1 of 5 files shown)


pkalita-lbl (approved):
Whoa, no backend changes. Cool!

mslarae13:
@naglepuff ready for testing on dev?

mslarae13:
quick test in dev & it worked!

Thansk @naglepuff

PR #1278, opened 2024-06-14
Add source to gh issue creation

JamesTessmer:
Small PR to add the environment the GH issue was created from.

### DIFF
--- a/nmdc_server/api.py
+++ b/nmdc_server/api.py
@@ -771,6 +771,7 @@ def create_github_issue(submission, user):
 
     # assemble the body of the API request
     body_lis = [
+        f"Issue created from {settings.environment} environment",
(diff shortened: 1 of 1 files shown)


pkalita-lbl (changes requested):
Oh no 🫠

All due props to you James for noticing that the `environment` setting exists. And in a logical world that would be the right thing to use. But unfortunately here in the stupid world we do live in we uhhhhhh don't actually set that variable consistently across environments. In fact, it's mostly just used for testing.

I think the `host` setting is going to be more reliable to use here.

JamesTessmer:
@pkalita-lbl 
I would be very surprised if there wasn't some kind of silly thing like that on every project haha.
I updated the line to have host now instead of environment.

pkalita-lbl (approved):

PR #1287, opened 2024-06-24
Rename `expires` to `expires_in` in TokenResponse

pkalita-lbl:
This came up in the discussion of an `nmdc-runtime` PR (https://github.com/microbiomedata/nmdc-runtime/pull/572). Even though we're not claiming to be a full OAuth 2.0-compatible authorization server, I had intended to remain close to the standard where possible. I think it was just an oversight that I had used `expires` instead of the [recommended](https://datatracker.ietf.org/doc/html/rfc6749#section-5.1) `expires_in` as a field in the `TokenResponse` class. 

Nothing in the front-end actually uses that value, so I think this is a fairly low-risk change. But I'd rather change it now before it hits production for the first time.

### DIFF
--- a/nmdc_server/auth.py
+++ b/nmdc_server/auth.py
@@ -75,7 +75,7 @@ class TokenResponse(BaseModel):
     access_token: bytes
     refresh_token: Optional[bytes] = None
     token_type: str = "bearer"
-    expires: int
+    expires_in: int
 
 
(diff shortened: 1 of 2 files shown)


eecavanna (approved):
Looks good to me! Thanks for closing the loop on this.

PR #1290, opened 2024-06-28
Implement endpoint that returns TSV file of submission summaries

eecavanna:
In this branch, I implemented an API endpoint at `/metadata_submission/report` that responds with a downloadable TSV file containing a high-level report of Submission Portal submissions. I am designing this endpoint with a specific use case in mind: quarterly metrics collection.

At this point, I have implemented the API endpoint and the underlying CRUD function.

The test code in this branch is still preliminary.

### DIFF
--- a/nmdc_server/api.py
+++ b/nmdc_server/api.py
@@ -1,6 +1,7 @@
+import csv
 import json
 import logging
-from io import BytesIO
+from io import BytesIO, StringIO
 from typing import Any, Dict, List, Optional
 from uuid import UUID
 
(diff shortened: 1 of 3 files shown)


naglepuff (approved):
This looks good to me! The new method in `crud.py` is slightly redundant, as the function `get_submissions_for_user` automatically returns all submissions if the given user is an admin. It does also add some sorting and ordering, so its not quite 1:1

That's mostly a nitpick, so I won't let it hold up this development. I tested locally and got a TSV file that looks like the description from the PR review.

pkalita-lbl (approved):

PR #1292, opened 2024-07-09
Add contact support panel

marySalvi:
Adds a contact support panel on the DH page and submission list page. Should match the styling in [figma](https://www.figma.com/design/KUtfSqxhYCn334xkoh2CBr/Submission-Portal?node-id=2002-72&t=vxvDnJJq7VoXRkov-0).

Closes https://github.com/microbiomedata/user_research/issues/20

### DIFF
--- a/web/src/views/SubmissionPortal/Components/SubmissionList.vue
+++ b/web/src/views/SubmissionPortal/Components/SubmissionList.vue
@@ -14,6 +14,7 @@ import OrcidId from '../../../components/Presentation/OrcidId.vue';
 import TitleBanner from '@/views/SubmissionPortal/Components/TitleBanner.vue';
 import IconBar from '@/views/SubmissionPortal/Components/IconBar.vue';
 import IntroBlurb from '@/views/SubmissionPortal/Components/IntroBlurb.vue';
+import ContactCard from '@/views/SubmissionPortal/Components/ContactCard.vue';
 
 const headers: DataTableHeader[] = [
   {
@@ -51,7 +52,7 @@ const headers: DataTableHeader[] = [
 
 export default defineComponent({
   components: {
-    IconBar, IntroBlurb, OrcidId, TitleBanner,
+    IconBar, IntroBlurb, OrcidId, TitleBanner, ContactCard,
   },
   setup() {
     const router = useRouter();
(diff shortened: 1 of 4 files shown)


web/src/views/SubmissionPortal/Components/SubmissionList.vue
```diff
@@ -99,81 +100,107 @@ export default defineComponent({
...
-              mdi-arrow-right-circle
-            </v-icon>
-          </v-btn>
-        </template>
-      </v-data-table>
+            mdi-message-question
+          </v-icon>
+        </v-btn>
+      </template>
+      <ContactCard />
```
naglepuff, line 126: Would it make sense to have the contact card be sticky? Currently it moves around as I scroll around the page. In extreme scenarios (changing the page size to be 50), I can scroll so it is off the screen. 

https://github.com/microbiomedata/nmdc-server/assets/7085625/f999b6a5-d75f-4c91-942b-0390e1c55c6b

naglepuff (commented):
Everything looks like its working as intended, I just had one thought about the placement on the submission list.

naglepuff:
Also make sure to link the associated issue

web/src/views/SubmissionPortal/Components/SubmissionList.vue
marySalvi: fixed in cad8fa53883b8dff5e229e07cdf7734676eaa46b

marySalvi:
@mslarae13 Note that I did not include the social media icons at the bottom. I did this for a few reasons:
1. One of the icons is for emailing and would be redundant of the rest of the component.
2. The other 3 are brand icons and there are strict rules for each of them
    a. As designed this component doesn't follow those rules.
    b. The icon library we currently use no longer supports brand icons because of these rules and how often they change
    c. From a brief search even icon libraries that do support these icons some (linkedIn) are already deprecated/out of date so they would need to be downloaded individually.
3. The top 'button' in the component brings a user to the Contact Page which has links for all of these socials so the actions are still available to a user interacting with this component. (Please note that the Contact Page does not follow the rules for there brand icons and this should perhaps be revisited)

Feel free to reach out to me if you have any questions regarding this.

naglepuff (approved):

mslarae13:
@marySalvi did you see @ljohnson09 's comment on the issue?

PR #1293, opened 2024-07-09
Change router to go to homepage instead of previous history index

### LINKED ISSUE #1250: Omics filter on study page redirects to other legacy study pages instead of faceted search
Clicking on the 'metagenome:103' from  https://data.microbiomedata.org/details/study/gold:Gs0128850 always redirects to another legacy study page instead of applying a study and omics type filter as a faceted search. 

Note if this is worth fixing since these will be re-ided shortly but wanted to note it.

marySalvi:
Fixes #1250

### DIFF
--- a/web/src/views/IndividualResults/StudyPage.vue
+++ b/web/src/views/IndividualResults/StudyPage.vue
@@ -150,7 +150,7 @@ export default defineComponent({
         }],
       );
       /* @ts-ignore */
-      router.go(-1);
+      router.go({ name: 'Search' });
     }
 
     function seeStudyInContext(item: StudySearchResults) {


naglepuff (approved):
LGTM!

I was able to reproduce the issue by taking the following steps:

1. In an environment where this change is not present (e.g. prod), navigate to the home page of the NMDC data portal
2. From the studies list, I opened two different studies in separate tabs (1000 Soils in Tab A and EMP500 in Tab B)
3. I copied/pasted the URL from Tab A into Tab B, so both tabs were now on the study page for 1000 Soils
4. I clicked on an omics chip in Tab B, and instead of going back to the search page, it went back (using history) to the EMP500 study page

I tried following the same steps in an environment with the changes and can confirm that navigation works as expected (you are taken to the home page with the appropriate search parameters)

PR #1294, opened 2024-07-09
Add user info page

### LINKED ISSUE #1277: Add simple user info page
For context, this was mentioned in the first paragraph of the Future Work section of https://github.com/microbiomedata/nmdc-server/pull/1257.

I propose we add a `/user` frontend route. This route should only be available for logged-in users. It should have a section for basic user info that displays what we store in the `user_logins` table -- their name, ORCID iD, and admin status. 

It should also have a section for API tokens that allows the user to copy their current refresh or access token to their clipboard and display the expiration time of those tokens. It should also make it clear that they should handle those tokens securely as if they are passwords and provide details on how to use the refresh token to obtain new access tokens and use access tokens to make API requests.

Currently the user's name in the navigation header is linked to their ORCID profile. I don't know if that's required by ORCID's standards. If not we could consider making that a link to the `/user` route instead. 

In the future we could consider expanding this page to collect user information that ORCID doesn't provide.

pkalita-lbl:
Fixes #1277 

### Summary

These changes add a new page served on `/user` that provides information about the currently logged-in user. For now this page is mainly to facilitate providing information about making authenticated API requests, but it could potentially be expanded in the future.

### Details

* The top of the new page displays the user's name and ORCID iD (linked to their ORCID profile). If they are an admin that is indicated next to their name.
* The new page's Developer Tools section allows a logged-in user to view and/or copy their existing Refresh Token and provides general instructions on how to use it in their application.
* The Swagger UI's header now has instructions for obtaining and using an Access Token
* The site header's auth button now links to the new User page instead of ORCID
* To show the Refresh Token's expiration time, the JWT needs to be decoded. I added `jwt-decode` as a new dependency to handle that. That package requires Node >= 18. This meant that I needed to update the Client Tests job in the main testing workflow to use a newer version of Node instead of 14. I have it set to use the latest LTS version, which is inline with what the frontend Dockerfile uses. The necessary `NODE_OPTIONS` environment variable was also added.

### DIFF
--- a/web/src/data/api.ts
+++ b/web/src/data/api.ts
@@ -701,6 +701,18 @@ function initiateOrcidLogin(state: string = '') {
 
 const REFRESH_TOKEN_KEY = 'storage.refreshToken';
 
+function getRefreshToken() {
+  return window.localStorage.getItem(REFRESH_TOKEN_KEY);
+}
+
+function setRefreshToken(token: string) {
+  return window.localStorage.setItem(REFRESH_TOKEN_KEY, token);
+}
+
+function clearRefreshToken() {
+  return window.localStorage.removeItem(REFRESH_TOKEN_KEY);
+}
+
 /**
  * Handle a token response by setting the API client's default Authorization header with the access
  * token and storing the refresh token (if provided) in local storage.
@@ -711,7 +723,7 @@ function handleTokenResponse(response: TokenResponse) {
   const { access_token, refresh_token } = response;
   client.defaults.headers.common.Authorization = `Bearer ${access_token}`;
   if (refresh_token) {
-    window.localStorage.setItem(REFRESH_TOKEN_KEY, refresh_token);
+    setRefreshToken(refresh_token);
   }
 }
 
@@ -743,7 +755,7 @@ async function logout() {
     });
   } finally {
     delete client.defaults.headers.common.Authorization;
-    window.localStorage.removeItem(REFRESH_TOKEN_KEY);
+    clearRefreshToken();
   }
 }
 
@@ -761,7 +773,7 @@ const REFRESH_REQUEST_MAX_AGE_MS: number = 1000 * 20;
  */
 function exchangeRefreshToken(): Promise<TokenResponse> {
   async function _doExchange(): Promise<TokenResponse> {
-    const refreshToken = window.localStorage.getItem(REFRESH_TOKEN_KEY);
+    const refreshToken = getRefreshToken();
     if (!refreshToken) {
       throw new RefreshTokenExchangeError('No refresh token found');
     }
@@ -771,7 +783,7 @@ function exchangeRefreshToken(): Promise<TokenResponse> {
       handleTokenResponse(data);
       return data;
     } catch (error) {
-      window.localStorage.removeItem(REFRESH_TOKEN_KEY);
+      clearRefreshToken();
       window.dispatchEvent(new CustomEvent(REFRESH_TOKEN_EXPIRED_EVENT, {
         detail: { error },
       }));
@@ -842,6 +854,7 @@ const api = {
   exchangeAuthCode,
   exchangeRefreshToken,
   logout,
+  getRefreshToken,
 };
 
 export {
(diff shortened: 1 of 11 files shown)


web/src/views/User/UserDetailPage.vue
```diff
@@ -0,0 +1,191 @@
...
+              cols="auto"
+            >
+              <v-text-field
+                label="Refresh Token"
+                readonly
+                filled
+                :type="isTokenVisible ? 'text' : 'password'"
+                :value="refreshToken"
+                @click="handleRefreshTokenInputClick"
+              >
```
naglepuff, line 115: Any thoughts on using a [text area component](https://dev.vuetifyjs.com/en/components/textareas/#auto-grow) or making the text area longer? The token is much longer than the current default width of the text field, which might make it difficult for users who miss the "copy to clipboard" button and try to highlight + copy.

naglepuff (commented):
Overall looks good. I just had a couple of suggestions.

web/src/views/User/UserDetailPage.vue
naglepuff: I'm pretty ambivalent about this one, BTW. In firefox on my OS (Ubuntu) I seemingly can't even highlight a partial bit of the token. On chrome It's easier to see that the token extends beyond the visual textbox (but I can highlight only a piece of it). A tooltip on the "copy to clipboard" button might be nice to have as well.

web/src/data/api.ts
```diff
@@ -701,6 +701,18 @@ function initiateOrcidLogin(state: string = '') {
...
+  return window.localStorage.getItem(REFRESH_TOKEN_KEY);
+}
+
+function setRefreshToken(token: string) {
+  return window.localStorage.setItem(REFRESH_TOKEN_KEY, token);
+}
+
+function clearRefreshToken() {
+  return window.localStorage.removeItem(REFRESH_TOKEN_KEY);
+}
```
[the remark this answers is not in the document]
pkalita-lbl: I think I had initially thought to put them in `web/src/store/index.ts` but that resulted in circular imports between it and `web/src/data/api.ts`. So I've now moved them to a new module, `web/src/store/localStorage.ts`. I also moved a bit of code from `web/src/store/index.ts` into the new module so that everything that interacts with `window.localStorage` is together there.

Down the road it would probably be good to properly type the get/setQueryState functions, but importing the `Conditions` interface from `web/src/data/api.ts` leads to circular imports again. Rather than _also_ moving a bunch of interface definitions around I just left it this way for now.

web/src/views/User/UserDetailPage.vue
```diff
@@ -0,0 +1,191 @@
...
+        isCopyRefreshTokenSnackbarVisible.value = true;
+      }
+    };
+
+    const handleRefreshTokenVisibilityButtonClick = () => {
+      isTokenVisible.value = !isTokenVisible.value;
+    };
+
+    const handleRefreshTokenInputClick = (event: MouseEvent) => {
+      event.preventDefault();
```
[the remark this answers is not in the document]
pkalita-lbl: Ah cool. Thanks for the Vue tip. I ended up removing `handleRefreshTokenInputClick` altogether; see below.

web/src/views/User/UserDetailPage.vue
pkalita-lbl: I did think about using a text area. I felt like my choices were 1) use a text input field and toggle the `type` attribute to obscure the token or 2) use a text area field and implement my own method of obscuring (since text areas don't have a `type` attribute). I opted for 1 in an attempt to keep things simpler. 

The `@click` handler on the input itself was an attempt at mitigating the issue that you can't see the whole token, and, worse still, double clicking on the text only selects up to the first period in the token. But as you pointed out maybe that interaction is a little confusing. So I've removed it in the latest changes and instead the entire text gets selected when you click the "reveal token" button.

Tooltips on both icon buttons have been added.

naglepuff (approved):
Thanks for addressing those initial bits of feedback. I think this works well, so let's merge this so people who are interested can start using it.

I tested to make sure the previous local storage stuff works (search filter state persists through the login flow) and that the user page behaves correctly when not logged in.

PR #1299, opened 2024-07-16
Add endpoint for explicitly requesting `SubmissionMetadata` lock

pkalita-lbl:
These changes are in support of having the Field Notes app respect the `locked_by` field on `SubmissionMetadata` records (https://github.com/microbiomedata/nmdc-field-notes/issues/90). There are two issues I'm trying to address here:

1. The Field Notes app attempts to cache and reuse responses from `GET` requests whenever possible. The underlying assumption of the caching layer is that `GET` requests have _no_ side effects. Currently, however, a request to `GET /metadata_submission/{id}` may also modify the record being returned by updating the `locked_by` field.
2. The Submission Portal has a unified view/edit interface for `SubmissionMetadata` records, whereas the the Field Notes app allows you to view a `SubmissionMetadata` before allowing the user to navigate to a separate edit interface. There would be no reason to read or write the `locked_by` field from the view interface.

To address both of those issues, these changes introduce a separate `/metadata_submission/{id}/lock` endpoint (called with a `PUT` method). This seemed like a natural counterpart to the existing `/metadata_submission/{id}/unlock` endpoint. Both of these endpoints now return a `LockOperationResult` object which indicates whether the lock or unlock was successful and who (if anyone) now holds the lock. Unsuccessful operations are also returned with a `409 Conflict` status. With the assumption that a client should _explicitly_ request the lock before making updates, there is no longer an attempt to claim the lock in the `GET /metadata_submission/{id}` handler.

On the frontend side there was only one function (see: `web/src/views/SubmissionPortal/store/index.ts`) where the `GET /metadata_submission/{id}` endpoint was invoked. So to keep the Submission Portal behaving the same way it did before, I added a call to `PUT /metadata_submission/{id}/lock` in that function. While testing I fixed a minor issues where there was an unnecessary extra call to `getRecord` (in `web/src/views/SubmissionPortal/Components/SubmissionList.vue`).

Tangentially related, there was another bug where a link presented in the "this submission is locked" case was forcing a full page refresh instead of client-side navigation (using `<a>` instead of `<router-link>`). Fixing that bug revealed that the `isEditingSubmission` state in `StepperView` was not correctly [reacting](https://v3.router.vuejs.org/guide/essentials/dynamic-matching.html#reacting-to-params-changes) to route changes. This is fixed by using the existing `id` prop on the component instead of getting it from the route.

### DIFF
--- a/nmdc_server/api.py
+++ b/nmdc_server/api.py
@@ -649,10 +649,10 @@ async def get_submission(
     db: Session = Depends(get_db),
     user: models.User = Depends(get_current_user),
 ):
-    submission = db.query(SubmissionMetadata).get(id)
+    submission: Optional[models.SubmissionMetadata] = db.query(SubmissionMetadata).get(id)
     if submission is None:
         raise HTTPException(status_code=404, detail="Submission not found")
-    _ = crud.try_get_submission_lock(db, submission.id, user.id)
+
     if user.is_admin or crud.can_read_submission(db, id, user.orcid):
         permission_level = None
         if user.is_admin or user.orcid in submission.owners:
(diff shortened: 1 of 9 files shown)


nmdc_server/api.py
```diff
@@ -862,22 +865,63 @@ async def delete_submission(
...
+@router.put("/metadata_submission/{id}/lock")
+async def lock_submission(
+    response: Response,
+    id: str,
+    db: Session = Depends(get_db),
+    user: models.User = Depends(get_current_user),
+) -> schemas.LockOperationResult:
+    submission: Optional[SubmissionMetadata] = db.query(SubmissionMetadata).get(id)
+    if not submission:
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail="Submission not found")
```
naglepuff, line 877: Should we prevent users who don't have any access (i.e. `crud.can_read_submission(...)`) from being able to obtain a lock on a submission?

naglepuff: Maybe this should be done in `crud.try_get_submission_lock` itself now that I'm thinking about it. In either case, this wasn't being done before so if you want to defer it/open up an issue to track this separately that's fine by me.

pkalita-lbl: Let me take a look at how much work / how much impact it would be to do that now.

pkalita-lbl: I added a role check to `try_get_submission_lock`. Seems like a good safeguard to have in place.

naglepuff (approved):
LGTM! Thanks for addressing the issue with `try_get_submission_lock`

PR #1308, opened 2024-07-22
Pass correct string to showColumnsBySectionTitle when showing single section

### LINKED ISSUE #1306: BUG: column filter not working
The "show hide" tool isn't working. When I select a specific section to show only/filter it just hides all the columns.

Screenshot 2024-07-19 at 4 38 56 PM
Screenshot 2024-07-19 at 4 39 04 PM

pkalita-lbl:
Fixes #1306 

Despite its name, DataHarmonizer's `showColumnsBySectionTitle` method actually expects a section _name_, and we were passing it a section _title_. 

The map that was previously being used to drive the options (`schemaSections`) only recorded the section titles. That map has been renamed to `schemaSectionColumns` since it is a map from section title to and array of column information. Otherwise it is unchanged and still used in computing `fields`. 

To actually fix the bug, these changes add a map (`schemaSectionNames`) from section titles (for display) and section names (to pass to `showColumnsBySectionTitle`).

### DIFF
--- a/web/src/views/SubmissionPortal/HarmonizerView.vue
+++ b/web/src/views/SubmissionPortal/HarmonizerView.vue
@@ -250,7 +250,7 @@ export default defineComponent({
       return allTabsValid && isOwner();
     });
 
-    const fields = computed(() => flattenDeep(Object.entries(harmonizerApi.schemaSections.value)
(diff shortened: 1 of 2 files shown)


naglepuff (approved):
LGTM

PR #1309, opened 2024-07-23
Remove unnecesary container

marySalvi:
Removes unnecessary container which created unwanted padding.


naglepuff (approved):
LGTM! I don't see that additional padding any more.

PR #1310, opened 2024-07-23
Fix DataHarmonizer help sidebar positioning

### LINKED ISSUE #1304: BUG: info box expansion is moving to far right
In the NMDC sandbox & in nmdc submission portal production the column help show hide is broken

If you hide the column help pop out, when you re-expand it you have to scroll to the FAR right to see it
Screenshot 2024-07-19 at 4 38 04 PM
Screenshot 2024-07-19 at 4 38 09 PM

pkalita-lbl:
Fixes #1304 

### Description 

I think I got a little overzealous with using flexbox layouts in https://github.com/microbiomedata/nmdc-server/pull/1267, which resulted in this bug being introduced. 

These changes fix the issue by having the help sidebar container be absolutely positioned within its parent element. When toggling the sidebar open/closed a corresponding amount of padding is added to/removed from its sibling element which holds the DataHarmonizer instance. This is so that the sidebar doesn't cover the rightmost columns.

This approach is conceptually similar to the way it worked before https://github.com/microbiomedata/nmdc-server/pull/1267 except it uses absolute positioning instead of `float`. I find `float` has more unpredictable side effects, so I tend to avoid it.

### Testing

Ensure that the help sidebar can be opened and closed while scrolled to any horizontal position in the DataHarmonizer grid. While scrolled all the way to the right, ensure that the final columns can always be accessed.

### DIFF
--- a/web/src/views/SubmissionPortal/HarmonizerView.vue
+++ b/web/src/views/SubmissionPortal/HarmonizerView.vue
@@ -62,6 +62,8 @@ const ColorKey = {
   },
 };
 
+const HELP_SIDEBAR_WIDTH = '300px';
+
 const EXPORT_FILENAME = 'nmdc_sample_export.xlsx';
 
 const SAMP_NAME = 'samp_name';
@@ -466,6 +468,7 @@ export default defineComponent({
 
     return {
       APP_HEADER_HEIGHT,
+      HELP_SIDEBAR_WIDTH,
       ColorKey,
       HARMONIZER_TEMPLATES,
       columnVisibility,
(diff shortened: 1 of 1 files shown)


naglepuff (approved):
I'm going to approve this although I do have one observation.

I can't tell if this is a regression (from before the bug being fixed was introduced), but if I am scrolled all the way to the right so that the last column is visible, then opening the sidebar covers up the last column or so. I scroll to the right again, so *all columns are still accessible*, but there is a weird edge case where a user is scrolled allllll the way over and opens the help sidebar.

I'm fine opening a separate issue to track that if it adds significant complexity to the changes here.

pkalita-lbl:
Yeah I think that's basically the expected behavior and consistent with what it was before https://github.com/microbiomedata/nmdc-server/pull/1267). It's kind of analogous to if you were to be scrolled to the bottom of a web page and you made your browser window shorter by using the resize handle on the bottom of the window.

PR #1311, opened 2024-07-23
Derive Excel worksheet names from schema annotations

pkalita-lbl:
Fixes https://github.com/microbiomedata/submission-schema/issues/213

### Details

Previously we used the "display name" for each template as the worksheet name when importing from/exporting to Excel files. The problem is that some of those names are > 31 characters, making them illegal values for worksheet names.

In https://github.com/microbiomedata/submission-schema/pull/215 I added annotations to the relevant classes in the submission schema to indicate what the corresponding Excel worksheet name should be. Having this in the schema instead of in the `nmdc-server` code is part of a broader effort to have the schema be the single source of truth (see also https://github.com/microbiomedata/submission-schema/issues/186)

In these changes I am:

* Upgrading to the latest schema versions
* Using the new annotation to determine the Excel worksheet names (or continuing to use the "display name" as a fallback, but that shouldn't happen)
* Presenting a new alert to the user in the event that they attempt to import an Excel file where the worksheet names do not map to a schema class.
* Removing generation of static JSON files for individual schema classes. We didn't end up using those files in the submission portal or the Field Notes app.

### Testing

* Create a new submission with the "miscellaneous natural or artificial environment" template.
* Click the Download XLSX button and verify that an Excel file is successfully downloaded.
* Open the Excel file and verify that the worksheet is named "misc natural or artificial".
* Add a value to the first column of the Excel file and save it.
* Click the Import XLSX button and choose your updated Excel file; verify that the import succeeds and you see the value you entered in the first column.


pkalita-lbl:
Marking this as draft because there is a new version of the schema incoming. Will update this PR and un-draft shortly.

pkalita-lbl:
@naglepuff this PR now has upgrades to the latest schema versions

naglepuff (approved):
This looks good! Tested out the workflow outlined in the PR description and can confirm that it works as expected.

PR #1313, opened 2024-07-24
Update ingest and data model to account for Metatranscriptomics

### LINKED ISSUE #1307: Test ingest for new WorkflowExecutionActivity subclasses - metatranscriptomes
Test that ingest code picks up new WorkflowExecutionActivity subclasses.  Example data that is in mongo dev has also been committed to nmdc-schema here 
https://github.com/microbiomedata/nmdc-schema/blob/768-generate-example-data-to-test-that-nmdc-server-code-can-ingest-new-metatranscriptome-workflow-execution-sub-classes/src/data/valid/Database-metatranscriptome_WorkflowExecution_DataObject.yaml

success criteria: 
1) a search for nmdc:bsm-11-cnc9ww90 has a black metatranscriptome box, when expanded there are results for the following:
nmdc:ReadQcAnalysisActivity
nmdc:MetatranscriptomeAssembly
MetatranscriptomeAnnotationActivity
MetatranscriptomeExpressionAnalysis
2) a KEGG search for `KEGG.ORTHOLOGY:K00005` returns biosample nmdc:bsm-11-cnc9ww90

We'd like this tested for the July 29th release.

naglepuff:
## Note:
This change will likely break ingest when attempting to ingest from databases that are not on NMDC schema version 10.7.0. It attempts to pull from the following collections, which did not exist until that version:
`metatranscriptome_annotation_set`
`metatranscriptome_assembly_set`
`metatrascriptome_expression_analysis_set`
Please ensure that the source mongo contains these collections before running ingest.

This essentially duplicates existing functionality for analogous Metagenome workflows. That is, MetagenomeAnnotation and MetagenomeAssembly now have Metatranscriptome counterparts. Also includes a migration to create the new tables. Additionally, gene function search to the metatranscriptome annotation workflows has been added (which again, is identical in functionality to the existing metaproteomic analysis gene function search).

This is in support of some changes to the NMDC schema that fully realize these classes as collections.

Fix #1307

Depends on #1311 (new schema version)

Now that Metatranscriptome workflows are more fleshed out in the schema and mongo database, we need to update our ingest process and models to capture those changes. This set of changes accomplishes the following:

- New models: `MetatranscriptomeAssembly` and `MetatranscriptomeAnnotation`. The existing `Metatranscriptome` model now specifically represents `MetatranscriptomeExpressionAnalysis` records
- New Tables: An `alembic` migration was generated to update the Postgres database(s) to include these new models. Association tables to `data_object` have also been added
- Completion of code that references workflow models: there were several places in the code that make reference to or categorize our models. The new tables have been added where appropriate. For example, new `pydantic` models have been created for the new workflows
- Updates to ingest: we need to pull from new collections, so we had to update the existing pipeline ingest code. This includes an simple pipeline ingest for metatranscriptome assembly, and a more complex one (based on prior art) for metatranscriptome annotation, since these have gene functions.
- More new tables: In order to support searching for gene functions over Metatranscriptome Annotation workflows, a new table had to be created, along with an associated migration to store the relationship between the workflows and Kegg terms.
- New filter: `MetaTGeneFunctionFilter` has been added as a class. Future work could abstract common functionality between this an existing gene function filters into a common parent class.


marySalvi (approved):
LGTM!

PR #1322, opened 2024-08-01
884 sortable submissions list

### LINKED ISSUE #884: Make submission portal submissions list sortable
Can we add (back) the ability to sort submissions

Screen Shot 2023-03-06 at 8 01 37 PM

marySalvi:
Closes #884 
Adds sorting functionality to submission list:
- Updates query to sort based on column and order selected
- Creates study_name and templates columns on submission_metadata model
- Includes migration script to update existing submissions


nmdc_server/migrations/versions/0ff690fb929d_add_study_name_and_template_columns_to_.py
naglepuff, line 1: I think it would be helpful to rebase this branch on top of main (or merge, no preference), and update this revision to point to the new head (see #1313, which introduced 2 migrations).

marySalvi: Definitely! Sorry, I meant to change this back to draft when I ran into the `jwt-encode` issue during the rebase. As we discussed that issue is resolved and I will rebase and update this now.

marySalvi: should be set now

nmdc_server/schemas_submission.py
```diff
@@ -110,6 +110,8 @@ class SubmissionMetadataSchema(SubmissionMetadataSchemaCreate):
     created: datetime
     status: str
     author: schemas.User
+    templates: List[str]
+    study_name: str
```
naglepuff, line 114: Should these be optional? I checked out your branch locally and when loading the submission list I get:

```
backend-1  | pydantic.error_wrappers.ValidationError: 1 validation error for MetadataSubmissionResponse
backend-1  | response -> results -> 5 -> study_name
```

There are ~50 submissions in dev that have no `studyName` (I haven't checked prod). I think this can happen if the submission never makes it past the study form.

naglepuff (changes requested):

nmdc_server/schemas_submission.py
marySalvi: fixed in f37cf212e3faa6934db6ac8e9c80d78dd8bb98da

naglepuff (approved):
Looks good to me! Tested this locally with a somewhat recent backup of production.

PR #1341, opened 2024-08-13
Get ORCID Base URL from environment variable (to enable usage of ORCID sandbox)

eecavanna:
In this branch, I made it so the backend would get the ORCID base URL from an environment variable. This will allow developers to configure the backend to use the sandbox ORCID environment versus the production ORCID environment.

This [comment](https://github.com/microbiomedata/infra-admin/issues/96#issuecomment-2240260134) is what prompted me to implement this change.

### DIFF
--- a/tests/test_auth.py
+++ b/tests/test_auth.py
@@ -14,7 +14,7 @@ def test_login(client: TestClient):
     )
 
     assert resp.status_code == 302
-    assert resp.next.url.startswith(f"{settings.orcid_base_url}/oauth/authorize"),  # type: ignore
+    assert resp.next.url.startswith(f"{settings.orcid_base_url}/oauth/authorize")  # type: ignore
 
 
 def test_current_user(client: TestClient, logged_in_user):
--- a/tests/test_auth.py
+++ b/tests/test_auth.py
@@ -14,7 +14,8 @@ def test_login(client: TestClient):
     )
 
     assert resp.status_code == 302
-    assert resp.next.url.startswith("https://orcid.org/oauth/authorize")  # type: ignore
+    assert resp.next.url.startswith("https://orcid.org/oauth/authorize") or \
+           resp.next.url.startswith("https://sandbox.orcid.org/oauth/authorize")   # type: ignore
 
 
 def test_current_user(client: TestClient, logged_in_user):
--- a/tests/test_auth.py
+++ b/tests/test_auth.py
@@ -14,10 +14,12 @@ def test_login(client: TestClient):
     )
 
     assert resp.status_code == 302
-    assert any((
-        resp.next.url.startswith("https://orcid.org/oauth/authorize"),  # type: ignore
-        resp.next.url.startswith("https://sandbox.orcid.org/oauth/authorize"),  # type: ignore
-    ))
+    assert any(
+        (
+            resp.next.url.startswith("https://orcid.org/oauth/authorize"),  # type: ignore
+            resp.next.url.startswith("https://sandbox.orcid.org/oauth/authorize"),  # type: ignore
+        )
+    )
 
 
 def test_current_user(client: TestClient, logged_in_user):
--- a/tests/test_auth.py
+++ b/tests/test_auth.py
@@ -14,8 +14,10 @@ def test_login(client: TestClient):
     )
 
     assert resp.status_code == 302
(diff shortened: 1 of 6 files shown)


eecavanna:
```py
tox: typecheck
  typecheck: commands[0]> mypy nmdc_server tests
  nmdc_server/config.py:4: error: Module "pydantic" has no attribute "computed_field"
  Found 1 error in 1 file (checked 83 source files)
  typecheck: exit 1 (24.[47](https://github.com/microbiomedata/nmdc-server/actions/runs/10377246515/job/28731021754?pr=1341#step:8:50) seconds) /home/runner/work/nmdc-server/nmdc-server> mypy nmdc_server tests pid=2920
```

Looks like the `computed_field` symbol is not available from the `pydantic` package we're using.

pkalita-lbl (approved):
Changes look good to me.

I made an issue about migrating to Pydantic V2: https://github.com/microbiomedata/nmdc-server/issues/1347. It doesn't seem like it would be very hard to do, unlike, ahem, other [certain upgrades](https://github.com/microbiomedata/nmdc-server/blob/8715c7a1ea71b433980edbd0e274c38ddee756a0/web/package.json#L41) we're a bit behind on.

naglepuff (approved):
Is there a reason not to merge this into `main`?

If we want to start using the ORCID sandbox for the Berkeley environment right away we could always open another PR to merge the `main` branch into the Berkeley base branch.

Anyways, I had one small suggestion which I'll leave up to you to implement or not.

tests/test_auth.py
```diff
@@ -14,7 +14,12 @@ def test_login(client: TestClient):
     )
 
     assert resp.status_code == 302
-    assert resp.next.url.startswith("https://orcid.org/oauth/authorize")  # type: ignore
+    assert any(
+        (
+            resp.next.url.startswith("https://orcid.org/oauth/authorize"),  # type: ignore
+            resp.next.url.startswith("https://sandbox.orcid.org/oauth/authorize"),  # type: ignore
+        )
+    )
```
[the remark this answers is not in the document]
eecavanna: Thanks! I like this. I don't want to lose the suffix (i.e. trailing path) here, so I'll commit the suggested change and then make an additional commit, restoring the suffix.

tests/test_auth.py
```diff
@@ -14,7 +14,7 @@ def test_login(client: TestClient):
     )
 
     assert resp.status_code == 302
-    assert resp.next.url.startswith("https://orcid.org/oauth/authorize")  # type: ignore
+    assert resp.next.url.startswith(settings.orcid_base_url),  # type: ignore
```
eecavanna, line 17:
```suggestion
    assert resp.next.url.startswith(f"{settings.orcid_base_url}/oauth/authorize"),  # type: ignore
```

tests/test_auth.py
```diff
@@ -14,7 +14,7 @@ def test_login(client: TestClient):
     )
 
     assert resp.status_code == 302
-    assert resp.next.url.startswith("https://orcid.org/oauth/authorize")  # type: ignore
+    assert resp.next.url.startswith(f"{settings.orcid_base_url}/oauth/authorize"),  # type: ignore
```
eecavanna, line 17:
```suggestion
    assert resp.next.url.startswith(f"{settings.orcid_base_url}/oauth/authorize")  # type: ignore
```

eecavanna:
Hi @naglepuff, in response to:

> Is there a reason not to merge this into `main`?

Not that I know of. I'll switch the base branch to be `main` once the GHA workflows finish (in a few minutes).

eecavanna:
When I temporarily changed the base branch to `main`, the diff showed that the Berkeley-specific changes on this branch (which was based upon the `berkeley-schema-migration` branch) would also be merged into `main` (makes sense). So, changing the base branch of this PR to `main` is not the path I want to go down, here. Instead, I'll reimplement these changes in a new branch that is based upon the `main` branch, and then I'll open up a PR to merge _that_ into `main` (and then close this PR).

eecavanna:
Update: Equivalent changes have been merged into `main` via https://github.com/microbiomedata/nmdc-server/pull/1348. I will close this PR now and pursue merging `main` into `berkeley-schema-migration` in order to achieve the same end result.

PR #1348, opened 2024-08-19
Make ORCID Base URL configurable via environment variable

eecavanna:
In this branch, I made it so the backend would get the ORCID Base URL from an environment variable. This will allow developers to configure the backend to use the sandbox ORCID environment vs. the production ORCID environment.

This [comment](https://github.com/microbiomedata/infra-admin/issues/96#issuecomment-2240260134) is what prompted me to implement this change.

---

Note: The changes in this PR were originally made with respect to the `berkeley-schema-migration` branch. Based upon some feedback in #1341, I have created this new branch in which the same changes have been made with respect to the `main` branch. Once this PR gets merged into `main`, I will close PR #1341 (and—instead—to a `main`-to-`berkeley-schema-migration` merge in order to "get these same changes" to exist in the `berkeley-schema-migration` branch).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -7,6 +7,11 @@
 # ORCID OAuth Setup
 # NMDC_ORCID_CLIENT_ID=changeme
 # NMDC_ORCID_CLIENT_SECRET=changeme
+#
+# Base URL (without a trailing slash) at which the application can access an instance of ORCID.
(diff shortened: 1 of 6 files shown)


eecavanna:
I will merge this branch in _without review_ because the changes are the same (aside from surrounding a value with double quotes in a `.env` file) as those that were already reviewed in #1341.

PR #1349, opened 2024-08-19
Merge `main` into `berkeley-schema-migration`

eecavanna:
Merging this PR will merge the `main` branch into the `berkeley-schema-migration` branch.

My motivation for doing this is to make the ORCID Base URL configurable via an environment variable, which is a capability introduced into `main` via https://github.com/microbiomedata/nmdc-server/pull/1348.

The `main` branch happens to contain some other changes, with which I'm not familiar. Looks to me like at least some of them are related to being able to sort a list of submissions.


web/src/util.js
```diff
@@ -188,3 +188,9 @@ export function formatBiosampleDepth(depthAnnotation, depth) {
   }
   return formattedStr;
 }
+
+/**
+ * Base URL (without a trailing slash) at which the user can access
+ * the same ORCID environment being used by the backend API.
+ */
+export const ORCID_BASE_URL = process.env.NMDC_ORCID_BASE_URL || "https://orcid.org";
```
eecavanna, line 196:
```suggestion
export const ORCID_BASE_URL = process.env.NMDC_ORCID_BASE_URL || 'https://orcid.org';
```

eecavanna: GitHub won't allow me to commit this suggestion (maybe because there's no branch dedicated to this PR). So, I'll fix this in `main`.

naglepuff (approved):
LGTM!

eecavanna:
Thanks, @naglepuff! I will merge this into `berkeley-schema-migration` now.

I see—in [.github/workflows/berkeley-image.yml](https://github.com/microbiomedata/nmdc-server/blob/berkeley-schema-migration/.github/workflows/berkeley-image.yml)—that that will trigger a new build and publish of the "berkeley" container images. I don't know whether anything will automatically "redeploy" the Spin workloads after that. I'll do that manually if I find out that nothing has done it automatically after a few minutes.

PR #1353, opened 2024-08-21
Support pathway search with multiple prefixes

### LINKED ISSUE #1342: Data Portal KEGG pathway search issue with pathway has ko prefix
when use the functional search with ko00010 it does populate the term name. 

see https://www.genome.jp/kegg/kegg1b.html for expected prefixes.

This term should populate with a name of `Glycolysis / Gluconeogenesis`

https://www.genome.jp/dbget-bin/www_bget?pathway+ko00010

naglepuff:
Fix #1342

### Changes
This updates the KEGG facet to allow users to search for pathways terms using prefixes other than `map`. Supported prefixes for pathway searches now include `ko`, `ec`, and `rn` in addition to `map`. 

We store pathway terms in the `ko_term_text` table in postgres, all with the `map` prefix. In order to support searching on these terms we do some string substitution so that users can use whichever prefix they'd like. 

This affects free-text search for the KEGG facet on the frontend. If a user types in a term that begins with one of the prefixes associated with a pathway search, the associated backend code (`crud.py::kegg_text_search`) will interpret that as a request to search for pathways, making sure the results keep the the prefix that the user wanted. For example, a user search for `ec00010` in the Data Portal will be translated into a search for `map00010` in our database, and the results will be translated back to use the `ec` prefix.

This also affects some KEGG-related functions in `encoding.ts`. In order to support transforming search terms into proper strings used by the search engine and proper links to external sites, the `short` and `long` representations are now functions that adapt to the input.

A similar change has been implemented during search itself, where the check for whether or not a given search term is a pathway has been expanded.

The help text on the KEGG filter flyout has been updated to mention the new supported formats.

### Testing
Testing requires working with a database that has the KEGG terms and annotations ingested. I would recommend using the `nmdc-server` CLI to obtain a backup of the production database. From there, test the KEGG Filter by searching for pathways with various different prefixes. Make sure they return the same results. Additionally, make sure that the free text search results are consistent with the prefix you are using for search, and that the clickable links in the flyout when you have terms selected use the same prefix you searched with.

### DIFF
--- a/nmdc_server/crud.py
+++ b/nmdc_server/crud.py
@@ -226,7 +226,7 @@ def list_omics_processing_data_objects(db: Session, id: str) -> Query:
 
 
 # KEGG
-def has_pathway_prefix(term) -> Optional[str]:
+def get_pathway_prefix(term) -> Optional[str]:
     pathway_prefixes = set(["map", "ko", "ec", "rn", "org"])
     pathway_re = f"^({'|'.join(re.escape(p) for p in pathway_prefixes)})"
     match = re.match(pathway_re, term)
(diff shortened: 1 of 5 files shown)


naglepuff:
b38715d changes the URL for modules. Previously they were not resolving correctly.

marySalvi (approved):
LGTM!

PR #1358, opened 2024-08-22
Add schema version identifiers to /api/version response

pkalita-lbl:
This is part of https://github.com/microbiomedata/nmdc-field-notes/issues/156

These changes add the `nmdc-schema` and `nmdc-submission-schema` version identifiers to the response of the existing `/api/version` endpoint.

⚠️ Quasi-breaking change ⚠️ 

The response used to contain the key `nmdc-server`. I changed it to `nmdc_server` along with adding new keys that also use underscores instead of dashes. This makes it easier to deal with in both Python and JavaScript/TypeScript code. I don't _think_ there are any existing clients of this endpoint, but if anyone is uncomfortable with that change let me know.

### DIFF
--- a/nmdc_server/api.py
+++ b/nmdc_server/api.py
@@ -10,7 +10,7 @@ from fastapi.responses import JSONResponse
 from sqlalchemy.orm import Session
 from starlette.responses import StreamingResponse
 
-from nmdc_server import __version__, crud, jobs, models, query, schemas, schemas_submission
+from nmdc_server import crud, jobs, models, query, schemas, schemas_submission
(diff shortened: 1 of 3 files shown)


eecavanna (approved):
Looks good to me! Thanks for adding these values to the response.

naglepuff (approved):
LGTM!

eecavanna:
FYI: The `/version` endpoint of the Runtime uses hyphens in its response property names.

image

pkalita-lbl:
Thanks for pointing out the similar endpoint in `nmdc-runtime`. I might make some inquiries about how it's being used and whether it would be disruptive to change that one, too.

PR #1366, opened 2024-08-25
Add click functionality to numbers

### LINKED ISSUE #1044: improved interface for clicking on the upset plot
During the workshop today we discovered it is difficult to select data that has metagenomes AND metaproteomics . The bar chart in the upset plot is too small (n=17) to click on and the faceted search is an OR selection. Can we make the bar + number for each row in the  upset chart clickable

cc @emileyfadrosh

marySalvi:
closes #1044 
The numbers for the upset bar chart are now clickable and should return the same results as the bar improving usability for the smaller values.

### DIFF
--- a/web/src/components/Presentation/UpSet.vue
+++ b/web/src/components/Presentation/UpSet.vue
@@ -77,6 +77,20 @@ export default defineComponent({
 
       select(el).selectAll('g').remove();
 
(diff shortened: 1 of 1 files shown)


naglepuff (approved):
LGTM!

Tested it out, numbers being clickable makes a big difference.

PR #1372, opened 2024-08-29
1367 multiple env select

### LINKED ISSUE #1367: Add ability to select multiple sample type environment
Currently, NMDC submission portal has a radio button to select only 1 type of sample (soil, water, sediment, etc)

We'd like this to be checkboxes and all for 1 study to have multiple sample types in 1 submission.

- IF submitting for JGI and EMSL, all the sample types will need to populate the JGI and EMSL tabs the same way the single sample does
- Need to make sure the sample type is populated for each sample when the ingest to mongo happens
- Can we update Environmental Packages to Environmental Extensions to accurately match GSC

marySalvi:
**_This is ready for a review but not ready to be merged yet._**

closes #1367
Allows the ability to select multiple enviroment extensions f/k/a environment packages.

Testing:
- Create new study.
- 'Soil' should be selected when on the environment extension step
-  Should be able to select more extensions
-  When navigating to the DH the corresponding template tabs should be available.

TODO:
Before this can be merged the corresponding runtime code needs to be updated.


pkalita-lbl (changes requested):
I tested this branch a bit locally and found a few issues:

1. I created a new submission. Once I got to the `/templates` page, I saw that "soil" was pre-checked, as expected. However I was able to uncheck this box and click the "go to next step" button. **I expected that button to be disabled if no checkboxes are checked.**
2.
    1. I created a new submission and selected the "soil" and "water" extensions. On the `/samples` page I see a "soil" tab and a "water" tab as expected. 
    2. In the "sample name" column I entered 1, 2, and 3 in the first three rows, respectively, of the "soil" tab. 
    3. I clicked on the "water" tab and entered 4, 5, and 6 into the "sample name" column of the first three rows, respectively. 
    4. I clicked on the "soil" tab again and saw the first three rows filled out as expected. 
    5. I clicked on the "water" tab again and saw the first three rows were blank. **I expected to see 4, 5, and 6 in the first column.**
3. 
    1. I created a new submission and on the `/context` page I selected "No" for "Have data already been generated for your study?" and "JGI" for "Are you submitting metadata for samples that will be sent to a DOE user facility?" and "CSP" for "What kind of project have you been awarded?". 
    2. On the `/multiomics` page I selected "Metagenome" in the JGI section and entered "123456" as a proposal ID. 
    3. On the `/templates` page I selected "soil" and "water". 
    4. On the `/samples` page I entered 1, 2, and 3 in the first three rows, respectively, of the "sample name" column and selected "metagenomics" for the first three rows in the "analysis/data type" column.
    5. I clicked on the "water" tan and entered 4, 5, and 6 into the first three rows, respectively, of the "sample name" column. I selected "metagenomics" for the first three rows in the "analysis/data type" column.
    6. I clicked on the "JGI MG" tab and saw no rows. **I _think_ would have expected to see six rows with 1 through 6 in the first column.** I say "think" because I'm not sure the behavior in this case has been clearly defined.

This is mostly a note to myself, but this change will need corresponding changes in the Field Notes app. This will be a good test case for releasing coordinating changes to the backend and the app.

marySalvi:
> 3. 1. I created a new submission and on the `/context` page I selected "No" for "Have data already been generated for your study?" and "JGI" for "Are you submitting metadata for samples that will be sent to a DOE user facility?" and "CSP" for "What kind of project have you been awarded?".
>    2. On the `/multiomics` page I selected "Metagenome" in the JGI section and entered "123456" as a proposal ID.
>    3. On the `/templates` page I selected "soil" and "water".
>    4. On the `/samples` page I entered 1, 2, and 3 in the first three rows, respectively, of the "sample name" column and selected "metagenomics" for the first three rows in the "analysis/data type" column.
>    5. I clicked on the "water" tab and entered 4, 5, and 6 into the first three rows, respectively, of the "sample name" column. I selected "metagenomics" for the first three rows in the "analysis/data type" column.
>    6. I clicked on the "JGI MG" tab and saw no rows. **I _think_ would have expected to see six rows with 1 through 6 in the first column.** I say "think" because I'm not sure the behavior in this case has been clearly defined.
> 

@mslarae13 @aclum Thoughts on correct behavior in this instance? Are there others I should loop in?

aclum:
you are correct, there should be 6 rows with 1-6

mslarae13:
just looking at this but yes. the current behavior with 1 sample type for JGI or EMSL metadata & sample submission (data not generated yet) is to have samples of specific types appear in the respective user facilities sheet. 
so if all 6 samples (3 water, 3 soil) have metaGs, they should appear in the JGI tab.

marySalvi:
@pkalita-lbl I think I have addressed all of the issues/bugs you found. Do you mind checking again?

pkalita-lbl (approved):
I believe everything is working as expected now! I made one minor comment, but I'll mark this as approved so that we can get this on dev soon and other folks can test it.

There's a good chance I might request one or two minor backend tweaks to help ease the transition from the Field Notes perspective. But I think it'll be easier if I just open a separate PR for that later.

nmdc_server/migrations/versions/afa1ff687968_convert_packagename_to_an_array.py
```diff
@@ -0,0 +1,57 @@
...
+    )
+
+    connection = op.get_bind()
+    submissions = connection.execute(
+        sa.select([submission_metadata.c.id, submission_metadata.c.metadata_submission])
+    )
+
+    for submission in submissions:
+        metadata_submission = submission.metadata_submission
+        package_name = metadata_submission.get("packageName")
```
naglepuff, line 40: Looks like we do add a default value to the package name from the frontend during submission creation. Also, there are no `null`s in any submission on dev, so this seems safe.

web/src/views/SubmissionPortal/HarmonizerView.vue
```diff
@@ -132,8 +132,10 @@ export default defineComponent({
       // WARNING: It's important to do the column settings update /before/ data. Otherwise,
       // columns will not be rendered with the correct width.
       harmonizerApi.setColumnsReadOnly(ALWAYS_READ_ONLY_COLUMNS);
-      // if we're not on the first tab, the common columns should be read-only
-      if (activeTemplateKey.value !== templateList.value[0]) {
+
+      // If the environment tab selected is a misin it should be readonly
```
naglepuff, line 136:
```suggestion
      // If the environment tab selected is a mixin it should be readonly
```

naglepuff (approved):
LGTM! Some incredibly nitpicky comments, which you can disregard if you'd like

mslarae13:
@pkalita-lbl status on getting this for the app so we can merge?

aclum:
Checking in on this since the last comment was that this would be merged shortly and that was 2 weeks ago. FYI there is currently a conflict in nmdc_server/schemas_submission.py

nmdc_server/migrations/versions/afa1ff687968_convert_packagename_to_an_array.py
```diff
@@ -0,0 +1,57 @@
...
+            .where(submission_metadata.c.id == submission.id)
+            .values(metadata_submission=metadata_submission)
+        )
+        connection.execute(update_stmt)
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    pass
```
naglepuff, line 56: Do you think there'd be any value gained from writing a `downgrade` operation here?

naglepuff (commented):
Tested it out, looks good, I'm just wondering if there's any value in adding a downgrade operation to the new migration.

nmdc_server/migrations/versions/afa1ff687968_convert_packagename_to_an_array.py
marySalvi: added in c91fd55b654c5b64166b4b2b37633d4d7184025b

naglepuff (approved):
LGTM

PR #1378, opened 2024-09-09
Assign study_name on patch and migrate existing names

### LINKED ISSUE #1377: BUG: Table of submissions doesn't show study name
I'm not sure where this bug lies... but

I made a study
Gave it a title
The title isn't appearing in the table?

Screenshot 2024-09-03 at 11 58 31 AM
Screenshot 2024-09-03 at 11 58 37 AM

marySalvi:
closes #1377


naglepuff (approved):
This looks good to me.

I think this is a good surgical fix for now, although we might want to open an issue to make a decision for the now redundant `submission.study_name` and `submission["studyForm"]["studyName"]`. 

@pkalita-lbl for now we're going to take care of syncing these two fields in the API, so I imagine there won't be a need to update the field notes app, although I do think in the future we want to lift `studyName` out of the study page form.

To test, I first started with v0.24.0 and make a submission. I gave it a name and reproduced the bug where the name did not appear in the submission list. I checked this branch out and ran the migration, which populated the study name as expected, so it showed up in the submission list.

I created a new submission and changed the names of submissions, and the list reflected those changes as expected.

marySalvi:
> I think this is a good surgical fix for now, although we might want to open an issue to make a decision for the now redundant submission.study_name and submission["studyForm"]["studyName"].

Agreed, this is sufficient for a hotfix but should definitely be resolved in a better manner long term

PR #1376, opened 2024-09-04
Ingest PFAM entries and clans

naglepuff:
Depends on #1359 (due to migrations)

Fix #1337

PR #1395, opened 2024-09-18
Ensure template props are synced on PATCH

### LINKED ISSUE #1392: New submission does not sort correctly by template
I created a new submission on dev with the name "air air air air air air air" and the template -- you guessed it -- "air". When I sort the table on `/submission/home` by "Template" I expected to see my new submission sorted with the existing airs:

image

Instead I found it several pages later hanging out with some soils:

image

I believe this is a similar bug to https://github.com/microbiomedata/nmdc-server/issues/1377 where a new top-level field was added to submission records that isn't kept in sync with the underlying source of truth. I don't think the symptom here is severe enough to warrant a hotfix like that previous issue.

naglepuff:
Fix #1392 

### Changes

On `PATCH` requests to save submissions, the property `submission.metadata_submission["templates"]` is copied up to `submission.templates`. 

A migration has been included to fix existing submissions where these may have gotten out of sync.


marySalvi (approved):
LGTM 

Please note that this and #1376 both have the same downgrade version. One will need to change.

PR #1420, opened 2024-10-15
Update dependencies and enable sentry tracing

### LINKED ISSUE #1264: Enable tracing in sentry

### LINKED ISSUE #1347: Upgrade to Pydantic 2
As pointed out in https://github.com/microbiomedata/nmdc-server/pull/1341 we are currently using Pydantic V1. While V1 versions are still maintained (for now anyway), I don't we're doing anything so fancy with Pydantic that it would make migrating to V2 very difficult. There are a few instance of using `validator` and `root_validator`, but those have fairly direct replacements: https://docs.pydantic.dev/latest/migration/#validator-and-root_validator-are-deprecated. Anyway, seems like we might want to cut over to V2 before it gets harder to do so.

naglepuff:
Fix #1264 
Fix #1347 

### Motivation
As part of the ongoing look at performance of the data portal's API, we decided that setting up [profiling with Sentry](https://docs.sentry.io/product/explore/profiling/) would likely give us valuable insight into where the application spends time in its long-running queries. Part of enabling this for `nmdc-server` required updating FastAPI, which cascaded into needing to update other packages, like Starlette and Pydantic. Our dependencies were cluttered with pinned versions, many of which were not documented well. Updating many dependencies here gives us access to the latest and greatest features of the tools we're using in `nmdc-server`, and can help resolve security warnings.

#### Dependencies
Based on the need to update `fastapi`, `starlette`, and `pydantic`, new version numbers were determined for our dependencies. See `pyproject.toml` for a full list.

#### Pydantic Models
The change from `pydantic` v1 to v2 forced us to [update](https://docs.pydantic.dev/latest/migration/#migration-guide) many of our models, since we were using features of `pydantic` that were deprecated in v2. The `bump-pydantic` [tool](https://github.com/pydantic/bump-pydantic) was used initially, which left several `# TODO[Pydantic]` comments that were addressed by hand.

We are still importing from `pydantic.v1` in some files under `ingest/`. I think its fine to address these at a later time in a follow-up effort.

#### Cleanup
Updating `pydantic` caused some of our tests to fail (`pytest` and `mypy`). Those issues were all addressed.

#### FastAPI Debug Toolbar
The [fastapi-debug-toolbar](https://pypi.org/project/fastapi-debug-toolbar/) has been added for use in development. It is accessible from the FastAPI swagger page (i.e. `127.0.0.1:8000`). This can be used by developers to take a look at endpoint performance.

#### Sentry Tracing
Configuration options have been added to `config.py` that are used when initializing Sentry that can turn on profiling. This will allow those with access to the sentry dashboard access to production information about performance.

### Follow up tasks
Determine a best practice for updating dependencies for `nmdc-server`. This could be monthly with releases, or at a different cadence. We'll still likely be updating the schema versions every month. We'll also need to determine which stuff we need to pin long term, and which things we can let `pip` or `uv` determine the version of. Examples of longer-term pins would be `sqlalchemy`.

Go back to the files in `ingest/` that are still using some Pydantic v1 features and migrate those.


nmdc_server/app.py
```diff
@@ -35,18 +49,16 @@ def create_app(env: typing.Mapping[str, str]) -> FastAPI:
         version=__version__,
         docs_url="/api/docs",
         openapi_url="/api/openapi.json",
+        debug=True,
```
danlamanna, line 52: This should probably be a setting I think?

nmdc_server/app.py
```diff
@@ -20,10 +22,22 @@ def attach_sentry(app: FastAPI):
     sentry_sdk.init(
         dsn=settings.sentry_dsn,
         integrations=[SqlalchemyIntegration()],
+        enable_tracing=settings.sentry_tracing_enabled,
+        traces_sample_rate=settings.sentry_traces_sample_rate,
```
danlamanna, line 26:
```suggestion
        integrations=[
               LoggingIntegration(level=logging.INFO, event_level=logging.WARNING),
               SqlalchemyIntegration(),
        ],
        in_app_include=["nmdc_server"],
        attach_stacktrace=True,
        traces_sample_rate=settings.sentry_traces_sample_rate,
```

This should allow sentry issues to be tracked when warnings get logged as well.

nmdc_server/app.py
naglepuff: Yes, makes sense to me. Fixed in [41f49af](https://github.com/microbiomedata/nmdc-server/pull/1420/commits/41f49af72185d4e41ffec18b525e6dab1a329d9e)

danlamanna (approved):

PR #1421, opened 2024-10-17
Remove Berkeley update banner from data and submission portals

naglepuff:
Also consolidates all of the boolean checks into one place (`AppBanner.vue`).

### DIFF
--- a/web/src/components/AppBanner.vue
+++ b/web/src/components/AppBanner.vue
@@ -1,5 +1,17 @@
+<script lang="ts">
+import { defineComponent, ref } from '@vue/composition-api';
+
+export default defineComponent({
+  setup() {
+    const showAppBanner = ref(true);
+    return { showAppBanner };
+  },
+});
+</script>
+
 <template>
   <v-banner
+    v-if="showAppBanner"
     color="blue"
     dark
     icon="mdi-information"
--- a/web/src/components/AppBanner.vue
+++ b/web/src/components/AppBanner.vue
@@ -3,7 +3,7 @@ import { defineComponent, ref } from '@vue/composition-api';
 
 export default defineComponent({
   setup() {
-    const showAppBanner = ref(true);
+    const showAppBanner = ref(false);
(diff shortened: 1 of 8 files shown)


naglepuff:
I created this branch off of `v1.0.0`, which, at the time of writing, is at the same commit as `main`. Once this is approved, I can initiate a [hotfix](https://github.com/microbiomedata/infra-admin/blob/main/releases/nmdc-server.md#developing-and-releasing-a-hotfix) to get these changes into dev.

eecavanna (approved):

eecavanna:
I have closed PR https://github.com/microbiomedata/nmdc-server/pull/1426, which only hid the banner and did not consolidate the recurring booleans. I want this PR (#1421) to be merged into `main` instead.

eecavanna:
On second thought, I will merge that other PR into `main` first; since it's been reviewed/approved. I still want the consolidation that this one contains, to be applied to `main` eventually.

naglepuff:
@eecavanna Good thought, I agree. I can either rebase this onto main once your PR is merged, or I could open up an issue to track consolidating the boolean. I'm also open to using this PR to add the desired functionality where the banner message is driven by an env var instead of being hard-coded and managed through hotfixes.

There's also the question of keeping the `v-if` statements as-is, in the case that a banner should be applied to some views but not all. I'm not sure of a use case or if the extra maintenance is worth it, I just though I'd mention that since it crossed my mind.

eecavanna:
_(I'm catching up after being OOO)_

The diff still looks good to me and I'm comfortable with this being merged in whenever @naglepuff considers it done. I am OK with the "further" goals being left for a separate PR (e.g. introducing an environment variable or having the client fetch the banner data from the server at run time).

Regarding keeping the `v-if` statements in the individual banners, I'd prefer leaving them out at this point. We can add them in if we find we need them (I don't think the need has arisen in the 2-3 events where we have put up banners so far).

PR #1428, opened 2024-10-21
Fix ingest post dependency upgrade

### LINKED ISSUE #1423: bug: `annotations` column not populated by ingest
This only affects the DEV environment. 

I believe this was introduced by #1420.

This breaks ingest, since the `annotations` column of the `omics_processing` table is used to populate the `multiomics` column of the `biosample` table.

This might be a side effect of the upgrade to Pydantic v2.

naglepuff:
Fix #1423 

Caused by #1420

### Changes
After merging #1420 into dev, ingest stopped working. The initial reason reported by the failed ingest job was related to the use of `pydantic` in `nmdc-server`. After addressing that issue, I noticed there were more issues preventing ingest from running successfully.

###### PydanticUserError: `Study` is not fully defined; you should define `List`, then call `Study.model_rebuild()`.
This was the original issue, which can be seen in the logs for the failed ingest jobs. I followed the instructions given in the message to resolve this.

###### Type mismatch between data and postgres tables
Several columns on `MetagenomeAssembly` and `MetatranscriptomeAssembly` were mistyped as `BigInteger`, where the most recent `nmdc-schema` documentation reports them as having a range of `Float`. The columns have had their types changed and a migration is included in this PR. The analogous pydantic model properties and faker properties have been changed as well. I'm not sure what the relationship between the changes in #1420 and this error being discovered is, but it was not reported before. 

###### Failure to populate `annotations` column
As suggested in the issue, #1423, the json column, `annotations` on a few of our tables was empty. This caused an ingest failure because the annotations are used to populate the `multiomics` column for biosamples. This was fixed here by using Pydantic v2 `model_validator`s instead of v1 `root_validators`. Some feature of Pydantic v1 can still be imported an used, but in this case those weren't working, while the v2 equivalent did.


nmdc_server/fakes.py
```diff
@@ -222,25 +222,25 @@ class Meta:
...
+    scaf_l90: int = Faker("pyfloat")
+    ctg_n90: int = Faker("pyfloat")
+    ctg_l90: int = Faker("pyfloat")
+    scaf_max: int = Faker("pyfloat")
+    ctg_max: int = Faker("pyfloat")
+    scaf_n_gt50k: int = Faker("pyfloat")
+    scaf_l_gt50k: int = Faker("pyfloat")
+    scaf_pct_gt50k: int = Faker("pyfloat")
+    num_input_reads: int = Faker("pyfloat")
+    num_aligned_reads: int = Faker("pyfloat")
```
jeffbaumes, line 243: @naglepuff should these be type `float` instead of `int` now?

naglepuff: Good catch - these should all be correctly typed as of a816215

nmdc_server/ingest/omics_processing.py
```diff
@@ -36,9 +36,9 @@
 class OmicsProcessing(OmicsProcessingCreate):
     _extract_value = validator("*", pre=True, allow_reuse=True)(extract_value)
 
-    @root_validator(pre=True)
+    @model_validator(mode="before")
```
marySalvi, line 39: I think changing the mode to `after` allows us to not have to use the type ignore here.

If I am understanding the docs correctly I do think that is what we want here, to validate the rest of the chain first:
https://docs.pydantic.dev/latest/api/functional_validators/#pydantic.functional_validators.AfterValidator

nmdc_server/ingest/study.py
```diff
@@ -41,9 +42,13 @@ def get_or_create_pi(db: Session, name: str, url: Optional[str], orcid: Optional
...
     _extract_value = validator("*", pre=True, allow_reuse=True)(extract_value)
 
-    @root_validator(pre=True)
+    @model_validator(mode="before")
     def extract_extras(cls, values):
-        return extract_extras(cls, values)
+        return extract_extras(cls, values)  # type: ignore
+
+
+List = list
```
danlamanna, line 50: Out of curiosity, does adding `from typing import List` to the top of the file remove the need for these 2 lines?

nmdc_server/ingest/omics_processing.py
naglepuff: I believe that here we're building a dictionary to put in the `annotations` column, so I would expect we want this to run before other validation, since we're passing in a dict.

nmdc_server/ingest/study.py
naglepuff: Yes, it does

marySalvi (approved):

nmdc_server/ingest/study.py
naglepuff: See 4e2f5af

PR #1442, opened 2024-11-12
767 add email field

marySalvi:
Closes https://github.com/microbiomedata/issues/issues/767
- Adds an e-mail column to users
- Adds and e-mail field to the user detail page, this also allows users to overwrite a currently stored e-mail
- If email for a user is not already defined, it is extracted and stored with orcid login
- Prevents users from making a submission in the submission portal without an e-mail associated with their profile


nmdc_server/auth.py
```diff
@@ -250,7 +251,35 @@ async def orcid_authorize(request: Request, db: Session = Depends(get_db)):
...
+
+    user_email = None
+    # Get the first valid email
+    # TODO - Consider storing all valid emails in the future
+    for email in email_elements:
+        if email.text and "@" in email.text:
+            user_email = email.text
+            break
+
+    user_email = user_email
```
naglepuff, line 278: Can this line be removed?

web/src/views/User/UserDetailPage.vue
```diff
@@ -89,6 +108,44 @@ export default defineComponent({
...
+              <v-text-field
+                v-model="user.email"
+                label="Email"
+                dense
+                filled
+                :rules="requiredRules('E-mail is required',[
+                  v => /.+@.+\..+/.test(v) || 'E-mail must be valid',
+                ])"
+                @blur="updateUser($event.target.value)"
+              >
```
naglepuff, line 125: Do we want to maybe wrap the `updateUser` call with some logic that ensures the email is valid? I seem to be able to save my user with a nonsense email:

image

nmdc_server/auth.py
```diff
@@ -250,7 +251,35 @@ async def orcid_authorize(request: Request, db: Session = Depends(get_db)):
     # TODO: use token_response.access_token to make requests to the ORCID API for more user info?
     # Or possibly store the access token in the database for later use?
     user = User(orcid=token_response["orcid"], name=token_response["name"])
+
+    orcid_email_response = await oauth2_client.orcid.get(
+        "https://pub.orcid.org/v3.0/0009-0004-9185-506X/email", token=token_response
+    )
```
naglepuff, line 257:
```suggestion
    orcid_email_response = await oauth2_client.orcid.get(
        f"https://pub.orcid.org/v3.0/{token_response['orcid']}/email", token=token_response
    )
```

I think this URL should use the current user's ORCID

marySalvi: Ugh, can't believe I forgot to change that.

nmdc_server/auth.py
marySalvi: removed in 3c6ce981d57ae696db0df296997f1ab1e62521ed

nmdc_server/auth.py
marySalvi: removed in 3c6ce981d57ae696db0df296997f1ab1e62521ed

naglepuff (approved):
LGTM!

mslarae13:
Hi @marySalvi 

clarification, should there be a new field in the submission portal? 
Screenshot 2024-11-24 at 1 36 53 PM


Or, is this just captured on the back end?

If it's just on the backend, how does someone add or changes their email? 

So, "Adds and e-mail field to the user detail page, this also allows users to overwrite a currently stored e-mail"

marySalvi:
I did add a field on the frontend but on the user's detail page.
[Screencast from 11-25-2024 12:49:19 PM.webm](https://github.com/user-attachments/assets/409af650-2883-4267-977c-8e78c386b89d)




Additionally when a user attempts to submit through the submission portal, they are blocked if they don't have an e-mail associated with their account. We can add a link to help user's find their detail page if you like.  Note that if there is an e-mail associated with their ORCiD, we save that when they login.

mslarae13:
@ssarrafan can you add this to the infrastructure meeting agenda? I think there was some lost details about what I was trying to accomplish here. I think this works, but we need to still make some changes to the submission portal to satisfy this.

mslarae13:
Decision from 2024-12-05 infra sync. @marySalvi is going to make a change to allow a user to submit even if their email is missing.

We will add a follow up change to make email required and get clear guidance on how a submitter can add their email.

mslarae13:
Looks to be working well!!

PR #1443, opened 2024-11-12
Implement API endpoint that returns TSV report about submissions that are pending review

cristina-stonepedraza:
Add a script to api.py and a query to crud.py to pull information from the submission portal database, and then generate a report of NMDC submissions that have been submitted as a TSV. 

This PR references issue 2047 in nmdc-schema: https://github.com/microbiomedata/nmdc-schema/issues/2047

### DIFF
--- a/nmdc_server/api.py
+++ b/nmdc_server/api.py
@@ -645,9 +645,9 @@ async def get_metadata_submissions_mixs(
     db: Session = Depends(get_db),
 ):
     r"""
-    Generate a TSV-formatted report of biosamples belonging to submissions 
+    Generate a TSV-formatted report of biosamples belonging to submissions
     that have a status of "Submitted- Pending Review".
-    
+
     The report indicates which environmental package/extension, broad scale,
     local scale, and medium are specified for each biosample. The report is
     designed to facilitate the review of submissions by NMDC team members.
--- a/nmdc_server/api.py
+++ b/nmdc_server/api.py
@@ -643,6 +643,7 @@ async def download_zip_file(
 )
 async def get_metadata_submissions_mixs(
     db: Session = Depends(get_db),
+    user: models.User = Depends(get_current_user)
 ):
     r"""
     Generate a TSV-formatted report of biosamples belonging to submissions
@@ -652,6 +653,9 @@ async def get_metadata_submissions_mixs(
     local scale, and medium are specified for each biosample. The report is
     designed to facilitate the review of submissions by NMDC team members.
     """
+    if not user.is_admin: 
+        raise HTTPException(status_code=403, detail="Your account has insufficient privileges.") 
+
     # Get the submissions from the database.
     q = crud.get_query_for_submitted_pending_review_submissions(db)
     submissions = q.all()
--- a/nmdc_server/api.py
+++ b/nmdc_server/api.py
@@ -638,7 +638,7 @@ async def download_zip_file(
 
 
 @router.get(
-    "/metadata_submission/mixs",
+    "/metadata_submission/mixs_report",
     tags=["metadata_submission"],
 )
 async def get_metadata_submissions_mixs(
--- a/nmdc_server/api.py
+++ b/nmdc_server/api.py
@@ -676,7 +676,7 @@ async def get_metadata_submissions_mixs(
         # Get sample names from each sample type
         for sample_type in sample_data:
             samples = sample_data[sample_type] if sample_type in sample_data else []
-        #     # Iterate through each sample and extract the name
+            # Iterate through each sample and extract the name
             for x in samples:
                 # Get the sample name
                 sample_name = x["samp_name"] if "samp_name" in x else ""
--- a/nmdc_server/api.py
+++ b/nmdc_server/api.py
@@ -657,9 +657,9 @@ async def get_metadata_submissions_mixs(
         'Submission ID',
         'Status',
         'Sample Name',
-        'Environmental Package/Extension', 
-        'Environmental Broad Scale', 
-        'Environmental Local Scale', 
+        'Environmental Package/Extension',
+        'Environmental Broad Scale',
+        'Environmental Local Scale',
         'Environmental Medium',
     ]
     data_rows = []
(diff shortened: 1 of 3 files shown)


nmdc_server/api.py
```diff
@@ -637,6 +637,97 @@ async def download_zip_file(
...
+    data_rows = []
+    for s in submissions:
+
+        metadata = s.metadata_submission  # creates a concise alias
+        sample_data = metadata["sampleData"] if "sampleData" in metadata else {}
+        env_package = metadata["packageName"] if "packageName" in metadata else {}
+
+        # Get sample names from each sample type
+        for sample_type in sample_data:
+            samples = sample_data[sample_type] if sample_type in sample_data else {}
```
eecavanna, line 674: Assuming `samples` is typically a list (as opposed to a dictionary), I'd recommend defaulting to an empty list (instead of an empty dictionary) here.

```suggestion
            samples = sample_data[sample_type] if sample_type in sample_data else []
```

nmdc_server/api.py
```diff
@@ -637,6 +637,97 @@ async def download_zip_file(
...
+            # Iterate through each sample and extract the name
+            for x in samples:
+
+                # Get the sample name
+                name = x["samp_name"] if "samp_name" in x else {}
```
eecavanna, line 679: Assuming `name` normally contains a string, I'd recommend defaulting to an empty string here (instead of an empty dictionary).

```suggestion
                name = x["samp_name"] if "samp_name" in x else ""
```

I have the same (analogous) thought about `broad`, `local`, and `medium` below.

nmdc_server/api.py
```diff
@@ -637,6 +637,97 @@ async def download_zip_file(
...
+                name = str(name)
+                name = name.replace("\t", "").replace("\r", "").replace("\n", "").lstrip("_")
+                # Get the env broad scale
+                broad = x["env_broad_scale"] if "env_broad_scale" in x else {}
+                broad = str(broad)
+                broad = broad.replace("\t", "").replace("\r", "").replace("\n", "").lstrip("_")
+                # Get the env local scale
+                local = x["env_local_scale"] if "env_local_scale" in x else {}
+                local = str(local)
+                local = local.replace("\t", "").replace("\r", "").replace("\n", "").lstrip("_")
```
eecavanna, line 689: I'd prefer the variable not be named `local`—for two reasons:
1. The word "local" can have a special meaning in programming (i.e. local versus global) and I think future readers (of this code or stack traces involving this code) will find it confusing (at first glance).
2. I think "local" is an adjective describing "scale". Based on that, I think the value this variable will contain is some kind of "scale".

With those things in mind, I'd recommend renaming this variable to `local_scale` (or even `env_local_scale`).

I have the same thought (analogous) about `broad`.

nmdc_server/crud.py
```diff
@@ -686,6 +686,19 @@ def get_query_for_all_submissions(db: Session):
     return all_submissions
 
 
+def get_query_for_submitted_pending_review(db: Session):
+    r"""
+    Returns a SQLAlchemy query that can be used to retrieve submissions pending review.
```
eecavanna, line 691: 👍 Thanks for including a docstring!

nmdc_server/crud.py
```diff
@@ -686,6 +686,19 @@ def get_query_for_all_submissions(db: Session):
     return all_submissions
 
 
+def get_query_for_submitted_pending_review(db: Session):
```
eecavanna, line 689: I'd prefer this be named...

```diff
- def get_query_for_submitted_pending_review(db: Session):
+ def get_query_for_submitted_pending_review_submissions(db: Session):
```

...as a way of indicating that the function can be used to get _submissions_.

eecavanna (commented):
Hi @cristina-stonepedraza, I think this looks great overall! the code looks clean to me and I had no issues following it.

I left some feedback about path, function, and variable naming, as well as one about the endpoint description. In case you have any questions about these, you can message me here/Slack/etc.

There is one more thing I want to see in this PR:
- An automated test

~I'll send you information about that on Slack.~ I posted information about that [below](https://github.com/microbiomedata/nmdc-server/pull/1443#issuecomment-2475084966).

eecavanna:
Hi @cristina-stonepedraza, here's an example of an automated test that targets an endpoint like this one:

https://github.com/microbiomedata/nmdc-server/blob/36dbba5c9f4f82e438f3a940e9eb63b03c240982/tests/test_submission.py#L41-L125

There are actually two tests there because the endpoint in question is only accessible to admins. One of the tests focuses on that aspect (i.e. security).

By the way, I do wonder whether we will restrict access to _this_ reporting endpoint also. You/we can discuss that with @mslarae13.

mslarae13:
> By the way, I do wonder whether we will restrict access to this reporting endpoint also. You/we can discuss that with @mslarae13.

@eecavanna I assume the submission status end point that I use for quarterly metrics is limited, yes? I think this should be the same.

eecavanna:
Great commit messages, by the way, @cristina-stonepedraza!

Next week, I can walk you through the processes of logging into the Data Portal such that you can access protected endpoints via the Swagger UI page; and how you can "promote" yourself into an admin locally.

eecavanna:
Just passing through as I've been seeing the notifications from GitHub. The code is looking excellent (in my opinion)! 👏

cristina-stonepedraza:
@eecavanna I've got this to a place where all the tests pass, I think it's ready for another look when you have a chance!

tests/test_submission.py
```diff
@@ -38,6 +38,75 @@ def test_list_submissions(db: Session, client: TestClient, logged_in_user):
...
+            },
+            "packageName": "Env Pkg 1",
+        },
+    )
+    db.commit()
+    response = client.request(method="get", url="/api/metadata_submission/mixs_report")
+    assert response.status_code == 200
+
+    # Check that the reponse payload is a TSV and that the result has the correct
+    # number of rows and information populated. The result should have 4 rows
```
eecavanna, line 71: After having seen the test submission above having two biosamples in it, I expected this number to be 3 (i.e. 2 data rows + 1 header row = 3 total rows) instead of 4.

```suggestion
    # number of rows and information populated. The result should have 3 rows
```

eecavanna (commented):
Excellent! Great job, @cristina-stonepedraza!

I found the code and comments easy to read and follow, and I think people will find it easy to maintain moving forward.

I left one suggestion about a number in a comment. I think it may have been left over from an earlier plan for the test.

There is one more thing I want to see included in this PR, which is: making the endpoint accessible to admin users only. That is based upon @mslarae13's comment [here](https://github.com/microbiomedata/nmdc-server/pull/1443#issuecomment-2477796296). I'll send you some information about that shortly.

eecavanna:
If you add this parameter to the endpoint request handler function, the endpoint will become accessible to **logged-in users** only (but does not ensure they are admins—that will come next):

https://github.com/microbiomedata/nmdc-server/blob/c408d3430e2cdc50af54e273e43d0039e0aaff29/nmdc_server/api.py#L666

If you then add this snippet to the beginning of the function's body, the endpoint will become accessible to logged-in users **who are admins** only:

https://github.com/microbiomedata/nmdc-server/blob/c408d3430e2cdc50af54e273e43d0039e0aaff29/nmdc_server/api.py#L672-L673

In terms of testing, you can add a test like this to demonstrate that a logged-in user that is not an admin is unable to access the endpoint:

https://github.com/microbiomedata/nmdc-server/blob/c408d3430e2cdc50af54e273e43d0039e0aaff29/tests/test_submission.py#L41-L45

Finally, in the existing test you've already written, you can replace the existing logged-in user with a logged-in **admin** user by following the example in this snippet. Here, we make the test require a logged-in admin user, then we use a variable called `logged_in_user` to refer to that logged-in **admin** user so that the code below it can remain the same (since it already uses the latter variable name).

https://github.com/microbiomedata/nmdc-server/blob/c408d3430e2cdc50af54e273e43d0039e0aaff29/tests/test_submission.py#L48-L54

eecavanna (approved):
Looks great! Thanks for adding the "is this user an admin" check. The last thing I want to do before this gets merged in is confirm that the user-facing behavior (including the shape of the TSV file) matches @mslarae13's expectations. Once you have that confirmation, I'd be comfortable with this being merged into `main`. Great job!

eecavanna:
I'm also going to tag @naglepuff and @pkalita-lbl here as they, too, maintain the Data Portal API and I want them to be aware that this branch may be merged into the `main` branch soon.

FYI, @naglepuff and @pkalita-lbl: I've reviewed this branch and approve of it being merged into `main` once @mslarae13 has approved of its user-facing behavior.

cristina-stonepedraza:
Here is a screenshot of part of the resulting TSV produced by the API endpoint for reference. 
Screenshot 2024-11-26 081434

mslarae13 (approved):

mslarae13:
@eecavanna any idea when this will be in production? With the next release?

mslarae13:
IF IT’S ON DEV NOW IT WILL BE PART OF THE NEXT RELEASE ALWAYS AND FOR EVERY CHANGE
:)

Missed the last release. Will be in production with the next release.

eecavanna:
Yup! It'll be deployed to production during next Monday's 2024.12 NMDC Release.

PR #1445, opened 2024-11-13
nginx payload size config and sample metadata loading indicator

### LINKED ISSUE #1444: Allow configuration of nginx `client_max_body_size` through environment
Recently we had a report from a user who was unable to save their submission's sample metadata. They reported that the saving spinner would just get stuck:

image

I was able to reproduce the issue by:

1. Create a submission using the soil template
2. Download the Excel file for the soil template, populate the file with about 2500 rows, filling in all the required columns (the exact values used aren't terribly important)
3. Import the updated Excel file. At this point the imported metadata appears to have been saved correctly, but the network request actually failed with a `413 (Request Entity Too Large)` error:

image

I believe the root cause of the issue is that we do not set `client_max_body_size` in our nginx config. The default value is 1 MB, and using the above reproduction steps the submission JSON object will be about 2 MB.

I think we need to address this with a few changes:

1. Update `web/nginx.conf.template` and `web/start.sh` to set `client_max_body_size` and ideally set in a way that it's configurable by an environment variable so that we can adjust it on the fly if needed.
2. Ensure that the "Saving progress" spinner is triggered by importing an XLSX file. The save does happen (or, is attempted at least), but the UI doesn't indicate that.
3. Ensure that the "Saving progress" spinner goes away and the message about "Failed to save changes" is displayed when the request fails. The logic is mostly in place, but it's faulty. [This call](https://github.com/microbiomedata/nmdc-server/blob/0fe0163c12f4df088dcc775496e6e9c560e01a05/web/src/views/SubmissionPortal/HarmonizerView.vue#L183) to `incrementalSaveRecord` should be wrapped in a try/catch since it will throw an exception when the request fails instead of continuing to subsequent lines (where `isSaving` is set to `false`).

pkalita-lbl:
Fixes #1444 

### Summary

These changes address two items exposed by a user attempting to save a large set of sample metadata:

1. The save request was being rejected because of nginx's maximum payload size setting
2. The UI was not telling the user that the request failed

### Details

First, the [`client_max_body_size`](https://nginx.org/en/docs/http/ngx_http_core_module.html#client_max_body_size) directive has been added to the nginx config template. It is populated with the value of the `NGINX_CLIENT_MAX_BODY_SIZE` environment variable. If this variable is not set manually, it will get set to `10m` via `web/start.sh`. This will allow payload sizes of 10 MB (the default is 1 MB). We can also choose to set the `NGINX_CLIENT_MAX_BODY_SIZE` in Rancher if we need to adjust the value on the fly for whatever reason.

Second, I updated `HarmoinzerView.vue` so that all of the `incrementalSaveRecord` calls are made via `useRequest`. This means that the loading and error state tracking doesn't need to be done manually, and it can be done uniformly for every call. The previous loading and error state tracking was a bit faulty because `await incrementalSaveRecord(...)` will raise an exception if the underlying HTTP request fails, meaning the bit where `isSaving` was set to `false` was never reached. These changes also make it so that the loading indicator is shown whenever a save is happening (including after importing an XLSX file).

### DIFF
--- a/web/nginx.conf.template
+++ b/web/nginx.conf.template
@@ -14,6 +14,8 @@ http {
 
         server_name localhost;
 
+        client_max_body_size ${NGINX_CLIENT_MAX_BODY_SIZE};
+
         location / {
             try_files ${DOLLAR}uri ${DOLLAR}uri/ /index.html;
         }
--- a/web/src/views/SubmissionPortal/HarmonizerView.vue
+++ b/web/src/views/SubmissionPortal/HarmonizerView.vue
@@ -171,19 +171,14 @@ export default defineComponent({
       ])),
     ));
 
-    const isSaving = ref();
-    const saveSuccess = ref();
-    const onBeforeChange = () => {
-      isSaving.value = true;
-    };
+    const saveRecordRequest = useRequest();
+
     const onDataChange = async () => {
       hasChanged.value += 1;
       const data = harmonizerApi.exportJson();
       mergeSampleData(activeTemplate.value.sampleDataSlot, data);
-      const httpStatus = await incrementalSaveRecord(root.$route.params.id);
+      await saveRecordRequest.request(() => incrementalSaveRecord(root.$route.params.id));
       tabsValidated.value[activeTemplateKey.value] = false;
-      saveSuccess.value = httpStatus === 200;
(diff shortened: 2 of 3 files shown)


naglepuff (approved):
Code LGTM!

PR #1446, opened 2024-11-14
Add delete submission button and overflow menu

JamesTessmer:
This PR is linked to issue: https://github.com/microbiomedata/issues/issues/826
The goal is to add a delete submission button inside an overflow menu that could be added to later.
Clicking the delete button will show a confirmation menu, and when confirmed the submission will be deleted from the front-end schema.

### DIFF
--- a/web/src/views/SubmissionPortal/Components/SubmissionList.vue
+++ b/web/src/views/SubmissionPortal/Components/SubmissionList.vue
@@ -68,8 +68,8 @@ export default defineComponent({
       mustSort: false,
     });
     const deleteDialog = ref(false);
-    let deleteConfirmation = false;
-    let dialogUpdated = false;
+    const deleteConfirmation = ref(false);
+    const dialogUpdated = ref(false);
 
     function getStatus(item: api.MetadataSubmissionRecord) {
       const color = item.status === submissionStatus.Complete ? 'success' : 'default';
@@ -91,7 +91,7 @@ export default defineComponent({
     function waitForDialogUpdate(): Promise<void> {
       return new Promise<void>((resolve) => {
         const intervalId = setInterval(() => {
-          if (dialogUpdated === true) {
+          if (dialogUpdated.value === true) {
             clearInterval(intervalId);
             resolve();
           }
@@ -109,14 +109,14 @@ export default defineComponent({
       }
 
       deleteDialog.value = false;
-      deleteConfirmation = false;
-      dialogUpdated = false;
+      deleteConfirmation.value = false;
+      dialogUpdated.value = false;
       return item;
     }
 
     function deleteDialogUpdate(confirmation: boolean) {
-      deleteConfirmation = confirmation;
-      dialogUpdated = true;
+      deleteConfirmation.value = confirmation;
+      dialogUpdated.value = true;
     }
 
     async function handleOverflowMenu(item: api.MetadataSubmissionRecord, title: String) {
@@ -125,7 +125,7 @@ export default defineComponent({
           deleteSubmission(item);
           break;
         default:
-          console.log('Something went wrong');
+          break;
       }
     }
 
@@ -297,7 +297,6 @@ export default defineComponent({
       <v-dialog
         v-model="deleteDialog"
         class="ma-5"
-        persistent
       >
         <v-card>
           <v-card-title class="text-h5">
--- a/web/src/views/SubmissionPortal/Components/SubmissionList.vue
+++ b/web/src/views/SubmissionPortal/Components/SubmissionList.vue
@@ -216,39 +216,41 @@ export default defineComponent({
             </v-chip>
           </template>
           <template #[`item.action`]="{ item }">
-            <v-btn
-              small
-              color="primary"
-              @click="() => resume(item)"
-            >
-              Resume
-              <v-icon class="pl-1">
-                mdi-arrow-right-circle
-              </v-icon>
-            </v-btn>
-            <v-menu
-              offset-x
-            >
-              <template #activator="{ on }">
-                <v-btn
-                  text
-                  icon
-                  class="ml-1"
-                  v-on="on"
-                >
-                  <v-icon>
-                    mdi-dots-vertical
-                  </v-icon>
-                </v-btn>
-              </template>
-              <v-list>
-                <v-list-item
-                  @click="() => handleOpenDeleteDialog(item)"
-                >
-                  <v-list-item-title>Delete</v-list-item-title>
-                </v-list-item>
-              </v-list>
-            </v-menu>
+            <div class="d-flex align-center">
+              <v-btn
+                small
+                color="primary"
+                @click="() => resume(item)"
+              >
+                Resume
+                <v-icon class="pl-1">
+                  mdi-arrow-right-circle
+                </v-icon>
+              </v-btn>
+              <v-menu
+                offset-x
+              >
+                <template #activator="{ on }">
+                  <v-btn
+                    text
+                    icon
+                    class="ml-1"
+                    v-on="on"
+                  >
+                    <v-icon>
+                      mdi-dots-vertical
+                    </v-icon>
+                  </v-btn>
+                </template>
+                <v-list>
+                  <v-list-item
+                    @click="() => handleOpenDeleteDialog(item)"
+                  >
+                    <v-list-item-title>Delete</v-list-item-title>
+                  </v-list-item>
+                </v-list>
+              </v-menu>
+            </div>
           </template>
         </v-data-table>
       </v-card>
--- a/web/src/views/SubmissionPortal/Components/SubmissionList.vue
+++ b/web/src/views/SubmissionPortal/Components/SubmissionList.vue
@@ -67,6 +67,9 @@ export default defineComponent({
       multiSort: false,
       mustSort: false,
     });
+    const deleteDialog = ref(false);
+    let deleteConfirmation = false;
+    let dialogUpdated = false;
 
     function getStatus(item: api.MetadataSubmissionRecord) {
       const color = item.status === submissionStatus.Complete ? 'success' : 'default';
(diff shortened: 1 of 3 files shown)


web/src/views/SubmissionPortal/Components/SubmissionList.vue
```diff
@@ -203,6 +219,31 @@ export default defineComponent({
               </v-icon>
             </v-btn>
           </template>
+          <template #[`item.menu`]="{ item }">
+            <v-menu>
+              <template #activator="{ on }">
+                <v-btn
+                  class="ml-1"
+                  icon="mdi-dots-vertical"
```
marySalvi, line 227: I know you aren't done yet, but just wanted to point out that the `icon` prop is a boolean and doesn't take a string. It indicates a _type_ of button, in this case round and 'text' (transparent). Therefore using the icon prop makes the `text` prop redundant. 
Admittedly icon is not an intuitive name for this behavior.

JamesTessmer: That makes a lot of sense why it was behaving unexpectedly. Thanks for pointing it out. I think that format is from vuetify 1 since I formatted it after those docs. I’m planning to go through and update all those kinds of things after looking at the vuetify2 docs some more.

JamesTessmer: Made this update

JamesTessmer:
Screenshot 2024-12-05 at 11 18 01 AM
Screenshot 2024-12-05 at 11 17 53 AM
Here's what the dialog and overflow menu currently look like.
There is a weird scroll bar on the dialog I couldn't quite figure out how to remove.

web/src/views/SubmissionPortal/Components/SubmissionList.vue
```diff
@@ -202,9 +259,69 @@ export default defineComponent({
...
+                >
+                  <v-list-item-title>{{ entry.title }}</v-list-item-title>
+                </v-list-item>
+              </v-list>
+            </v-menu>
           </template>
         </v-data-table>
       </v-card>
     </v-card>
+    <v-row justify="center">
```
pkalita-lbl, line 291: Is the `v-row` needed here?

web/src/views/SubmissionPortal/Components/SubmissionList.vue
```diff
@@ -202,9 +259,69 @@ export default defineComponent({
...
+      <v-dialog
+        v-model="deleteDialog"
+        :width="630"
+        class="ma-5"
```
pkalita-lbl, line 295: I think the reason that you're getting the unwanted scrollbars is because of the margin classes (`ma-*`) here and below.

web/src/views/SubmissionPortal/Components/SubmissionList.vue
```diff
@@ -202,9 +259,69 @@ export default defineComponent({
...
+      >
+        <v-card>
+          <v-spacer />
+          <v-card-title class="ma-3 text-h4">
```
pkalita-lbl, line 299: Ditto margin class comment. I would also recommend using `text-h5` here. `text-h4` appear a bit too large to me.

web/src/views/SubmissionPortal/Components/SubmissionList.vue
```diff
@@ -202,9 +259,69 @@ export default defineComponent({
...
+            Are you sure you want to delete this submission in progress?
+          </v-card-title>
+          <v-card-text class="ma-3 text-h5">
```
pkalita-lbl, line 302: Can probably remove both of these classes.

web/src/views/SubmissionPortal/Components/SubmissionList.vue
```diff
@@ -202,9 +259,69 @@ export default defineComponent({
...
+            >
+              Delete
+            </v-btn>
+            <v-btn
+              color="primary"
+              class="ma-3"
+              @click="deleteDialogUpdate(false)"
+            >
+              Cancel
+            </v-btn>
```
pkalita-lbl, line 321: I'd recommend swapping the order of these buttons. I think it's generally more standard to have the primary action (the delete in this case) on the right. 

I'd also recommend not using the `primary` color on the Cancel button. Having both a purple and red button next to each other creates a bit of confusing visual hierarchy (that is, the primary action isn't as visually distinct from the other).

web/src/views/SubmissionPortal/Components/SubmissionList.vue
```diff
@@ -202,9 +259,69 @@ export default defineComponent({
...
+    <v-row justify="center">
+      <v-dialog
+        v-model="deleteDialog"
+        :width="630"
+        class="ma-5"
+      >
+        <v-card>
+          <v-spacer />
+          <v-card-title class="ma-3 text-h4">
+            Are you sure you want to delete this submission in progress?
```
pkalita-lbl, line 300: If we keep a ref of which submission triggered the dialog (i.e. `deleteDialogSubmission` in my previous comment, then we can use that to add more contextual information here. That is, it can be something like:

```vue
<p v-if="deleteDialogSubmission">
  <!-- todo: handle case where study_name is blank -->
  This will delete "{{ deleteDialogSubmission.study_name }}" and all associated data.
</p>
<p>This cannot be undone.</p>
```

web/src/views/SubmissionPortal/Components/SubmissionList.vue
```diff
@@ -202,9 +259,69 @@ export default defineComponent({
...
+          <v-card-text class="ma-3 text-h5">
+            This cannot be undone.
+          </v-card-text>
+          <v-divider />
+          <v-card-actions>
+            <v-spacer />
+            <v-btn
+              color="red"
+              class="ml-3 white--text"
+              @click="deleteDialogUpdate(true)"
```
pkalita-lbl, line 311: Following example from previous comment:

```
@click="handleDelete(deleteDialogSubmission)"
```

web/src/views/SubmissionPortal/Components/SubmissionList.vue
```diff
@@ -202,9 +259,69 @@ export default defineComponent({
...
+            >
+              Delete
+            </v-btn>
+            <v-btn
+              color="primary"
+              class="ma-3"
+              @click="deleteDialogUpdate(false)"
```
pkalita-lbl, line 318: Following example from previous comment:

```
@click="isDeleteDialogOpen = false"
```

pkalita-lbl (changes requested):
I think some of the state and event handling can be simplified quite a bit here. Let me know if you want to meet to go over this in more detail!

web/src/views/SubmissionPortal/Components/SubmissionList.vue
```diff
@@ -80,26 +84,79 @@ export default defineComponent({
...
       options,
       submission,
     };
   },
+  data: () => ({
+    OverFlowMenuItems: [
+      { title: 'Delete' },
+      { title: 'Future button' },
+    ],
+  }),
```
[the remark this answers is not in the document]
JamesTessmer: My initial thought with this structure was that it could make adding list options easier in the future and that's why it's set up to be kind of dynamic, but in hindsight I agree. I don't think it actually makes adding things any easier really since extra work would be required in the `handleOverflowMenu` code too.

mslarae13:
> Screenshot 2024-12-05 at 11 18 01 AM Screenshot 2024-12-05 at 11 17 53 AM
> Here's what the overflow menu and dialog currently look like. There is a weird scroll bar on the dialog I couldn't quite figure out how to remove.
> 
> BEFORE merging the 'future button' will be removed. That's just in the current PR and branch to show how it would look with multiple buttons in the future and to test handling of multiple buttons.

No comments on the code, I defer to @pkalita-lbl from a content perspective, I'm good with this.

web/src/views/SubmissionPortal/Components/SubmissionList.vue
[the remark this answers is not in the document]
JamesTessmer: Added this change

web/src/views/SubmissionPortal/Components/SubmissionList.vue
```diff
@@ -202,9 +259,69 @@ export default defineComponent({
...
+                </v-btn>
+              </template>
+              <v-list>
+                <v-list-item
+                  v-for="(entry, i) in OverFlowMenuItems"
+                  :key="i"
+                  @click="() => handleOverflowMenu(item,entry.title)"
+                >
+                  <v-list-item-title>{{ entry.title }}</v-list-item-title>
+                </v-list-item>
```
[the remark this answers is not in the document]
JamesTessmer: added this change

web/src/views/SubmissionPortal/Components/SubmissionList.vue
```diff
@@ -80,26 +84,79 @@ export default defineComponent({
...
+      switch (title) {
+        case 'Delete':
+          if (getStatus(item).text === 'in-progress') {
+            handleDeleteSubmission(item);
+          }
+          break;
+        default:
+          break;
+      }
+    }
```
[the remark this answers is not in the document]
JamesTessmer: Added this change

web/src/views/SubmissionPortal/Components/SubmissionList.vue
JamesTessmer: Nothing changes when it's removed so it's now deleted.

web/src/views/SubmissionPortal/Components/SubmissionList.vue
JamesTessmer: removed margin classes, scroll bar issue was resolved.

web/src/views/SubmissionPortal/Components/SubmissionList.vue
JamesTessmer: Done

web/src/views/SubmissionPortal/Components/SubmissionList.vue
JamesTessmer: Removed

web/src/views/SubmissionPortal/Components/SubmissionList.vue
JamesTessmer: Added this change

web/src/views/SubmissionPortal/Components/SubmissionList.vue
JamesTessmer: Added this change

web/src/views/SubmissionPortal/Components/SubmissionList.vue
JamesTessmer: Added this change

web/src/views/SubmissionPortal/Components/SubmissionList.vue
JamesTessmer: Added the change, and a check to see if deleteDialogSubmission is null before making the dialog visible.

pkalita-lbl:
Sorry, one more thing that I don't know how I missed before. At narrower browser width breakpoints, it looks like we're getting a bit of undesired wrapping between the Resume button and the menu trigger button:

image

I'd recommend wrapping those two elements in a [flexbox](https://v2.vuetifyjs.com/en/styles/flex/), like so:

```html
<div class="d-flex align-center">
  <v-btn>...</v-btn>
  <v-menu>...</v-menu>
</div>
```

pkalita-lbl (approved):

PR #1465, opened 2024-11-21
Enable gene function search on GO terms

### LINKED ISSUE #1388: Data Portal - backend support GO facet search via mappings - Pfam
Search for a GO term in the data portal should use a mapping file to check mapping files since the GO terms aren't stored directly in the data files.

The Pfam -> GO mapping file is https://current.geneontology.org/ontology/external2go/pfam2go

The code should be able to handle multiple mapping files, there will be a related ticket for the KEGG to GO mapping file.

naglepuff:
Fix #1388 

#### Data Model
New tables: `go_term_text` and `go_term_to_pfam_entry` to model GO terms and how they map to PFAM entries. A new migration has been added to create the tables.

#### Ingest
Ingest GO terms by iterating over the "nodes" in the graph defined by: http://current.geneontology.org/ontology/go-basic.json

Ingest PFAM entry to GO term mapping using a file derived from: [current.geneontology.org/ontology/external2go/pfam2go](https://current.geneontology.org/ontology/external2go/pfam2go)

#### API
New endpoint for text search of GO terms.

#### Query
New logic that transforms conditions using GO terms to their associated PFAM entries.

#### UI
New facet for GO has been added.

### Testing
To test, you'll have to have to run a local ingest. Make sure you obtain the new file added to NERSC: `.../data/ingest/go/pfam_go_mappings.txt`.

Verify ingest populates the `go_term_text` and `go_term_to_pfam_entry` tables, and searching on GO terms works through the UI. That is, make sure searching for GO terms works using the new endpoint, and make sure you get the results you'd expect when searching using this facet. Pick some PFAM value from the `gene_function` table, use the new mapping table to find associated GO terms, and use those for testing. For example, GO:0004930 maps to PF00001


marySalvi (approved):
Looks Good!

naglepuff:
BEFORE MERGE: Update the down-revision of the migration introduced here. (#1471 has been merged an includes a migration)

PR #1466, opened 2024-11-22
Post message to Slack when ingest completes successfully

eecavanna:
In this branch, I updated the ingest CLI script so that, when it finishes a successful run, **if a Slack incoming webhook URL is defined**, the script uses that URL to post a message to a Slack channel. This technique for posting messages to Slack is documented [here](https://api.slack.com/messaging/webhooks).

Outside of this PR, I have created a Slack app and—for that Slack app—generated an incoming webhook URL. That is what I plan to use on Spin (in all of our environments) once this code gets deployed there. For any reviewers that intend to test this locally, I can also provide that URL to you via Snappass (or Spin). Using that specific URL, the Slack message will be posted to the newly-created `#ingest-notifications` channel in the NMDC Slack workspace.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -21,3 +21,7 @@ NMDC_ENVIRONMENT="development"
 
 # Uncomment to enable CORS for local development of the nmdc-field-notes
 # NMDC_CORS_ALLOW_ORIGINS=capacitor://localhost,ionic://localhost,http://localhost,http://127.0.0.1:8100
+
+# (Optional) Slack incoming webhook URL the ingester can use to post messages to Slack.
+# Reference: https://api.slack.com/messaging/webhooks#create_a_webhook
+# SLACK_WEBHOOK_URL_FOR_INGESTER=changeme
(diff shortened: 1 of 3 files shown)


pkalita-lbl (approved):

eecavanna:
Thanks for reviewing/approving this.

#### Spin/Rancher has been configured with the Slack incoming webhook URL

Before merging this branch in, I have added the Slack incoming webhook URL to Rancher. 

Specifically, in both the development and production environments (i.e. `nmdc-dev` and `nmdc` namespaces): I created an opaque secret named `slack`, which has an item whose key is `SLACK_WEBHOOK_URL_FOR_INGESTER`and whose value is that Slack incoming webhook URL. Then, I configured the `portal-ingest` CronJob so that it has an environment variable named `SLACK_WEBHOOK_URL_FOR_INGESTER` whose value comes from that secret key (as shown in the following screenshot).

image

As a result, once this gets merged into `main`, the next time the ingest runs successfully in the development environment, I expect a message to appear in the `#ingest-notifications` Slack channel.

In addition, after the next monthly NMDC release (i.e. when this code gets deployed to production), I expect a message to appear in that Slack channel whenever ingest runs successfully in production.

PR #1470, opened 2024-11-26
Add KEGG/GO Mappings

### LINKED ISSUE #1389: Data Portal - backend support GO faceted search via mappings - KEGG
Search for a GO term in the data portal should use a mapping file to check mapping files since the GO terms aren't stored directly in the data files.

The KEGG -> GO mapping file is https://www.genome.jp/kegg-bin/show?file=ko2go.xl

The code should be able to handle multiple mapping files, there is a related Pfam ticket. #1388 

cc @sierra-moxon

naglepuff:
Fix #1389 

#### 1. Update Postgres Models
Adds new model and table to track KEGG/GO mappings.

#### 2. Update Ingest to Populate New Tables
Uses the data found here: https://www.genome.jp/kegg-bin/show?file=ko2go.xl, which has been downloaded as a file and added to NERSC.

#### 3. Update Query Logic
Modify the SQLAlchemy used to translate a filter on GO terms to a filter on both KEGG and PFAM.

### Note:
This builds on top of #1465 (note the base branch for this PR).


naglepuff:
Closing in favor of #1477

PR #1473, opened 2024-11-27
Add ability to store Field Notes-related metadata on submissions

pkalita-lbl:
Related to https://github.com/microbiomedata/nmdc-field-notes/issues/150

### Summary

We have a use case for the Field Notes app where we want to attach a bit of Field Notes-specific information to a submission and have it _shared_ between all users with access to that submission (i.e. the app's local storage wont cut it). These changes add a `field_notes_metadata` column to the `submission_metadata` table. They also update the API to provide this information in `GET` requests and to accept this information in `PATCH` requests.

### Details

The `field_notes_metadata` column which is being added here is a JSONB column. I'm sort of intentionally not strongly typing the column any further than `Optional[Dict[str, Any]]`. My thought is that this column is a black box as far as `nmdc-server` is concerned (the Field Notes code will, however, make [stronger claims](https://github.com/microbiomedata/nmdc-field-notes/blob/e9a23831bc79ee29364ebc5a5f3d4d1ee7169489/src/api.ts#L179-L188) about what format it will save to and retrieve from the field). This also should allow us to expand what Field Notes uses `field_notes_metadata` for in the future without `nmdc-server` changes.

A bit of an aside, but I got called out by `mypy` for the `JSONB()` call in the new migration. I saw that elsewhere in the code standard practice was to just add a `type: ignore`. I was curious and started poking around, and I found that the [`sqlalchemy-stubs`](https://github.com/dropbox/sqlalchemy-stubs) package is more-or-less unmaintained at this point and doesn't support SQLAlchemy 1.4 (which we use) particularly well (hence the need for `type: ignore`). The SQLAlchemy 1.4 [docs](https://docs.sqlalchemy.org/en/14/orm/extensions/mypy.html#installation) now point you to the [`sqlalchemy2-stubs`](https://github.com/sqlalchemy/sqlalchemy2-stubs/) package. I was hoping maybe I could just drop that in as a replacement. But I gave it a whirl, and it came up with _a lot_ of typing errors (but not the JSONB one!) So I guess just something to consider in our expansive free time.


nmdc_server/migrations/versions/ceb93b48d6ae_add_field_noes_metadata.py
```diff
@@ -0,0 +1,34 @@
...
+
+from typing import Optional
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision: str = "ceb93b48d6ae"
+down_revision: Optional[str] = "4aee1e10bb24"
```
naglepuff, line 17: #1465 was just merged, which introduced a migration, you'll have to update the `Revises:` comment and the `down_revision` identifier here.

naglepuff (commented):
There's just one thing you'll need to do here before merging, but other than that comment, looks good to me.

naglepuff (approved):

PR #1474, opened 2024-11-28
Update ingester to report more events via Slack

eecavanna:
On this branch, I made changes to the ingester, related to sending Slack messages.

First, I extracted the code that sends a Slack message into a helper function, so it can be called from more places.

Then, I made it so the ingester sends a Slack message whenever any of the following events occurs:
- **Ingest starts**
- **Ingest finishes successfully** (note: the ingester was already sending a Slack message when this event would occur)
- **Certain types of errors occur** (one of which is the failure to obtain the ingest lock)

Each message includes the ingest's **start time** and the **MongoDB host name**. My goals were (a) that Slack users find it easy to associate multiple messages from the same ingest run, with one another; and (b) that Slack users find it easy to know what environment the ingest was running in. I didn't include the Postgres host name because it is stored in an environment variable that also contains the Postgres password.

In addition to those two pieces of information, the message that gets sent when ingest fails includes the **Exception message**; and the message that gets sent when ingest finishes successfully includes the **duration** in minutes.

I also updated a few lines of the developer docs based upon some recent experience (not related to ingest).


eecavanna:
I was surprised to see this most recent error from the GitHub Actions workflow run: "`ingest` is too complex"

image

I take it the changes in the most recent commit pushed the function past some threshold of (overall) "complexity," not that the specific code I added, itself, is particularly "complex."

eecavanna:
### Still to do

- [ ] 1. Starting message: Include start time
- [ ] 2. Ending message: Include end time, or start time and duration
- [ ] 3. Include environment distinguisher (e.g. Mongo hostname or Spin namespace)
- [ ] 4. Send messages when an error occurs

I will ask for help from other maintainers about item 4 after the 🦃  holiday.

eecavanna:
I haven't tested these changes yet, as I have never run ingest locally before. Tomorrow, I'll look into testing them.

naglepuff (approved):

pkalita-lbl (approved):

eecavanna:
Thanks for reviewing/approving this.

I also tested the helper function, by defining some mock dependencies manually.

image

_(Lines 21-50 contain the function-under-test, `send_slack_message`)_

image

I'll merge it in now.

PR #1481, opened 2024-12-05
Support Lipidome Results

### LINKED ISSUE #1464: Data portal- support lipidome results
We expect to add lipidome records along with workflow results. We need lipodome added in the following:
- [x] bar chart
- [x] facted omics type search
- [x] upset plot
- [x] tick boxes below samples
- [ ] be able to download workflow results 

@kheal added test data to mongo dev in https://github.com/microbiomedata/issues/issues/944

cc @shreddd @ssarrafan 

Needs prioritization, we expect to have lipidome results in mongo in December and would like to highlight this in the annual report which needs to be finalized mid-January 2024.

naglepuff:
Fix #1464 

#### 1. Updates to Ingest
Incoming `data_generation` documents with an `analyte_category` of `"lipidome"` are now ingested correctly, and the associated fields in postgres, including the `multiomics` bitmap are populated with that in mind. (The bitmap used for type now uses 6 bits instead of 5).

#### 1. Lipidomics added to upset plot
Added Lipidomics as a new category and updated the bit manipulation on the frontend to makes sure the "multiomics" facet works with lipidomics.

image

#### 2. Uncommented filters that took out lipidomics data
Previously there were explicit filters in front end code that prevented display of lipidomics data. These have been removed. As a result, lipidomics now appear in the bar chart, the flyout for the Data Generation type facet, and as a drop-down in the biosample result section.

image

image

image

### DIFF
--- a/nmdc_server/database.py
+++ b/nmdc_server/database.py
@@ -145,15 +145,17 @@ with m as (select
     bit_or(
         case
             when op.annotations->>'omics_type' = 'Metabolomics' then
-                b'{MultiomicsValue.mb.value:05b}'
+                b'{MultiomicsValue.mb.value:06b}'
             when op.annotations->>'omics_type' = 'Metagenome' then
-                b'{MultiomicsValue.mg.value:05b}'
+                b'{MultiomicsValue.mg.value:06b}'
             when op.annotations->>'omics_type' = 'Proteomics' then
-                b'{MultiomicsValue.mp.value:05b}'
+                b'{MultiomicsValue.mp.value:06b}'
             when op.annotations->>'omics_type' = 'Metatranscriptome' then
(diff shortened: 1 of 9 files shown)


naglepuff:
@marySalvi this is ready. The issues brought up during the infra sync meeting have been addressed. Let me know if you need any help testing. Ingesting from dev should get you all the data you need, and the specific records can be found in this issue: https://github.com/microbiomedata/issues/issues/944#issuecomment-2489873716

aclum:
@marySalvi when do you expect to be able to review this PR? This was something we would like in the December release so would be nice to test this week.

marySalvi (approved):
LGTM. The wrapping solution looks good

PR #1488, opened 2024-12-09
Upgrade to `nmdc-schema` v11.2.1

pkalita-lbl:
Standard upgrade of `nmdc-schema` and `nmdc-submission-schema` in advance of next week's release.


web/package.json
```diff
@@ -90,5 +90,6 @@
     "node-forge": "^1.3.0",
     "serialize-javascript": "^3.1",
     "http-proxy": "1.18.1"
-  }
+  },
+  "packageManager": "yarn@1.22.19+sha1.4ba7fc5c6e704fce2066ecbfb0b0d8976fe62447"
```
naglepuff, line 94: Does this just make sure the correct version of `yarn` is used for this project?

pkalita-lbl: Whoa, I didn't even notice that got added; I didn't do anything intentional to add it. If I'm reading [the docs](https://nodejs.org/api/packages.html#packagemanager) right, I guess it happened because I have `yarn` installed via Corepack? 

Do you want to leave it there and see if it causes any issues?

naglepuff: That sounds about right. It doesn't seem to affect me at all (I've used `corepack` only a little bit on other projects). I don't have a problem with leaving it in. I don't know what the behavior would be if someone ran `yarn` in this project without having enabled corepack.

naglepuff (approved):

web/package.json
pkalita-lbl: My takeaway from the docs is that if you haven't enabled corepack then the `packageManager` field will be ignored. But if you find that it's causing any issues just let me know.

PR #1492, opened 2024-12-11
Fix GO search bugs

### LINKED ISSUE #1462: Data portal - UI updates for GO functional search
This should be implemented in a similar fashion to https://github.com/microbiomedata/nmdc-server/issues/1338

Depends on 
 - [x] #1389 
 - [x] #1388 

Expansion prefix http://purl.obolibrary.org/obo/GO_

General documentation link:
https://geneontology.org/docs/ontology-documentation/

cc @sierra-moxon 

This is a target for the Dec 16, 2024 release. It is part of [milestone 2.12.2](https://github.com/microbiomedata/issues/issues/518) which is due in Q1 FY25 (end of Dec) 

This ticket completes the cluster of related tickets to expand functional annotation search.

naglepuff:
Fix #1462 

### Changes
This fixes some bugs that occur when there is no mapping from a GO term to either KEGG or PFAM. It does NOT prevent the user from selecting such terms.

#### Currently, in dev
image

A search for this particular GO term incorrectly reports that it matches ALL samples. The search function technically fails to handle this case, since it expects that the GO term has at least one mapped term.

#### In this branch
The same query now (correctly) returns no biosample results. The code that translates the GO term to KEGG and PFAM terms handles the case where no such mappings exist. Additionally, some of the logic that transforms search terms into those stored in the database has been streamlined.
image

### DIFF
--- a/nmdc_server/query.py
+++ b/nmdc_server/query.py
@@ -259,6 +259,16 @@ ConditionSchema = Union[
 ]
 
 
+def _transform_gene_term(term: tuple[str, Any]) -> str:
+    if term[0].startswith("KO:K"):
+        return term[0].replace("KO:K", KeggTerms.ORTHOLOGY[0])
+    if term[0].startswith("COG"):
+        return term[0].replace("COG", "COG:COG")
+    if term[0].startswith("PF"):
+        return term[0].replace("PF", "PFAM:PF")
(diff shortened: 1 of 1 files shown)


marySalvi (approved):
LGTM!

PR #1495, opened 2024-12-17
Add initial button for test submission

JamesTessmer:
This PR is for https://github.com/microbiomedata/issues/issues/827

This PR will includes:
A new button to support creating test submissions.
These submissions have supporting logic to keep them from being submitted as an actual submission and an added tooltip explaining when to use a test submission.


web/src/views/SubmissionPortal/Components/SubmissionList.vue
```diff
@@ -172,11 +172,31 @@ export default defineComponent({
...
-          @click="createNewSubmission"
+          @click="createNewSubmission(false)"
         >
           <v-icon>mdi-plus</v-icon>
-          Create New Submission
+          Create Submission
         </v-btn>
+        <v-btn
+          color="primary"
+          class="ml-3"
```
pkalita-lbl, line 182: In order to not have two visually prominent primary buttons next to each other, I'd recommend adding the `outlined` prop here. More details in Slack conversation.

web/src/views/SubmissionPortal/HarmonizerView.vue
```diff
@@ -1034,7 +1034,10 @@ export default defineComponent({
                   <v-card-title>
                     Submit
                   </v-card-title>
-                  <v-card-text>You are about to submit this study and metadata for NMDC review. Would you like to continue?</v-card-text>
+                  <v-card-text>
```
pkalita-lbl, line 1037: I think it would be better to completely swap out what we show in this card based on whether this submission is or is not a test. There is a bit of plumbing needed in order to do that, but I think it'll be worth it. The general outline would be like this:

Add a new `ref` in `web/src/views/SubmissionPortal/store/index.ts` to hold the `isTestSubmission` value. So something like 

```ts
const isTestSubmission = ref(false);
```

Then in the `reset`, `generateRecord`, and `loadRecord` functions you'd update `isTestSubmission.value` based on the incoming data (or to `false` in `reset`).

That new `ref` can then be exported by that module and imported in `HarmonizerView.vue`. Then you can use that value here to show the original message if it's not a test submission and the simple "Test submissions cannot be submitted" message otherwise.

pkalita-lbl (commented):
I tested these changes locally and everything pretty much works as expected. I just have a few small enhancements to consider.

web/src/views/SubmissionPortal/HarmonizerView.vue
JamesTessmer: That sounds good. I had initially looked at doing something like but wasn't sure where to set the value, but this makes sense.

nmdc_server/migrations/versions/4b2d6ee63752_add_isTestSubmission_to_SubmissionMetadata.py
```diff
@@ -0,0 +1,32 @@
...
+# revision identifiers, used by Alembic.
+revision: str = "4b2d6ee63752"
+down_revision: Optional[str] = "afa1ff687968"
+branch_labels: Optional[str] = None
+depends_on: Optional[str] = None
+
+
+def upgrade():
+    op.add_column(
+        "submission_metadata", sa.Column("isTestSubmission", sa.Boolean(), nullable=False)
```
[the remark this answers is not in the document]
JamesTessmer: Added

web/src/views/SubmissionPortal/Components/SubmissionList.vue
JamesTessmer: Added

pkalita-lbl (approved):
Looks good!

PR #1503, opened 2025-01-08
Use same object for tooltips and legend

### LINKED ISSUE #1501: Update key of upset plot for lipidomics
Current:
Screenshot 2024-12-19 at 11 53 35 AM

Desired:
Key should say "LIP: Lipidomics"

naglepuff:
Fix #1501 

### Changes

Previously, the legend for the upset plot used hard-coded string values to map abbreviations to data type. There is already an object being used as a dictionary to map abbreviations to data type. This PR removes the hard-coded `<span>`s in favor of using the existing object.

It also fixes the issue where the label on the upset plot for lipidomics didn't match the key in the legend.


marySalvi (approved):
LGTM

PR #1506, opened 2025-01-09
1281 update shipping info

### LINKED ISSUE #1281: Update shipping information for EMSL samples
When submitters come to NMDC, if they indicate
1. Have data already been generated for your study? * - NO
2. Are you submitting metadata for samples that will be sent to a DOE user facility? - EMSL
 
You're prompted to complete shipping information. For this feature updates I'd like:

- [ ] Add a "samples will not be shipped" box that makes all the extra shipping information not required
- [ ] For shipping info, only require country. 
- [ ] Add a "i" info hover / guidance text that says "provide the address of where the samples are being shipped FROM"

marySalvi:
Closes #1281 
Included in this PR

- An additional user input regarding if the samples will be shipped (EMSL)
- Adds the 'country' field to the shipping info
- Makes _only_ the country field required
- Adds a tooltip explaining that the sender's info is needed and updates help text

### DIFF
--- a/nmdc_server/migrations/versions/18127fd43a8b_add_country_to_shipper_info.py
+++ b/nmdc_server/migrations/versions/18127fd43a8b_add_country_to_shipper_info.py
@@ -1,7 +1,7 @@
 """add-country-to-shipper-info
 
 Revision ID: 18127fd43a8b
-Revises: afa1ff687968
+Revises: 4b2d6ee63752
 Create Date: 2025-01-24 15:53:45.897421
 
 """
@@ -21,7 +21,7 @@ Base = declarative_base()
 
 # revision identifiers, used by Alembic.
 revision: str = "18127fd43a8b"
-down_revision: Optional[str] = "afa1ff687968"
+down_revision: Optional[str] = "4b2d6ee63752"
 branch_labels: Optional[str] = None
 depends_on: Optional[str] = None
 
--- a/nmdc_server/migrations/versions/18127fd43a8b_add_country_to_shipper_info.py
+++ b/nmdc_server/migrations/versions/18127fd43a8b_add_country_to_shipper_info.py
@@ -1,0 +1,116 @@
+"""add-country-to-shipper-info
+
+Revision ID: 18127fd43a8b
+Revises: afa1ff687968
+Create Date: 2025-01-24 15:53:45.897421
+
+"""
+
+from datetime import datetime
+from typing import List, Optional
+from uuid import uuid4
+
+from alembic import op
+from pydantic import BaseModel
+from sqlalchemy import Column, orm
+from sqlalchemy.dialects.postgresql import JSONB, UUID
+from sqlalchemy.ext.declarative import declarative_base
+
+Base = declarative_base()
+
+
+# revision identifiers, used by Alembic.
+revision: str = "18127fd43a8b"
+down_revision: Optional[str] = "afa1ff687968"
+branch_labels: Optional[str] = None
+depends_on: Optional[str] = None
+
+
+# Redefine SubmissionMetadata, including only what we need for this migration
+# to ensure this migration is self-contained.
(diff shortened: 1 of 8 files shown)


nmdc_server/schemas_submission.py
```diff
@@ -47,6 +47,7 @@ class NmcdAddress(BaseModel):
     city: str
     state: str
     postalCode: str
+    country: str
```
naglepuff, line 50:
```suggestion
    country: str = ""
```
I came across a `500` status code when loading the submission landing page. Looks like there's some backend validation that is not happy.

Since `country` is a new required field, we need to stay backwards-compatible with existing submissions. There are a couple of options.

1. We could have the serializer add a default value (empty string?)
2. We could write a migration to add a default value

I don't mind sticking with option 1, but we also need to make sure that any consumers of this data are aware that `country` might be missing (I'm not sure how robust the submission to mongo pipeline is about this).

I also don't know what the default value should be. The empty string allows us to sneak existing submissions past the pydantic classes, but these submissions are still technically invalid. It's probably fine though.

web/src/views/SubmissionPortal/store/index.ts
```diff
@@ -110,6 +111,7 @@ const addressFormDefault = {
 const contextFormDefault = {
   dataGenerated: undefined as undefined | boolean,
   awardDois: [] as string[] | null,
+  ship: undefined as undefined | boolean,
```
naglepuff, line 114: I think this also needs to be added to the pydantic class in `schemas_submission.py`. Otherwise users will have to select an option for this each time they open the submission.

naglepuff (changes requested):
Front end validation looks good to me, but there were a couple of interactions with saving/loading that should be addressed.

nmdc_server/schemas_submission.py
marySalvi: fixed in d30368d5c437c4425ac72c382d057e9815fb4f22
As discussed with team, the default was set to `USA`

mslarae13:
@marySalvi still think this will be done today?

nmdc_server/migrations/versions/18127fd43a8b_add_country_to_shipper_info.py
```diff
@@ -0,0 +1,116 @@
+"""add-country-to-shipper-info
+
+Revision ID: 18127fd43a8b
+Revises: afa1ff687968
```
naglepuff, line 4:
```suggestion
Revises: 4b2d6ee63752
```

nmdc_server/migrations/versions/18127fd43a8b_add_country_to_shipper_info.py
```diff
@@ -0,0 +1,116 @@
...
+from sqlalchemy import Column, orm
+from sqlalchemy.dialects.postgresql import JSONB, UUID
+from sqlalchemy.ext.declarative import declarative_base
+
+Base = declarative_base()
+
+
+# revision identifiers, used by Alembic.
+revision: str = "18127fd43a8b"
+down_revision: Optional[str] = "afa1ff687968"
```
naglepuff, line 24:
```suggestion
down_revision: Optional[str] = "4b2d6ee63752"
```

naglepuff (changes requested):
The changes look good to me, and my original set of issues are resolved. I did notice that [this PR](https://github.com/microbiomedata/nmdc-server/pull/1495/files#diff-b01abb6de931d3068e0eda9a4d108091607deeaec8c29f45d3b574c18d69081f) introduced a new migration as well, and was merged into master recently, so I think we'll need to tweak the migration file slightly before merge.

marySalvi:
> @marySalvi still think this will be done today?

Yes

naglepuff:
@marySalvi I think you might also need to rebase off of main

naglepuff (approved):

Issue #1517, opened 2025-01-17
Add Search Facet for LC and MS Configurations


kheal:
Under the `Data Generation` facet on the left, add a search option for filtering samples by ChromatographyConfiguration (displayed as "Chromatography Method") and a second option for filtering samples by MassSpectrometryConfiguration (displayed as "Mass Spectrometry Method").  Use the name field from each of the records as the value in the search panel.  

These should have high counts (>1000) because they are applicable to NOM, Metabolomics, Lipidomics, and Proteomics analyses.

![Image](https://github.com/user-attachments/assets/23c7842c-3b7a-4d47-8bae-84606927d7dd)

PR #1525, opened 2025-01-24
Display LC/MS Configurations for Lipidomics and Metabolomics

### LINKED ISSUE #1514: Show LC/MS Configuration Info in Data Portal
Refer to the first several slides of https://docs.google.com/presentation/d/1ubYn0TMVNJ6VkG5UoQNns5NsJaqPUnro/edit#slide=id.p1 for more details.

naglepuff:
Fix #1514 

### Ingest
On ingest of `data_generation` records, there is now an additional check to see if slots `has_mass_spectrometry_configuration` and `has_chromatography_configuration`. In the case that one or both of those slots is populated, we'll do an additional lookup to the `configuration_set` collection in mongo, and store the `id` and `name` of the `configuration` document in the `annotations` column of the `omics_processing` table.

They are stored in the `annotations` blob as:
- `mass_spectrometry_configuration_name`
- `mass_spectrometry_configuration_id`
- `chromatography_configuration_id`
- `chromatography_configuration_name`

Storing these fields in the json blob has the advantage of not needing to update our database model. Our search already handles lookups into the `annotations` column (so we won't need to change how this is stored when it comes to implementing #1517), and the `annotations` blob is available on the client and can be put into the UI as needed.

### Data Object Table (Vue component)
There is some additional processing to make the new data available to the vue component. This data is disaplayed in a header just above relevant data objects.

image

#### Update 1/30/25
image

## Testing
To test, you'll need to run an ingest using the changes in this PR (from prod). The, use the biosample ID search to find a biosample such as `nmdc:bsm-13-2yga6857`, which has connections to `data_generation` records that point to LC/MS config records. View the data object tables for the relevant data types and verify the information displays as expected.

This PR does not add a facet to the search sidebar for LC/MS configs, but a side effect of the ingest changes is that one can use the API to filter on config names and IDs. For example, you can test out the `/api/data_generation/search` endpoint with the following request body:
```
{
  "conditions": [{
    "value": "nmdc:mscon-14-5jk7rg60",
    "field": "mass_spectrometry_configuration_id",
    "table": "omics_processing"
  }]
}
```

### DIFF
--- a/nmdc_server/ingest/omics_processing.py
+++ b/nmdc_server/ingest/omics_processing.py
@@ -96,7 +96,7 @@ def get_biosample_input_ids(input_id: str, mongodb: Database, results: set) -> s
     return results
 
 
-def load_omics_processing(db: Session, obj: Dict[str, Any], mongodb: Database, logger):
+def load_omics_processing(db: Session, obj: Dict[str, Any], mongodb: Database, logger, config_map):
     logger = get_logger(__name__)
     input_ids: list[str] = obj.pop("has_input", [""])
     biosample_input_ids: set[str] = set()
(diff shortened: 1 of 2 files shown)


naglepuff:
Waiting for #1532 to be merged to fix the failing `black` test.

kheal:
I'm realizing after seeing this example that the header "LC/MS Configurations" isn't quite appropriate for these types of data (GC/MS).  To make compliant with _all_ of the records of `type:nmdcMetabolomicsAnalysis`.  Can we update the display to remove the "Data Generation Activity" line and change the name on the remaining header to say "Data Generation Configurations" rather than "LC/MS Configurations".

Screen shot below of line and text to remove.  "LC/MS" should be replaced with "Data Generation Configurations".
Screenshot 2025-01-30 at 12 40 44 PM

Otherwise this looks great.

naglepuff:
> I'm realizing after seeing this example that the header "LC/MS Configurations" isn't quite appropriate for these types of data (GC/MS). To make compliant with _all_ of the records of `type:nmdcMetabolomicsAnalysis` (which includes lipidomics), can we update the display to remove the "Data Generation Activity" line and change the name on the remaining header to say "Data Generation Configurations" rather than "LC/MS Configurations".
> 
> Otherwise this looks great.

Will do! Thanks for the feedback.

marySalvi (approved):
LGTM

PR #1533, opened 2025-02-03
Add date_last_modified and supporting changes

JamesTessmer:
Adds `date_last_modified` to the `MetadataSubmission` schema. 
This field is displayed on the submission portal, and is reported via API endpoint for review.

Currently on the submission portal `date_last_modified` replaces `created` in the view, but `created` can be added again if needed.

Relevant issue: https://github.com/microbiomedata/issues/issues/828


web/src/views/SubmissionPortal/Components/SubmissionList.vue
```diff
@@ -35,8 +35,8 @@ const headers: DataTableHeader[] = [
     value: 'status',
   },
   {
-    text: 'Created',
-    value: 'created',
+    text: 'Date Last Modified',
+    value: 'date_last_modified',
```
pkalita-lbl, line 39: The default sort for the table is specified when the `options` object is defined in `setup`. We'll want to change that from `created` to `date_last_modified` as well if we change the column.

web/src/views/SubmissionPortal/Components/SubmissionList.vue
```diff
@@ -226,8 +226,8 @@ export default defineComponent({
           <template #[`item.templates`]="{ item }">
             {{ item.metadata_submission.templates.map((template) => HARMONIZER_TEMPLATES[template].displayName).join(' + ') }}
           </template>
-          <template #[`item.created`]="{ item }">
-            {{ new Date(item.created).toLocaleString() }}
+          <template #[`item.date_last_modified`]="{ item }">
+            {{ new Date(item.date_last_modified).toLocaleString() }}
```
pkalita-lbl, line 230: I can't believe I've missed this bug before. We store all of our datetime as UTC in the database (which is good). But when they're serialized back out to JSON via Pydantic, it doesn't know what timezone they're in, so they come out as timezone-less strings like `"2024-12-17T01:20:58.801132"`. That means the we end up displaying the UTC time on the frontend.

You can see this now on dev and prod. If you create a new submission and then immediately go back to the submission portal home page you'll see the Created column for you new study reads 8 hours from now (assuming you're in Pacific Standard Time). 

The most comprehensive fix would probably be to change our Pydantic models so they serialize datetimes as either UNIX timestamps or strings with timezone info so it's less ambiguous. Alternatively, a short-term fix would be to append the letter `Z` (indicating UTC) to the `date_last_modified` value as it's passed to the `Date` constructor.
```suggestion
            {{ new Date(item.date_last_modified + 'Z').toLocaleString() }}
```

web/src/views/SubmissionPortal/Components/SubmissionList.vue
JamesTessmer: Ah I missed that, good catch thanks.

nmdc_server/migrations/versions/6e6dec0daa7c_add_date_last_modified.py
```diff
@@ -0,0 +1,34 @@
...
+down_revision: Optional[str] = "18127fd43a8b"
+branch_labels: Optional[str] = None
+depends_on: Optional[str] = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column(
+        "submission_metadata", sa.Column("date_last_modified", sa.DateTime(), nullable=False)
+    )
```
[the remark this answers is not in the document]
JamesTessmer: That makes sense, I'll add that.

web/src/views/SubmissionPortal/Components/SubmissionList.vue
JamesTessmer: I had noticed that the time zone was off when doing testing for this, but I had just assumed that was just something funky due to running a local server on Docker, but didn't notice it on prod! I'll add the short term fix in for now

pkalita-lbl (approved):

PR #1535, opened 2025-02-04
Add metadata suggestion UI

### LINKED ISSUE #1383: Implement metadata suggestion/enrichment UI
Part of Milestone 3.3 (https://github.com/microbiomedata/issues/issues/473)

This task is to implement the UI that was mocked up and approved in https://github.com/microbiomedata/nmdc-server/issues/1381. It should utilize the endpoint implemented in https://github.com/microbiomedata/nmdc-server/issues/1382. Refinements to the API endpoint based on the needs of the UI can also be considered here.

This task is done when the UI has been implemented and tested in the dev environment.

pkalita-lbl:
Fixes #1383 

Part of Milestone 3.3 (https://github.com/microbiomedata/issues/issues/473)

### Summary

The primary goal of these changes is to add functionality to the Submission Portal whereby a user can receive metadata suggestions based on what metadata has been entered so far. The UI talks to a endpoint previously implemented in #1403  (with a few minor updates here). Once fetched, the UI manages the display of the suggestions and the interactions to accept or reject suggestions (either one-by-one or as a batch).

The new UI has been integrated into the sidebar to the right of the DataHarmonizer grid, which previously displayed column help info and find/replace tools. The sidebar is now tabbed to accommodate the expanded set of tools it hosts. In order to simplify the overall Submission Portal UI, the import/export buttons and site-wide help box have also been moved into tabs of the sidebar.

### Comment about local storage

Information about pending (i.e. not yet accepted or rejected) and rejected suggestions is persisted to local storage so that it persists across DataHarmonizer tab changes and page reloads. The reason for using local storage (as opposed to pushing the information back to the server) was that it was a relatively straightforward way to keep those details user-specific. That is, if two users have access to the same submission, one user's rejection of a suggestion shouldn't prevent that suggestion from being displayed to the other user. In retrospect, it might not have been much more work to introduce a new Postgres table with a `user_id` column to keep track of that information. If anyone has strong feeling about that I'm willing to reconsider the current design.

### Details

* The backend changes in `nmdc_server/metadata.py` and `nmdc_server/api.py` do the following:
  * Upgrade to `nmdc-geoloc-tools` version `0.2.0` which eliminates the `GeoEngine` class in favor of plain functions. That version also adds a bit of caching so it doesn't always make requests to ORNL services when it doesn't need to.
  * Return the current value along with the suggested value in the suggestion response.
* The tabbed sidebar is implemented in `HarmonizerSidebar.vue`. This component is really only responsible for rendering the tabbed layout. The content of individual tabs is delegated to separate components:
  * `ColumnHelp.vue` shows the column help for the given column. This was previously implemented directly in `HarmonizerView.vue`. These changes just lift that into a new component and make some of the styling more consistent with other sidebar tab components.
  * `FindReplace.vue` was already its own component. It just needed a bit of styling adjustment to make it consistent with other sidebar tab components. Also toggling the "Replace" field visibility was removed as it is no longer needed in this new layout.
  * `ImportExportButtons.vue` is a new component to show the Import and Export buttons that were removed from `HarmonizerView.vue`.
  * `ContactCard.vue` is an existing component. It just got some minor styling adjustments to make it consistent with other sidebar tab components. It is also displayed on the Submission Portal home page, and I tried to ensure it still looks good there too.
  * `MetadataSuggester.vue` is the main new component. The main logic is that it receives pending suggestions from the Submission Portal store (`metadataSuggestions`). It also loads previously rejected suggestions from local storage. It uses them to filter the pending suggestions and then group the remaining ones by row (`suggestionsByRow`). That is what is displayed to the user. When a user accepts suggestions, it is responsible for using a `HarmonizerApi` instance to update the data grid and remove the suggestions from the pending list. When a user rejects suggestions, it is responsible for adding those suggestions to the rejected list in local storage. Finally it is also responsible for calling the suggestion endpoint and adding the results to the pending list (via `addMetadataSuggestions`) in "on demand" mode.
* Aside from moving some content into separate components, the main changes to `HarmonizerView.vue` are:
  * When data in DataHarmonizer changes (`onDataChange`) and we are in "live" suggestion mode, add those changes to a batch that will be sent to the suggestion endpoint. Once there have been no changes for 3 seconds (a somewhat arbitrary debouncing timeout), actually make the API request and add the results to the pending suggestions list (`addMetadataSuggestions`).
  * When the component is first mounted (`onMounted`) or when the template tab changes (`changeTemplate`), load any pending suggestions previously saved to local storage.
* The `HarmonizerApi` class got a new methods to extract data from DataHarmonizer only for specific rows (`getDataByRows`) and to get which cells are currently selected (`getSelectedCells`). It also has new fields (`slotNames`, `slotInfo`) which aid in translating between slot names and column numbers (and vice-versa) and slot titles.
* `api.ts` has a new method for calling the suggestion endpoint
* `views/SubmissionPortal/store/index.ts` has new refs for keeping track of pending suggestions (`metadataSuggestions`) and state tied to UI controls for suggestion mode and suggestion type. The `addMetadataSuggestions` function is responsible for calling the suggestion endpoint (in batches in case many rows were updated at once), adding the results to the `metadataSuggestions` state, and persisting the suggestions to local storage. The `removeMetadataSuggestions` function is responsible for removing items from `metadataSuggestions` and persisting the result to local storage.
* A lot of type definitions were moved into a separate `types.ts` module. The reason is that many of these types are needed by several different modules (e.g. `api.ts` and `store/index.ts` and `harmonizerApi.ts`), and I was constantly running into circular import issues.


web/src/views/SubmissionPortal/Components/HarmonizerSidebar.vue
```diff
@@ -0,0 +1,162 @@
...
+      required: true,
+    },
+  },
+  emits: ['export-xlsx', 'import-xlsx'],
+  setup(props, { emit }) {
+    const tabModel = ref(0);
+    const TABS = [
+      {
+        icon: 'mdi-information-outline',
+        label: 'Column Help',
```
naglepuff, line 60:
```suggestion
        icon: 'mdi-information-outline',
        label: 'Column Info',
```
I think having "Column Help" and "Help" could potentially be confusing. This label matches the icon better and I think still fits what this column has to offer.

web/src/views/SubmissionPortal/Components/HarmonizerSidebar.vue
```diff
@@ -0,0 +1,162 @@
...
+        label: 'Find & Replace',
+      },
+      {
+        icon: 'mdi-assistant',
+        label: 'Metadata Suggester',
+      },
+      {
+        icon: 'mdi-swap-vertical',
+        label: 'Import & Export',
+      },
```
naglepuff, line 73: Do you have any thoughts on making this the default column? I imagine that for a new submission, this would be the first thing a user would want to do (download the template).

Subsequent visits might be worse though, if the data is already imported and the remaining work is small edits.

No action requested here, just a thought I had.

web/src/views/SubmissionPortal/Components/HarmonizerSidebar.vue
```diff
@@ -0,0 +1,162 @@
...
+
+<style lang="scss" scoped>
+.harmonizer-sidebar-content {
+  font-size: 14px;
+
+  .v-tab {
+    min-width: 10px;
+  }
+}
+</style>
```
naglepuff, line 162: What about adding a border and/or some height/shadow to the sidebar to make it feel more separate from the data harmonizer?

web/src/views/SubmissionPortal/Components/HarmonizerSidebar.vue
```diff
@@ -0,0 +1,162 @@
...
+  },
+});
+</script>
+
+<template>
+  <div class="harmonizer-sidebar-content">
+    <v-tabs
+      v-model="tabModel"
+      grow
+    >
```
naglepuff, line 98: I don't think this would show up for most people's setups, but when I had the dev console open I noticed that I could scroll the tabs out of frame. Do you think it might be worth pinning them?

image

web/src/views/SubmissionPortal/Components/MetadataSuggester.vue
```diff
@@ -0,0 +1,500 @@
...
+        <v-col>
+          <v-btn
+            class="text-no-transform"
+            color="grey darken-2"
+            block
+            plain
+            @click="handleResetRejectedSuggestions"
+          >
+            Reset rejected suggestions
+          </v-btn>
```
naglepuff, line 467: Is there a reason that this button is `plain`? I would have expected to match the style of existing buttons on the page.

naglepuff (commented):
This seems to be working well! I had a bunch of comments, but mostly style questions/suggestions. Let me know what you think.

tests/conftest.py
```diff
@@ -22,21 +22,21 @@ def set_seed(connection):
 def patch_geo_engine(monkeypatch):
     """Patch all the GeoEngine methods that make external network requests."""
```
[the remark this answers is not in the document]
pkalita-lbl: Fixed!

web/src/views/SubmissionPortal/Components/FindReplace.vue
```diff
@@ -130,113 +127,103 @@ export default defineComponent({
...
+            </template>
+            <span>Replace all</span>
+          </v-tooltip>
+        </v-col>
+      </v-row>
+    </v-card-text>
+  </v-card>
 </template>
 
 <style lang="scss">
```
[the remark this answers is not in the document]
pkalita-lbl: Removed!

web/src/views/SubmissionPortal/HarmonizerView.vue
```diff
@@ -180,13 +185,42 @@ export default defineComponent({
...
+    const onDataChange = async (changes: any[]) => {
+      // If we're in live suggestion mode and the user can edit the metadata, add the changes to a batch. Once the user
+      // has not made further changes for a certain amount of time, send the batch to the backend for suggestions.
+      if (suggestionMode.value === SuggestionsMode.LIVE && canEditSampleMetadata()) {
+        // Many "empty" changes can be fired when clearing an entire row or column. We only care about the ones
+        // where either the previous value or updated value (or both) are non-empty.
+        const nonEmptyChanges = changes.filter((change) => isNonEmpty(change[2]) || isNonEmpty(change[3]));
+        changeBatch.push(...nonEmptyChanges);
+        clearTimeout(suggestionRequestTimer);
+        suggestionRequestTimer = setTimeout(async () => {
```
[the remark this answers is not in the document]
pkalita-lbl: I think I was mainly concerned about the ability to cancel a pending call if the suggestion mode changes. Now I see that the wrapped function returned by `lodash.debounce` includes a `cancel` method, so I've switch over to use that. Definitely better to not have to manually manage timer IDs.

web/src/views/SubmissionPortal/Components/HarmonizerSidebar.vue
pkalita-lbl: I updated the tooltip and title within the sidebar to "Column Info".

web/src/views/SubmissionPortal/Components/HarmonizerSidebar.vue
pkalita-lbl: TBH I'm not sure how often users use the import/export functionality. I guess that's a user research question. I'll ask Montana about it as a follow-on task.

web/src/views/SubmissionPortal/Components/HarmonizerSidebar.vue
pkalita-lbl: I did this via adding a few `v-navigation-drawer` props:

* `temporary` -- this adds a box shadow to the drawer element
* `hide-overlay` -- we don't want to dim the rest of the page content; we just want the shadow
* `stateless` -- probably should have been using this previously since we fully manage the open/closed state and we don't need mobile support

web/src/views/SubmissionPortal/Components/HarmonizerSidebar.vue
pkalita-lbl: That's a good observation. Even without the dev console you can get into this state with some of the columns (like "broad-scale environmental context") that have a long description or in the suggester tab if you have a lot of pending suggestions.

I added some flex utility classes to ensure that the tab content is what scrolls.

web/src/views/SubmissionPortal/Components/MetadataSuggester.vue
pkalita-lbl: Yeah I don't know 😂 Just trying to lower the visual priority of it, but ended up overcooking it. I've updated it so that it uses the same style as the "full reference" button in the Column Info tab.

pkalita-lbl:
Thanks for the feedback @naglepuff! I think I've addressed everything you mentioned. Please let me know if you see anything else that needs another look.

web/src/views/SubmissionPortal/Components/HarmonizerSidebar.vue
naglepuff: I'm still seeing this after pulling the most recent changes. I messed around a bit and it looks like [position: sticky](https://developer.mozilla.org/en-US/docs/Web/CSS/position?v=example#sticky) might do the trick. If you want to try the flexbox solution as well that's fine. Either way I don't think this should hold up the merge, we can spin this out into an issue. 

Let's get this into dev for more thorough testing.

naglepuff (approved):
I added another comment about the scroll issue, but I'll leave it up to you to resolve however you see fit, and merge when ready.

web/src/views/SubmissionPortal/Components/HarmonizerSidebar.vue
pkalita-lbl: Yeah agreed on moving forward. Maybe you'll have to show me how you're reproducing the issue; I might have been doing something slightly different.

Issue #1536, opened 2025-02-04
Cannot bulk download files whose data objects lack size information


aclum:
I can bulk download data from EMSL
I can download individual files from NERSC
I can not bulk download files from NERSC

ex 
https://data.microbiomedata.org/api/bulk_download/52add6e2-4f08-41e4-917b-838d359ba5a6

```json
{
"detail": "Bulk download expired"
}
```

I suspect an issue with the nersc globus endpoint based on other tests that worked so starting off with Eric as the assignee. 

cc @naglepuff @marySalvi @shreddd 

This was originally reported to @ljohnson09  by a user and I was able to reproduce it by trying to bulk download a subset of the annotation files ([filter](https://data.microbiomedata.org/?q=CmwIABAJGAQiZCJVU0E6IE1pbm5lc290YSwgTWFyY2VsIEV4cGVyaW1lbnRhbCBGb3Jlc3QsIFNwcnVjZSBhbmQgUGVhdGxhbmQgUmVzcG9uc2VzIHRvIEVudmlyb25tZW50YWwgQ2hhbmdlcyI)).

ljohnson09:
Here is the example from our user! 

![Image](https://github.com/user-attachments/assets/4075330c-106e-4b24-b052-7db33287b0d5)

ljohnson09:
![Image](https://github.com/user-attachments/assets/a876cf6a-9412-4d74-802b-84bf2dc64a5a)

And the filtering the user applied - He said soil but the ecosystem is showing "Unclassified" here

eecavanna:
Thanks, @ljohnson09! 

For future reference, here's a link to the Data Portal search page with those "Study ID," "Ecosystem subtype," and "Data Type" filters applied: [link](https://data.microbiomedata.org/?q=ChQIABACGAIiDCJNZXRhZ2Vub21lIgoWCAAQEBgEIg4iVW5jbGFzc2lmaWVkIgocEAAYAyIWIm5tZGM6c3R5LTExLTM0eGoxMTUwIgocEAAYAyIWIm5tZGM6c3R5LTExLTU0N3J3cTk0IgocEAAYAyIWIm5tZGM6c3R5LTExLTI4dG01ZDM2Ig)

Next, I'll select the same download items and try to download them.

eecavanna:
I get the originally-reported error when I click "Download ZIP" and accept the terms in the popup.

```json
{
  "detail": "Bulk download expired"
}
```

[...]

PR #1540, opened 2025-02-11
Hotfix: Don't include invalid data objects in mod_zip file list

naglepuff:
### Previously
If a data object selected for bulk download had a file size of `NULL` or `0` (both perfectly valid NMDC data objects according to the [schema](https://microbiomedata.github.io/nmdc-schema/DataObject/)), we would include it in the file list for [mod_zip](https://github.com/evanmiller/mod_zip) like so:
```
-  data/object/url/...
```
This is not valid, per, mod_zip's documentation, and was preventing people from executing bulk downloads including these files.

### After this change
Construction of the file list for mod_zip has been changed. Files whose `file_size_bytes` is `NULL` or `0` are no longer included. Additionally, we log a warning.

## Hotfix

This is a hotfix. This branch was created from v1.3.0 (the January 2025 release). Normal hotfix procedure includes issuing a new tag to use for the production deployment at the same time this gets merged into development. Since this is hard to test locally, we could modify that procedure slightly by:

1. Merging into `main`, and testing on dev
2. Once testing is satisfactory, restore this branch and create tag v1.3.1

### DIFF
--- a/nmdc_server/crud.py
+++ b/nmdc_server/crud.py
@@ -475,14 +475,6 @@ def get_zip_download(db: Session, id: UUID) -> Optional[str]:
             logger.warning(f"Data object url for {file.path} was {data_object.url}")
             continue
 
-        # Skip data objects whose `file_size_bytes` is `0` or `None`, since `mod_zip`
-        # considers a file list containing any file sizes of "0" or "" to be invalid.
-        if not data_object.file_size_bytes:
-            logger.warning(
-                f"Data object file_size_bytes for {file.path} was {data_object.file_size_bytes}"
-            )
-            continue
-
         # TODO: add crc checksums to support retries
         # TODO: add directory structure and metadata
         content.append(f"- {data_object.file_size_bytes} {url} {file.path}")
--- a/nmdc_server/crud.py
+++ b/nmdc_server/crud.py
@@ -477,7 +477,7 @@ def get_zip_download(db: Session, id: UUID) -> Optional[str]:
 
         if not data_object.file_size_bytes:
             logger.warning(
-                f"Data object file_size_bytes for {file.path} was {data_object.file_sizey_bytes}"
+                f"Data object file_size_bytes for {file.path} was {data_object.file_size_bytes}"
             )
             continue
 
--- a/nmdc_server/crud.py
+++ b/nmdc_server/crud.py
@@ -475,6 +475,8 @@ def get_zip_download(db: Session, id: UUID) -> Optional[str]:
             logger.warning(f"Data object url for {file.path} was {data_object.url}")
             continue
 
+        # Skip data objects whose `file_size_bytes` is `0` or `None`, since `mod_zip`
+        # considers a file list containing any file sizes of "0" or "" to be invalid.
         if not data_object.file_size_bytes:
             logger.warning(
                 f"Data object file_size_bytes for {file.path} was {data_object.file_size_bytes}"
--- a/nmdc_server/crud.py
+++ b/nmdc_server/crud.py
@@ -475,10 +475,15 @@ def get_zip_download(db: Session, id: UUID) -> Optional[str]:
             logger.warning(f"Data object url for {file.path} was {data_object.url}")
             continue
 
+        if not data_object.file_size_bytes:
+            logger.warning(
+                f"Data object file_size_bytes for {file.path} was {data_object.file_sizey_bytes}"
+            )
+            continue
+
         # TODO: add crc checksums to support retries
         # TODO: add directory structure and metadata
-        file_size_string = data_object.file_size_bytes if data_object.file_size_bytes else ""
-        content.append(f"- {file_size_string} {url} {file.path}")
+        content.append(f"- {data_object.file_size_bytes} {url} {file.path}")
 
     bulk_download.expired = True
     db.commit()
--- a/nmdc_server/ingest/data_object.py
+++ b/nmdc_server/ingest/data_object.py
@@ -52,6 +52,12 @@ def load(db: Session, cursor: Cursor, file_types: List[Dict[str, Any]]):
         else:
             objects_without_type += 1
 
+        if obj.get("file_size_bytes", None) is None:
+            logger.warning(
+                f"data_object {obj['id']} has file_size_bytes {obj.get('file_size_bytes')} "
+                "and cannot be included in bulk downloads"
+            )
+
         db.add(DataObject(**obj))
 
     if objects_without_type:
--- a/nmdc_server/query.py
+++ b/nmdc_server/query.py
@@ -926,7 +926,12 @@ class DataObjectQuerySchema(BaseQuerySchema):
 
     def aggregate(self, db: Session) -> DataObjectAggregation:
         """Return the number of files and total size of matched data objects."""
-        subquery = self.query(db).filter(models.DataObject.url != None).subquery()
+        subquery = (
+            self.query(db)
+            .filter(models.DataObject.url != None)
+            .filter(models.DataObject.file_size_bytes != None)
+            .subquery()
+        )
         row = db.query(
             func.count(subquery.c.id),
             func.sum(func.coalesce(subquery.c.file_size_bytes, 0)),


nmdc_server/crud.py
```diff
@@ -475,10 +475,15 @@ def get_zip_download(db: Session, id: UUID) -> Optional[str]:
             logger.warning(f"Data object url for {file.path} was {data_object.url}")
             continue
 
+        if not data_object.file_size_bytes:
+            logger.warning(
+                f"Data object file_size_bytes for {file.path} was {data_object.file_sizey_bytes}"
```
eecavanna, line 480: I think there was a typo here:

```suggestion
                f"Data object file_size_bytes for {file.path} was {data_object.file_size_bytes}"
```

nmdc_server/crud.py
```diff
@@ -475,10 +475,15 @@ def get_zip_download(db: Session, id: UUID) -> Optional[str]:
             logger.warning(f"Data object url for {file.path} was {data_object.url}")
             continue
 
+        if not data_object.file_size_bytes:
```
eecavanna, line 478: I suggest documenting our rationale here.

```suggestion
        # Skip data objects whose `file_size_bytes` is `0` or `None`, since `mod_zip`
        # considers a file list containing any file sizes of "0" or "" to be invalid.
        if not data_object.file_size_bytes:
```

eecavanna:
Thanks for drafting this!

Previously (before this hotfix), the download would completely fail and they'd get 0 files, so I do think this is an improvement in terms of how many files they can access.

My only concern is that the bulk downloader UI will still be telling people (effectively) "You will be downloading `N` files," but the ZIP file they end up with will only contain, for example, `N`-`M` files (where `M` is the number that got filtered out by this hotfix). I'm uncomfortable with giving contradictory information to the user.

I am wondering what the status of the upstream fix is.

eecavanna:
I'll tag the product owner (@aclum) as I am interested in her opinion about that implication:

> the bulk downloader UI will still be telling people (effectively) "You will be downloading N files," but the ZIP file they end up with will only contain, for example, N-M files (where M is the number that got filtered out by this hotfix).

naglepuff:
> the bulk downloader UI will still be telling people (effectively) "You will be downloading `N` files," but the ZIP file they end up with will only contain, for example, `N`-`M` files (where `M` is the number that got filtered out by this hotfix).

That's valid. Some alternate solutions: 
1. Ignore those data objects on ingest. (I don't think that's a good solution either.I don't think individual file download is broken - even for the problematic data objects).
2. Update the logic that builds the bulk download summary to not include problematic files. That logic is [here](https://github.com/microbiomedata/nmdc-server/blob/d074e2800857101c0cf76850d218b1394d5305e3/nmdc_server/query.py#L927-L936). In this case we'd expect the UI to match what ultimately gets downloaded.

I think we'd still want to log _somewhere_ that certain data objects would be excluded from bulk downloads, but we could do that during ingest instead of in the request thread.

eecavanna:
Thanks for presenting those solutions. I like the 2nd one—it would resolve my concern. 

In my mind, that still leaves a gap between "what the user can download one-by-one" and "what the user can download in bulk," but, given that the bulk downloader doesn't explicitly say that it is bundling up _all_ files that could be downloaded one-by-one, I'd be comfortable proceeding with "solution 2" and filing a ticket about that gap.

I agree about still logging that certain data objects would be excluded. Doing it during ingest—where those data objects are being "ingested" from the Mongo database—makes sense to me.

shreddd:
Are we sure 0 doesn't work as a file size in mod zip? I couldn't find anything in the documentation about this - @naglepuff can you point me to a reference. Note that the issue with #1536 was around missing file size information (not 0 bytes for empty files).

naglepuff:
@eecavanna PTAL at the recent changes and let me know your thoughts. When I outlined this approach at the sync today I didn't hear any dissent, but @aclum and @shreddd if you have concerns please let me know.

eecavanna:
I'll review them now (TIL: "PTAL"!)

nmdc_server/query.py
```diff
@@ -926,7 +926,12 @@ def _data_object_filter_subquery(
 
     def aggregate(self, db: Session) -> DataObjectAggregation:
         """Return the number of files and total size of matched data objects."""
-        subquery = self.query(db).filter(models.DataObject.url != None).subquery()
+        subquery = (
+            self.query(db)
+            .filter(models.DataObject.url != None)
+            .filter(models.DataObject.file_size_bytes != None)
```
eecavanna, line 932: Note to self: In this subquery, we are filtering out `DataObject`s whose `file_size_bytes` column contains `NULL` (as opposed to `0`).

nmdc_server/crud.py
```diff
@@ -477,8 +477,7 @@ def get_zip_download(db: Session, id: UUID) -> Optional[str]:
 
         # TODO: add crc checksums to support retries
         # TODO: add directory structure and metadata
-        file_size_string = data_object.file_size_bytes if data_object.file_size_bytes else ""
-        content.append(f"- {file_size_string} {url} {file.path}")
+        content.append(f"- {data_object.file_size_bytes} {url} {file.path}")
```
eecavanna, line 480: Note to self:
- `DataObject`s whose `file_size_bytes` value is `None` will have been filtered out during ingest
- `DataObject`s` whose `file_size_bytes` value is `0` will be included here, and the line in the file list will have a `0` (as opposed to an empty string) in the size spot

nmdc_server/ingest/data_object.py
```diff
@@ -52,6 +52,12 @@ def load(db: Session, cursor: Cursor, file_types: List[Dict[str, Any]]):
         else:
             objects_without_type += 1
 
+        if obj.get("file_size_bytes", None) is None:
+            logger.warning(
+                f"data_object {obj['id']} has file_size_bytes {obj.get('file_size_bytes')} "
+                "and cannot be included in bulk downloads"
```
eecavanna, line 58: Not a deal breaker—a thought while reading this.

I recommend changing this to `will not` to emphasize that we made the decision to omit them, as opposed to us making the decision to implementing some kind of workaround that allows us to include them.

```suggestion
                "and will not be included in bulk downloads"
```

All changes in this PR branch look good to me and make sense to me. I am going to catch up on the latest comment(s) in the related issue now.

eecavanna:
Assuming bulk downloads — where both (a) the set of files being downloaded includes 0-byte files and (b) the file list fed into `mod_zip` includes some lines where the file size is represented with a `0` — do work, I'm OK with this being merged in. It matches the "alternative solution" number 2, [listed above](https://github.com/microbiomedata/nmdc-server/pull/1540#issuecomment-2652063126), which I was/am in favor of.

nmdc_server/ingest/data_object.py
eecavanna: I'm going to convert this suggestion into a regular comment so it doesn't hold up the merge.

naglepuff:
@eecavanna based on the status of your last three comments, I believe you are done with review. In that case, do you have any objects to me:

1. Merging this PR into `main`
2. Restoring this branch (it will be deleted automatically upon merge) in order to facilitate a "hotfix" based on the head of this branch
3. Testing in dev
4. Issuing a hotfix once this change is confirmed to have some positive effect in dev

I understand this breaks a little from the established hotfix protocol, but I'd like to test this in dev today, just in case there's something being missed. This is especially relevant here because bulk download is so hard to test locally.

eecavanna:
> based on the status of your last three comments, I believe you are done with review

Correct. Sorry I didn't say that explicitly. I am done with review. I approve of this being merged into `main` and I'm OK with you taking the four steps you listed.

Here are my comments (thoughts) on those four steps (using the same numbers here):
1. This will make it so (a) it gets deployed to dev immediately, and (b) it gets deployed to production during the next NMDC Release (on February 24)
2. (no comment)
3. (no comment)
4. Sounds good. If, in dev, it proves to be a **regression** compared to today's behavior, we can make additional changes on `main` to (effectively) revert it.

In short (in my own words), you want it to be on `main` so you can test it in dev, since testing it locally is difficult. Once you test it in dev, you'll issue a hotfix to get it into prod.

eecavanna:
On a side note, I would like to rig up an isolated "nginx with `mod_zip` + some files in a container" to facilitate testing `mod_zip`-related things locally.

eecavanna:
I forgot to CC @pkalita-lbl on my [comment above](https://github.com/microbiomedata/nmdc-server/pull/1540#issuecomment-2660281362), as I think he is the member of our team that is most familiar with the release process.

eecavanna:
Reminder: 👆 We are intentionally **not deleting** this branch yet. See comments above for details.

PR #1541, opened 2025-02-12
Update API endpoint to support date_last_modified

JamesTessmer:
Updating the `get_metadata_submissions_report` endpoint to include `is_test_submission` and `date_last_modified`

### DIFF
--- a/tests/test_submission.py
+++ b/tests/test_submission.py
@@ -142,6 +142,8 @@ def test_get_metadata_submissions_report_as_admin(
         author=logged_in_user,
         author_orcid=logged_in_user.orcid,
         created=now,
+        date_last_modified=datetime.utcnow(),
+        is_test_submission=False,
     )
     fakes.SubmissionRoleFactory(
         submission=submission,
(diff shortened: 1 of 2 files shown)


eecavanna:
I'll review this now.

tests/test_submission.py
```diff
@@ -209,6 +216,7 @@ def test_get_metadata_submissions_report_as_admin(
     assert data_row["PI Email"] == ""
     assert data_row["Source Client"] == ""  # upstream faker lacks `source_client` attribute
     assert data_row["Status"] == "In Progress"  # matches value in upstream faker
+    assert data_row["Is Test Submission"] == "False"
```
eecavanna, line 219: Hi @JamesTessmer, I recommend testing the case where this not "False". You can do that by replacing `False` with `True` when creating the `other_submission` (currently on line 160), and then updating this assertion accordingly.

Also, I recommend asserting something about the `"Date Last Modified"` value. For example, you can assert that its value is a string (as opposed to, for example `None`); like this:

```py
assert isinstance(data_row["Date Last Modified"], str)
```

eecavanna (changes requested):
Hi @JamesTessmer, thanks for updating this test. The changes are looking good to me. There are two additional changes that I recommend be made, which I've written in an single comment. Once those changes are in place, I'll approve/merge. I can also discuss them, in case you want to.

eecavanna:
Thanks, @JamesTessmer. I'll take a look now.

eecavanna (approved):
Looks good to me! Thanks for making those changes. Looks like I mixed up the two test submissions in my original PR feedback (oops!). I'll approve and merge.

JamesTessmer:
Thanks Eric!

PR #1556, opened 2025-02-26
Add facets for mass spec and chromatography configs

### LINKED ISSUE #1517: Add Search Facet for LC and MS Configurations

naglepuff:
Fix #1517 

### Changes

Adds UI for filtering by Mass Spec configuration and Chromatography configuration. The new facets are grouped under "Data Generation" like so:
image

Each of these facets allows users to select from a list of configuration names:
image

image

Note that these screenshots are from my local development environment, so any counts could be out of sync with the data currently available in mongo.

@kheal adding you as a reviewer to catch anything that might be off in the screenshots before this gets merged into dev for testing.

### DIFF
--- a/web/src/encoding.ts
+++ b/web/src/encoding.ts
@@ -604,7 +604,16 @@ const tableFields: Record<entityType, Record<string, FieldsData>> = {
   },
   biosample: {},
   study: {},
-  omics_processing: {},
+  omics_processing: {
+    mass_spectrometry_configuration_name: {
+      name: 'Mass Spectrometry Method',
+      schemaName: 'MassSpectrometryConfiguration',
+    },
+    chromatography_configuration_name: {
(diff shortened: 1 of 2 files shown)


kheal:
Facets look awesome and counts look in the right ballpark.

kheal (approved):

marySalvi (approved):
LGTM

PR #1557, opened 2025-02-28
Add ability to filter submissions based on submission type

JamesTessmer:
This PR will update the `GET api/metadata_submission` endpoint to support filtering and add front end changes to allow users to filter submissions on the portal.

### DIFF
--- a/nmdc_server/api.py
+++ b/nmdc_server/api.py
@@ -869,22 +869,13 @@ async def list_submissions(
     pagination: Pagination = Depends(),
     column_sort: str = "created",
     sort_order: str = "desc",
-    filter_column: Optional[str] = None,
-    filter_value: Optional[str] = None,
+    is_test_submission_filter: Optional[bool] = None,
 ):
-    query = crud.get_submissions_for_user(db, user, column_sort, sort_order)
-    result = pagination.response(query)
-    if filter_column and filter_value:
-        filtered_result = []
-        tables = result["results"]
-        if filter_column in tables[0].__table__.columns.keys():  # type: ignore
-            for table in tables:
-                if filter_value.lower() in (str(getattr(table, filter_column))).lower():
-                    filtered_result.append(table)
-            result["results"] = filtered_result
-            result["count"] = len(filtered_result)
-
-    return result
+    query = crud.get_submissions_for_user(
+        db, user, column_sort, sort_order, is_test_submission_filter
+    )
+    # print(filters)
+    return pagination.response(query)
 
 
 @router.get(
--- a/nmdc_server/api.py
+++ b/nmdc_server/api.py
@@ -877,8 +877,6 @@ async def list_submissions(
     if filter_column and filter_value:
         filtered_result = []
         tables = result["results"]
-        print(tables[0])
-        print(type(tables[0]))
         if filter_column in tables[0].__table__.columns.keys():  # type: ignore
             for table in tables:
                 if filter_value.lower() in (str(getattr(table, filter_column))).lower():
--- a/nmdc_server/api.py
+++ b/nmdc_server/api.py
@@ -874,7 +874,6 @@ async def list_submissions(
     query = crud.get_submissions_for_user(
         db, user, column_sort, sort_order, is_test_submission_filter
     )
-    # print(filters)
     return pagination.response(query)
 
 
--- a/nmdc_server/api.py
+++ b/nmdc_server/api.py
@@ -869,9 +869,24 @@ async def list_submissions(
     pagination: Pagination = Depends(),
     column_sort: str = "created",
     sort_order: str = "desc",
+    filter_column: Optional[str] = None,
+    filter_value: Optional[str] = None,
 ):
     query = crud.get_submissions_for_user(db, user, column_sort, sort_order)
-    return pagination.response(query)
+    result = pagination.response(query)
+    if filter_column and filter_value:
+        filtered_result = []
+        tables = result["results"]
+        print(tables[0])
+        print(type(tables[0]))
+        if filter_column in tables[0].__table__.columns.keys():  # type: ignore
+            for table in tables:
+                if filter_value.lower() in (str(getattr(table, filter_column))).lower():
+                    filtered_result.append(table)
+            result["results"] = filtered_result
+            result["count"] = len(filtered_result)
+
+    return result
 
 
 @router.get(
--- a/nmdc_server/crud.py
+++ b/nmdc_server/crud.py
@@ -687,7 +687,9 @@ def can_edit_entire_submission(db: Session, submission_id: str, user_orcid: str)
     return (role and models.SubmissionEditorRole(role.role) in contributors_edit_roles) is True
 
 
-def get_submissions_for_user(db: Session, user: models.User, column_sort: str, order: str):
+def get_submissions_for_user(
+    db: Session, user: models.User, column_sort: str, order: str, is_test_submission_filter: bool | None
+):
     """Return all submissions that a user has permission to view."""
     column = (
         models.User.name
@@ -700,6 +702,11 @@ def get_submissions_for_user(db: Session, user: models.User, column_sort: str, o
         .join(models.User, models.SubmissionMetadata.author_id == models.User.id)
         .order_by(column.asc() if order == "asc" else column.desc())
     )
+
+    if is_test_submission_filter != None:
+        all_submissions = all_submissions.filter(
+            models.SubmissionMetadata.is_test_submission == is_test_submission_filter
+        )
 
     if user.is_admin:
         return all_submissions
--- a/nmdc_server/crud.py
+++ b/nmdc_server/crud.py
@@ -692,7 +692,7 @@ def get_submissions_for_user(
     user: models.User,
     column_sort: str,
     order: str,
-    is_test_submission_filter: bool | None,
+    is_test_submission_filter: Optional[bool] = None,
 ):
     """Return all submissions that a user has permission to view."""
     column = (
--- a/nmdc_server/crud.py
+++ b/nmdc_server/crud.py
@@ -688,7 +688,11 @@ def can_edit_entire_submission(db: Session, submission_id: str, user_orcid: str)
 
 
 def get_submissions_for_user(
-    db: Session, user: models.User, column_sort: str, order: str, is_test_submission_filter: bool | None
+    db: Session,
+    user: models.User,
+    column_sort: str,
+    order: str,
+    is_test_submission_filter: bool | None,
 ):
     """Return all submissions that a user has permission to view."""
     column = (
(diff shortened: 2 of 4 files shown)


mslarae13 (approved):
no comments on the code, once it's in -dev I'll test it!

JamesTessmer:
Sounds good, there's a little more work to be done to update the UI but it should hopefully be done this sprint.

pkalita-lbl (commented):
Unfortunately that's not going to work. Filtering the result of the `pagination` step is effectively the same as doing the filtering on the front-end: you're only applying the filter to _a single page_ of results and not to the underlying query.

I just [re-loaded my local database](https://github.com/microbiomedata/nmdc-server/blob/main/docs/development.md#load-production-data) with production data. I can see that I have a total of 295 submissions in my database now:

```sql
select count(*) from submission_metadata; -- 295
```

And 8 of them have the `is_test_submission` flag set:

```sql
select count(*) from submission_metadata where is_test_submission = 't'; -- 8
```

I ran this branch and made an API request with the new filter params:

```bash
curl --request GET \
  --url 'http://localhost:8000/api/metadata_submission?offset=0&limit=10&column_sort=created&sort_order=desc&filter_column=is_test_submission&filter_value=true' \
  --header 'Authorization: Bearer eyJ...gBc'
```

And I only get 2 results back:

```json
{
  "count": 2,
  "results": [
    { ... },
    { ... }
  ]
}
```

That's because only 2 of the 8 test submission are within the first 10 submissions (when sorted by creation date).

Compare that to when the filtering is part of the query itself. In the `get_submissions_for_user` function in `crud.py`, add a `.filter` step after the initial creation of the `all_submissions` query:

```python
all_submissions = (
    db.query(models.SubmissionMetadata)
    .join(models.User, models.SubmissionMetadata.author_id == models.User.id)
    .order_by(column.asc() if order == "asc" else column.desc())
)
# ⚠️ This filter shouldn't always be applied, but just to demonstrate what needs to be done:
all_submissions = all_submissions.filter(models.SubmissionMetadata.is_test_submission == True)
```

Obviously that's not _exactly_ what you want to do because it's now _always_ applying the filter; that's just to demonstrate where the appropriate place to do the filtering is. When I call the endpoint now I get the 8 I was expecting:

```bash
curl --request GET \
  --url 'http://localhost:8000/api/metadata_submission?offset=0&limit=10&column_sort=created&sort_order=desc' \
  --header 'Authorization: Bearer eyJ...gBc'
```
```json
{
  "count": 8,
  "results": [ ... ]
}
```

The second thing I would say is that _if_ we want to do the filtering generically, then we should build in the ability to apply multiple filters simultaneously. I've seen APIs sometimes use separate query parameters with a prefix like:
```text
/metadata_submission?filter.foo=bar&filter.fizz=buzz`
```

Or I've seen them sometimes encode little JSON strings:
```text
/metadata_submission?filter={"foo":"bar","fizz":"buzz"}
```

Again, explore that if you are interested, but I think it's also totally fine to just have a statically named parameter for this:
```text
/metadata_submission?is_test_submission=true
```

JamesTessmer:
Screenshot 2025-03-06 at 12 42 55 PM
@mslarae13 Here's a screenshot of what the portal currently looks like for this.

marySalvi:
If we want to keep the hint for the dropdown, this might be closer to what you were thinking design-wise @JamesTessmer 

Screenshot from 2025-03-10 17-33-09


If we don't want to keep the hint (which I personally don't think we need to keep) this might be closer:
Screenshot from 2025-03-10 17-38-29

web/src/views/SubmissionPortal/Components/SubmissionList.vue
```diff
@@ -202,9 +221,31 @@ export default defineComponent({
...
+          <v-select
+            v-model="isTestFilter"
+            class="my-2"
+            :items="filterSelectValues"
+            item-text="text"
+            item-value="val"
+            :max-width="330"
+            label="Test Submission Filter"
+            hint="Filter whether all submissions should be shown or only some."
+            target="#submissionFilterDropdown"
```
marySalvi, line 245: I am not aware of a target prop on v-select. What is the intent here?

marySalvi (changes requested):

web/src/views/SubmissionPortal/Components/SubmissionList.vue
JamesTessmer: This was a hold over I neglected to remove when I first implemented this as a `v-overflow-btn`, I've removed this now.

web/src/views/SubmissionPortal/Components/SubmissionList.vue
```diff
@@ -124,6 +137,12 @@ export default defineComponent({
       submission,
     };
   },
+  data: () => ({
+    filterSelectValues: [
+      { text: 'All Submissions', val: null },
+      { text: 'Test Only', val: true },
+      { text: 'Real Only', val: false }],
+  }),
```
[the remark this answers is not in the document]
JamesTessmer: Thanks for pointing this out! I'll change this and try to keep it in mind for future work.

nmdc_server/api.py
```diff
@@ -869,8 +869,12 @@ async def list_submissions(
     pagination: Pagination = Depends(),
     column_sort: str = "created",
     sort_order: str = "desc",
+    is_test_submission_filter: Optional[bool] = None,
 ):
-    query = crud.get_submissions_for_user(db, user, column_sort, sort_order)
+    query = crud.get_submissions_for_user(
+        db, user, column_sort, sort_order, is_test_submission_filter
+    )
+    # print(filters)
```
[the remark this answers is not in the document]
JamesTessmer: Removed

nmdc_server/crud.py
```diff
@@ -701,6 +707,11 @@ def get_submissions_for_user(db: Session, user: models.User, column_sort: str, o
         .order_by(column.asc() if order == "asc" else column.desc())
     )
 
+    if is_test_submission_filter != None:
+        all_submissions = all_submissions.filter(
+            models.SubmissionMetadata.is_test_submission == is_test_submission_filter
+        )
+
```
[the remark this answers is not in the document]
JamesTessmer: That's odd, I'm not sure why something would show up as a test with that filter. I'll add a check.

JamesTessmer:
> design

That is much closer to what I was thinking and I like that design.

nmdc_server/crud.py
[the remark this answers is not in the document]
JamesTessmer: @marySalvi Is that same submission returned returned as a non-test as well? Just curious what could be causing that issue.

web/src/views/SubmissionPortal/Components/SubmissionList.vue
```diff
@@ -202,9 +220,31 @@ export default defineComponent({
...
+        </v-col>
+        <v-col
+          cols="3"
+        >
+          <v-select
+            v-model="isTestFilter"
+            :items="filterSelectValues"
+            item-text="text"
+            item-value="val"
+            label="Filter My Submissions"
```
pkalita-lbl, line 243: I'm thinking ahead to a time when we might have more filtering options (I could easily imagine one based on your role in the submission -- editor, viewer, author, etc). In that case this label might be too generic. 

I think the fact that it's a filter is pretty well established based on the option text, so I don't think that word needs to be in the label. The only thing that comes to mind is just labeling this "Test Submissions". I'm not sure I love it, but I don't have any better ideas.

pkalita-lbl (commented):
Just a couple of comments on UI wording

web/src/views/SubmissionPortal/Components/SubmissionList.vue
```diff
@@ -65,6 +66,15 @@ export default defineComponent({
     });
     const isDeleteDialogOpen = ref(false);
     const deleteDialogSubmission = ref<MetadataSubmissionRecord | null>(null);
+    const isTestFilter = ref(null);
+    const filterSelectValues = [
+      { text: 'All Submissions', val: null },
+      { text: 'Test Only', val: true },
+      { text: 'Real Only', val: false }];
```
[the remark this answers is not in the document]
JamesTessmer: Updated these values

JamesTessmer:
Feedback from infra sync:
- No comments on the select drop down
- People want something to mark that submissions are 'real' or 'test' in the table itself

JamesTessmer:
@pkalita-lbl @marySalvi I pushed up a change to add an icon with a tooltip for test submissions. There's a commented version I left in to also include an icon for real/non-test submissions.

I'm open to other approaches of adding the icon for test submissions as well, this was just the first working implementation I thought of.

pkalita-lbl:
I would definitely avoid using the `mdi-test-tube` icon. The meaning of that icon is too overloaded in this domain, in general, and in fact we already use it in the data portal to represent biosamples.

Also, instead of an entirely separate column, I had interpreted Shreyas's suggestion to be an additional bit _within_ the study name column to indicate test submissions. Maybe something like this with a `<v-chip>`?
```vue
<template #[`item.study_name`]="{ item }">
  {{ item.study_name }}
  <v-chip
    v-if="item.is_test_submission"
    color="orange"
    text-color="white"
    small
  >
    TEST
  </v-chip>
</template>
```
image

JamesTessmer:
That makes sense, the v-chip was what I had kind of initially envisioned, or at least attaching something to the submission name like that but I couldn’t find the right way to do it.
 I’ll add your implementation.

web/src/views/SubmissionPortal/Components/SubmissionList.vue
pkalita-lbl: I think updating this label was the only change I was still hoping to see.

JamesTessmer: Let me double check that, I thought I had changed `filterSelectValues` to `testSubmissions` but I might not have pushed it up.

JamesTessmer: @pkalita-lbl Pushed up the change to the label.

pkalita-lbl: Hmmm I'm not seeing the change yet. Did it get pushed?

JamesTessmer: Ah it didn't the push failed for some reason. I have actually pushed it up now.

web/src/views/SubmissionPortal/Components/SubmissionList.vue
```diff
@@ -14,14 +14,20 @@ import TitleBanner from '@/views/SubmissionPortal/Components/TitleBanner.vue';
...
 const headers: DataTableHeader[] = [
   {
     text: 'Study Name',
     value: 'study_name',
   },
+  {
+    text: '',
+    value: 'is_test_submission',
+    sortable: false,
+  },
```
pkalita-lbl, line 30: I believe this needs to be removed still. Otherwise we get a column of the raw boolean values:
image

pkalita-lbl (approved):
I think this is ready to go now. @marySalvi do you have any other feedback here?

marySalvi (approved):

PR #1558, opened 2025-02-28
Obtain data object download counts in single query

### LINKED ISSUE #1548: Explore alternative ways of populating data object download counts
Currently, data download object counts add a seemingly enormous overhead to the biosample search. This is likely due to how download counts are not actually stored in the database, but computed at lookup time by counting rows in the `bulk_download` and `file_download` tables.

Since the results of the biosample search endpoint include this information, this lookup is done for each data object included in the result of a biosample search.

Possible things to experiment with/investigate include:

- Looking at indices (can we speed up the count queries?)
- Materialized views (can we store integer counts of all data objects?, how important is consistency? can we tolerate differences between actual download counts and those displayed on the data portal?)
- Reducing the number of queries (can we rewrite the biosample search so that only one query is run which hydrates the reponse with ALL relevant download counts?)

naglepuff:
Fix #1548 

### Changes

This modifies how the response from the `/biosample/search` endpoint populates download counts for data objects present in the response.

Previously, there would be 2 queries per data object: one to check the number of times a file had been included in a bulk download, and one to check the number of times a file had been downloaded by itself.

Now, the relevant data object IDs are collected, and a single query is made to retrieve bulk download and single file download counts for each of those files.


nmdc_server/api.py
```diff
@@ -184,19 +203,27 @@ async def search_biosample(
...
+    data_object_ids = set()
+
     # Inject file object selection information before serialization.
     # This could potentially be more efficient to do in the database query,
     # but the code to generate the query would be much more complicated.
+    # As a side effect, track all relevant data object IDs for this query.
+    # They will be used to get download counts for all data objects in one
+    # query.
     def insert_selected(biosample: schemas.Biosample) -> schemas.Biosample:
+        print(biosample)
```
danlamanna, line 215:
```suggestion
```

nmdc_server/crud.py
```diff
@@ -320,6 +322,30 @@ def delete_biosample(db: Session, biosample: models.Biosample) -> None:
...
+    bulk_downloads = (
+        db.query(
+            models.BulkDownloadDataObject.data_object_id.label(labels[0]),
+            func.count().label(labels[1]),
+        )
+        .filter(models.BulkDownloadDataObject.data_object_id.in_(data_object_ids))
+        .group_by(models.BulkDownloadDataObject.data_object_id)
+        .order_by(models.BulkDownloadDataObject.data_object_id)
+    )
+    all_counts = file_downloads.union(bulk_downloads).order_by("data_object_id")
```
danlamanna, line 342:
```suggestion
    all_counts = file_downloads.union(bulk_downloads)
```

danlamanna (approved):

PR #1560, opened 2025-03-03
Conditionally prefetch omics processing data

danlamanna:
This aims to reduce the number of queries that the biosample search endpoint uses by using `selectinload` to reduce n+1 queries.

This table shows the before/after locally when querying with no conditions/filters with different limits.

| limit | queries before | queries after | time before (ms) | time after (ms) |
| ----- | -------------- | ------------- | ----------- | ---------- |
| 1     | 26             | 27            | 100         | 65         |
| 15    | 306            | 27            | 770         | 125        |
| 25    | 500            | 27            | 1250        | 170        |

FYI @naglepuff This appears to change the ordering of `omics_processing` results, though it seems like it was already an undefined sort order so maybe this isn't an issue.

### DIFF
--- a/nmdc_server/api.py
+++ b/nmdc_server/api.py
@@ -227,7 +227,10 @@ async def search_biosample(
         return biosample
 
     results = pagination.response(
-        crud.search_biosample(db, query.conditions, data_object_filter), insert_selected
+        crud.search_biosample(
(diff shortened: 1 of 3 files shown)


naglepuff (approved):
This looks good to me! I tested it locally with various filters, looking at the toolbar to see if any shenanigans happened when someone starts adding filters. It appears to be robust to that.

PR #1572, opened 2025-03-14
Add Metaproteomics Analysis Category to UI

### LINKED ISSUE #1550: Add context to Metaproteomics analyses
With the addition of the kaiko metaproteomics analysis option, we need a way to distinguish the data objects associated with two types of  MetaproteomicsAnalysis records.

As is, if there are two MetaproteomicsAnalysis records associated with the same biosample with different values in the metaproteomics_analysis_category, they are displayed like below:

Image

We'd like to add an annotation line below each "Workflow Activity" line following the rules outlined below:

- if the value for metaproteomics_analysis_category == matched_metagenome; display like below
**Workflow Activity:** Metaproteomics Analysis Acvitivty for nmdc:wfmp-XX-XXXX; 
**NMDC metagenome matched** 

OR

- if the value for metaproteomics_analysis_category == in_silico_metagenome; display like below
**Workflow Activity:** Metaproteomics Analysis Acvitivty for nmdc:wfmp-XX-XXXX; 
**Uniprot reference genome matched**

See slides 12 and 13 in https://docs.google.com/presentation/d/1ubYn0TMVNJ6VkG5UoQNns5NsJaqPUnro/edit?usp=sharing&ouid=107901205882214522292&rtpof=true&sd=true

naglepuff:
Fix #1550 

### Data Model

The `metaproteomics_analysis_category` slot is now reflected in the Data Portal's data model. The slot has been added as:

- A property on the `MetaproteomicAnalysis` SQLAlchemy model
- Via migration, a column on the `metaproteomic_analysis` table in postgres
- A property on the `MetaproteomicAnlaysis` pydantic models (used for serialization)

### Front end

A new helper object has been created to map enum values for the `MetaproteomicsAnalysisCategory` [enum](https://microbiomedata.github.io/nmdc-schema/MetaproteomicsAnalysisCategoryEnum/) to display values, as per #1550. As of opening this PR, I did not see anything in the schema itself I could use to translate enum values to display values.

This helper object is used to add the value to the data object tables for Proteomics.

image


kheal (approved):
UI looks as desired - thanks!

PR #1574, opened 2025-03-21
Allow v-pagination to grow to necessary space

### LINKED ISSUE #1221: Pagination widget hides current page number on narrow viewports
There is an issue with the pagination when there are more than 60 samples where it doesn't show what page number you are on. So if you are on sample 70 with a pagination size of 20 I would expect it to show that I'm on page 4 but it just shows ellipses.

Screenshot 2024-04-22 at 10 10 14 AM

Filter used:
https://data.microbiomedata.org/?q=ChgIABABGAMiECJnb2xkOkdzMDEyODg1MCIKHQgAEAsYAyIVIkplbm5pZmVyIFBldHQtUmlkZ2Ui

pkalita-lbl:
Fixes #1221. As a tall-and-narrow-browser-window person, it was driving me nuts.

The issue was that we were putting the `v-pagination` component into a `v-col` that was fixed at 3 columns. That isn't guaranteed to be enough space for the `v-pagination` component to fully display everything.

The fix is to eschew the `v-row` and `v-col` setup in favor of a plain `div` and flexbox utilities. This allows the `v-pagination` component to expand to as much space as it needs.

Before:
image

After:
image

### DIFF
--- a/web/src/components/Presentation/SearchResults.vue
+++ b/web/src/components/Presentation/SearchResults.vue
@@ -55,38 +55,29 @@ export default defineComponent({
 
 <template>
   <div>
-    <v-col
+    <div
       v-if="!disablePagination"
-      cols="12"
-      class="pb-0"
+      class="d-flex pt-2 pb-0 align-end justify-center"
     >
-      <v-row
-        no-gutters
-        justify="center"
-      >
-        <v-col
-          cols="3"
-        >
-          <v-pagination
-            :value="page"
-            :length="Math.ceil(count / rows)"
-            :total-visible="7"
-            class="pt-3"
-            depressed
-            @input="$emit('set-page', $event)"
-          />
-        </v-col>
-        <v-col cols="1">
-          <v-select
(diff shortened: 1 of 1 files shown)


web/src/components/Presentation/SearchResults.vue
```diff
@@ -55,38 +55,29 @@ export default defineComponent({
...
+           positioned it doesn't count towards the `auto` width -->
+      <v-select
+        v-model="rows"
+        :items="[5, 10, 15, 20]"
+        label="Items per page"
+        class="ml-4 flex-grow-0"
+        :style="{ 'flex-basis': '6rem' }"
+        hide-details
+        @input="$emit('set-items-per-page', $event)"
+      />
```
marySalvi, line 79: This is _so minor_ but it looks like the alignment of this in relation the the pagination has changed a bit. I don't Think this should be a blocker but wanted to point it out if its something you didn't want to happen.

marySalvi (approved):

Issue #938, opened 2023-04-06
biosample page logos


aclum:
Similar to what happens on the study pages we would like logos for some of the identifier slots. 

If for class Biosample [img_identifiers](https://microbiomedata.github.io/nmdc-schema/img_identifiers/) is populated then use the attached IMG logo. 
If for class Biosample [gold_biosample_identifiers](https://microbiomedata.github.io/nmdc-schema/gold_biosample_identifiers/) is populated please use the GOLD logo (same logo as what is on the study pages). 
If for class Biosample [emsl_biosample_identifiers](https://microbiomedata.github.io/nmdc-schema/emsl_biosample_identifiers/) is populated please use the EMSL logo
![emsl_logo](https://user-images.githubusercontent.com/8196591/230498905-2c165920-8bb8-44bc-958c-65d58fb7948a.png)

cc @mslarae13

ssarrafan:
Moved to Data Portal board. Removed from post GSP board.

aclum:
example of GOLD study page link logo https://data.microbiomedata.org/details/study/nmdc:sty-11-r2h77870

yxu-lanl:
@aclum Are these acceptable?

https://data.microbiomedata.org/details/sample/nmdc:bsm-11-6p6mhh86

<img width="606" alt="Image" src="https://github.com/user-attachments/assets/fc989b60-655c-49f5-92e5-42f8aa3619ac" />

https://data.microbiomedata.org/details/sample/nmdc:bsm-13-2yga6857

<img width="488" alt="Image" src="https://github.com/user-attachments/assets/d3d97b36-1f40-4bd5-99dc-c01cbc2ac4c0" />

https://data.microbiomedata.org/details/sample/nmdc:bsm-13-wktc3357

<img width="516" alt="Image" src="https://github.com/user-attachments/assets/e5b82263-b0cf-4047-8f55-5cd1edc00116" />

aclum:
looks great

yxu-lanl:
I submitted a PR. Who will review it? Thanks.
https://github.com/microbiomedata/nmdc-server/pull/1577

aclum:
please add @naglepuff and @marySalvi

PR #1577, opened 2025-03-25
938 biosample page logos

yxu-lanl:
https://github.com/microbiomedata/nmdc-server/issues/938


web/src/components/Presentation/AttributeList.vue
```diff
@@ -128,11 +136,13 @@ export default defineComponent({
         v-for="({ name, target }) in alternateIdentifiers"
         :key="name"
         v-bind="{ item, field, link: { name, target } }"
+        :image="name.startsWith('gold') ? images.gold : name.startsWith('img') ? images.img : undefined"
```
naglepuff, line 139: Minor issue, but the prop definition in [AttributeItem.vue ](https://github.com/microbiomedata/nmdc-server/blob/da8bb307b45492fe971f05c62d5d0d5ee7b60957/web/src/components/Presentation/AttributeItem.vue#L31-L33) has it typed as a string, so it might be a little cleaner to pass in `''` if no image matches.

```suggestion
        :image="name.startsWith('gold') ? images.gold : name.startsWith('img') ? images.img : ''"
```

web/src/components/Presentation/AttributeList.vue
```diff
@@ -98,6 +98,14 @@ export default defineComponent({
...
       displayFields,
       relatedBiosamples,
+      images: {
+        // eslint-disable-next-line global-require
+        gold: require('@/assets/GOLD.png'),
+        // eslint-disable-next-line global-require
+        img: require('@/assets/IMG.png'),
+        // eslint-disable-next-line global-require
+        emsl: require('@/assets/EMSL.png'),
+      },
```
naglepuff, line 108: This is great, it's exactly what we do for study links in [StudyPage.vue](https://github.com/microbiomedata/nmdc-server/blob/da8bb307b45492fe971f05c62d5d0d5ee7b60957/web/src/views/IndividualResults/StudyPage.vue#L222-L229).

I do however wonder about the use of `require` here. Most of the code base follows ES6 import/export syntax, but here we're using `require` to import images.

I think it might be worth opening up an issue to track this, and at some point investigate switching over to using `import` syntax. I don't think that needs to be done as part of this PR though.

web/src/components/Presentation/AttributeList.vue
yxu-lanl: Thanks!

web/src/components/Presentation/AttributeList.vue
naglepuff: FYI I've opened issue #1579 to track updating these `require` statements in the future.

yxu-lanl:
I am not authorized to push to this branch.

Issue #902, opened 2023-03-23
Data not persisted on transition between Environment Package and DataHarmonizer views


pkalita-lbl:
1. From https://data-dev.microbiomedata.org/submission/home start a new submission.
2. Fill in necessary information up to the Environment Package view. If you're watching the browser's Network monitor you should have seen that a `PATCH /api/metadata_submission/{id}` request was sent when transitioning between each screen.
3. On the Environment Package screen select anything _other_ than the default package (e.g. choose "air"). Click the "Go to next step" button. Note that there was _no_ `PATCH /api/metadata_submission/{id}` request made when transitioning to the DataHarmonizer view.
4. On the DataHarmonizer view the first tab should read "air". However if you reload the page without making any changes to the spreadsheet, the first tab will be "soil". This is a direct consequence of the missing request. (Had you modified any data in the spreadsheet a `PATCH` request with the correct template information would be sent and you wouldn't hit this bug.)

pkalita-lbl:
I think this is a consequence of having the DataHarmonizer view's route defined _outside_ of the `StepperView` which is the parent of all the other steps: https://github.com/microbiomedata/nmdc-server/blob/950a8b2bff8ec3981c9ae410a529eb3b25417b32/web/src/plugins/router.ts#L47-L98

@naglepuff Am I missing something here or is there any reason to not include `HarmonizerView` within `StepperView`?

PR #1588, opened 2025-04-02
Remove default template choice

JamesTessmer:
This PR changes the metadata submission to not have a default template choice for the data harmonizer.
Now users must select at least one template.
Supporting changes were also added for clarity on the template selection page.

Relevant issue: https://github.com/microbiomedata/issues/issues/1002

### DIFF
--- a/web/src/plugins/router.ts
+++ b/web/src/plugins/router.ts
@@ -20,6 +20,7 @@ import ValidateSubmit from '@/views/SubmissionPortal/Components/ValidateSubmit.v
 import SubmissionList from '@/views/SubmissionPortal/Components/SubmissionList.vue';
 
 import { unlockSubmission } from '@/views/SubmissionPortal/store/api';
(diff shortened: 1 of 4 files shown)


pkalita-lbl:
As I was testing I noticed that under some circumstances, the environment package selection was not being saved. Upon closer inspection I realized I had rediscovered #902. But now it's a bit more noticeable with no default environment package.

I pushed a change that should fix #902. It's basically what Mike suggested in that issue: pushing the call to `incrementalSaveRecord` when the route changes "up" in the routing hierarchy. In particular, I hooked it into the same  `router.beforeEach` callback that the lock/unlock mechanism uses because the logic is partially the same.

@naglepuff make sense to you?

naglepuff (approved):
This looks good to me! Tested it locally and verified that I saw a `PATCH` request when moving from the templates tab to the harmonizer tab.

JamesTessmer:
Sounds like we should be good to merge this!

Issue #1485, opened 2024-12-05
`refactor:` Remove `omics_processing` from codebase


naglepuff:
Spun out of #1331, post-Berkeley cleanup

This would be an effort to keep the SQL schema used by the data portal more in line with the NMDC schema. It would include:

- [ ]  Rename the model `OmicsProcessing` to `DataGeneration`, including the table name
- [ ] Update FKs, association tables, etc.
- [ ] Update corresponding endpoint function names, and any classes that reference `OmicsProcessing`
- [ ] Update variable names

I expect this to be quite a large undertaking.

PR #1589, opened 2025-04-02
Adjust data generation counts by type for poolable replicates

### LINKED ISSUE #1463: Data portal - how manifest_set records should look in the UI
This ticket replaces #1365

naglepuff:
Fix #1463 

### Changes
New data modeling and counting logic has been implemented to accommodate the use case outlined in issue #1463 where `data_generation` objects that have outputs in the same `manifest` of type `poolable_replicates` should only "count" as 1 for the purposes of the data portal.

#### Data model changes
A new column has been added to the `omics_processing` table (note: this table represents `data_generation` records. there is an issue #1485 to track renaming). This column contains either the ID of the `data_generation` itself, or an ID of a `manifest` of type `poolable_replicates`. This column is used for counting.

The column is populated during ingest. When ingesting a `data_generation`, all of the output `data_object` IDs are collected. Previously, we had logic that set the `workflow_type` for all of these data objects to `raw_data`. Now, there is additional logic that checks each data object for the `in_manifest` slot.

If `manifest` records are found, each one is checked for its `manifest_category`. By iterating over all of the output data objects and each manifest connected to those data objects, we capture the _last_ `manifest` of category `poolable_replicates` and store it in the aforementioned column.

*Note that these changes are tailored specifically to the use case described in the linked issue, and make an assumption that for this use case, each `data_generation` will only be linked via `data_object` output to a _single_ `manifest`.

#### Counting logic changes
The `OmicsProcessingQuerySchema` class now overrides the `facet` method inherited from `BaseQuerySchema`. It includes special handling if the facet attribute is `omics_type`. In this case, instead of counting all rows in the final query, it counts distinct values in the new column.

Since the new column contains the `manifest` ID that groups certain `data_generation`s **or** the ID of the `data_generation` itself (in the case where no `manifest` exists), this will effectively group together related `data_generation`s in the final count.


naglepuff:
@marySalvi I've rebased this so that the migration now points to the correct "down revision," and I've fixed the failing CI test, so this is ready for review. An ingest would be needed to see the changes in action. #1463 has an example biosample search criteria that will be affected by this change.

marySalvi:
I've had ingest stall out twice on 'Loading omics processing...' I am not seeing anything obvious that is causing this but I am concerned since the changes are related to loading the omics processing.

@naglepuff Would you mind doing a full ingest from prod to see if you can recreate?

nmdc_server/ingest/omics_processing.py
```diff
@@ -106,6 +106,29 @@ def get_configuration_name(mongodb: Database, configuration_id: str, config_map)
...
+    Determine which poolable replicate manifest, if any, this data_generation is associated with.
+
+    Returns either the ID of a manifest of type poolable_replicates, or the ID of the
+    data_generation itself. Used for the purposes of counting data_generations for the Data Portal.
+    """
+    data_object_document = mongodb["data_object_set"].find_one({"id": data_object_id})
+    if not data_object_document or not data_object_document.get("in_manifest", None):
+        return omics_processing_id
+
+    manifest_id = data_object_document.get("in_manifest")[0]
```
marySalvi, line 122: Is it possible for "in_manifest" to be empty? Could this be the hang up?

naglepuff: I don't think so. I think that if `in_manifest` were empty, the condition on line 119 would be `False`, and the function would just return the data generation's id (`omics_processing_id`)

marySalvi (approved):
LGTM and @naglepuff was able to successfully do a full ingest from prod quelling my concerns

aclum (approved):

PR #1590, opened 2025-04-03
#2047 - Add enum/permissible value validation to API endpoint

cristina-stonepedraza:
Add to the existing MIxS environmental triad slot review API endpoint to include enum validation to ensure that env package, local scale, broad scale, and medium fields include slots that are part of each enum's permissible value list.

### DIFF
--- a/nmdc_server/api.py
+++ b/nmdc_server/api.py
@@ -10,6 +10,7 @@ from fastapi import APIRouter, Depends, Header, HTTPException, Query, Response, 
 from fastapi.responses import JSONResponse
 from sqlalchemy.orm import Session
 from starlette.responses import StreamingResponse
+from linkml_runtime.utils.schemaview import SchemaView
 
 from nmdc_server import crud, jobs, models, query, schemas, schemas_submission
 from nmdc_server.auth import admin_required, get_current_user, login_required_responses
@@ -28,6 +29,8 @@ from nmdc_server.models import (
 )
 from nmdc_server.pagination import Pagination
 from nmdc_server.table import Table
+
+import yaml
 
 router = APIRouter()
 
@@ -704,8 +707,12 @@ async def download_zip_file(
     "/metadata_submission/mixs_report",
     tags=["metadata_submission"],
 )
+# async def get_metadata_submissions_mixs(
+#     db: Session = Depends(get_db), user: models.User = Depends(get_current_user)
+# ):
+    
 async def get_metadata_submissions_mixs(
-    db: Session = Depends(get_db), user: models.User = Depends(get_current_user)
+    db: Session = Depends(get_db)
 ):
     r"""
     Generate a TSV-formatted report of biosamples belonging to submissions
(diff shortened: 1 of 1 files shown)


cristina-stonepedraza:
Here is an example of the TSV file results so far: 
image

nmdc_server/api.py
```diff
@@ -807,6 +884,27 @@ async def get_metadata_submissions_mixs(
 
     return response
 
+def fetch_nmdc_submission_schema_view():
+
+    # Use SchemaView to open nmdc_submission_schema.yaml and get the enums
+    view = SchemaView("https://raw.githubusercontent.com/microbiomedata/submission-schema/refs/heads/main/src/nmdc_submission_schema/schema/nmdc_submission_schema.yaml")
```
pkalita-lbl, line 890: `nmdc-submission-schema` is already a dependency of this project, so there's no need to fetch from GitHub here. In addition to the unnecessary overhead of making a request over the internet, the bigger concern is that fetching it from the `main` branch will, in general, get a version that is ahead of what's used in here `nmdc-server`.

The good news is that there is already code which generates a `SchemaView` instance from the schema file bundled in the `nmdc-submission-schema` package: https://github.com/microbiomedata/nmdc-server/blob/cd9bc6d2ef0c2b8e7c476673c87eb103411bbc5f/nmdc_server/static_files.py#L21-L27

I would recommend extracting that code into a utility function which returns a `SchemaView` instance (the result can/should even be [cached](https://docs.python.org/3/library/functools.html#functools.cache)). The new function can be called from the `generate_submission_schema_files` function and this function (although I might nitpick the name of this `fetch_nmdc_submission_schema_view` function since it doesn't return a `SchemaView`).

pkalita-lbl (commented):
Looks pretty good. Two minor comments below.

One bigger comment is that it seems like logic around report generation is starting to accumulate in `api.py`. I think ideally that module would be focussed the functions that raise `HTTPExceptions` or return responses; the main logic would be delegated out to functions in other modules (e.g. `crud.py` for reading/writing to the database).

I think it might be time to add a new module (maybe `reporting.py`?) where all the actual report generation logic can live. So then the endpoint handler functions in `api.py` might look like:

```python
@router.get(
    "/metadata_submission/mixs_report",
    tags=["metadata_submission"],
)
async def get_metadata_submissions_mixs(
    db: Session = Depends(get_db), user: models.User = Depends(get_current_user)
):
    """ ... """
    if not user.is_admin:
        raise HTTPException(status_code=403, detail="Your account has insufficient privileges.")

    q = crud.get_query_for_submitted_pending_review_submissions(db)
    submissions = q.all()

    buffer = reporting.mixs_report(submissions)

    return StreamingResponse(...)


@router.get(
    "/metadata_submission/report",
    tags=["metadata_submission"],
)
async def get_metadata_submissions_report(
    db: Session = Depends(get_db),
    user: models.User = Depends(get_current_user),
):
    """ ... """
    if not user.is_admin:
        raise HTTPException(status_code=403, detail="Your account has insufficient privileges.")

    q = crud.get_query_for_all_submissions(db)
    submissions = q.all()

    buffer = reporting.basic_info_report(submissions)

    return StreamingResponse(...)
```

That's not a make-or-break thing right now, but we should do it at some point. If you have the bandwidth to do it now maybe think about it.

cristina-stonepedraza:
@pkalita-lbl I think you make a good point in your comment above about separating out the code for report generation into a new module! I created a separate issue to address it: #1662 

For this PR, does it look good to merge? 
I still do not have merge permissions, so if it looks good, could you merge it for me or grant me permissions to? 
Thanks!

pkalita-lbl (approved):
Thanks, yes I'm perfectly happy to have the refactoring split off as a separate issue.

I'll merge this one now, but I'll also look into updating the repo settings so you'll be able to merge your own PRs in the future.

PR #1594, opened 2025-04-09
Fix multiomics form validation

pkalita-lbl:
This is follow-on work to #1581 in order to finesse some of the validation logic of the multiomics form.

1. These changes ensure that we ask about non-user facility metagenome and metatranscriptome sequencing data, the compatibility questions are truly required. The gist of the fix is to add `:rules` to the relevant radio groups and to revalidate the form as new elements are added to it. I also updated the logic so that the answers the the "data compatible" and "data interleaved" questions are cleared when the related checkbox is unchecked.

| Before | After |
| --- | --- |
| image | image |

2. These changes fix the validation logic around the "What kind of project have you been awarded?" question so that text field in the "Other" option is not in an error state by default. The fix involves disentangling the state attached to the text box and the checkbox (previously `multiOmicsForm.otherAward` was attached to _both_). 

| Before | After |
| --- | --- |
| image | image |

3. These changes prevent the questions about data compatibility from being presented when the data have _not_ already been generated. This, at least, is what we said we'd do in the mockups.

| Before | After |
| --- | --- |
| image | image |

### DIFF
--- a/web/src/views/SubmissionPortal/Components/DataTypes.vue
+++ b/web/src/views/SubmissionPortal/Components/DataTypes.vue
@@ -15,7 +15,7 @@ export default defineComponent({
       default: true,
     },
   },
-  setup(props, { emit }) {
+  setup(_, { emit }) {
     const dataCaveat = 'You may proceed with your submission for sample metadata capture. However, there will not be place to provide information about your existing sequencing data as the methods are not supported by NMDC Workflows';
 
     const handleMetagenomeChange = (value: string[]) => {
--- a/web/src/views/SubmissionPortal/Components/DataTypes.vue
+++ b/web/src/views/SubmissionPortal/Components/DataTypes.vue
@@ -11,13 +11,47 @@ export default defineComponent({
       default: 'Data Types',
     },
   },
-  setup() {
+  setup(props, { emit }) {
     const dataCaveat = 'You may proceed with your submission for sample metadata capture. However, there will not be place to provide information about your existing sequencing data as the methods are not supported by NMDC Workflows';
+
+    const handleMetagenomeChange = (value: string[]) => {
+      if (!value.includes('mg')) {
+        multiOmicsForm.mgCompatible = undefined;
+        multiOmicsForm.mgInterleaved = undefined;
+      }
+      emit('revalidate');
+    };
+
+    const handleMgCompatibleChange = (value: boolean) => {
+      if (!value) {
+        multiOmicsForm.mgInterleaved = undefined;
+      }
+      emit('revalidate');
+    };
+
+    const handleMetatranscriptomeChange = (value: string[]) => {
+      if (!value.includes('mt')) {
+        multiOmicsForm.mtCompatible = undefined;
+        multiOmicsForm.mtInterleaved = undefined;
+      }
+      emit('revalidate');
+    };
+
+    const handleMtCompatibleChange = (value: boolean) => {
+      if (!value) {
+        multiOmicsForm.mtInterleaved = undefined;
+      }
+      emit('revalidate');
+    };
 
     return {
       dataCaveat,
       multiOmicsForm,
       templateChoiceDisabled,
+      handleMetagenomeChange,
+      handleMgCompatibleChange,
+      handleMetatranscriptomeChange,
+      handleMtCompatibleChange,
     };
   },
 
@@ -42,6 +76,7 @@ export default defineComponent({
       value="mg"
       :disabled="templateChoiceDisabled"
       hide-details
+      @change="handleMetagenomeChange"
     />
     <div
       v-if="multiOmicsForm.omicsProcessingTypes.includes('mg')"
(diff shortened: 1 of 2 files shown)


web/src/views/SubmissionPortal/Components/MultiOmicsDataForm.vue
```diff
@@ -268,15 +277,15 @@ export default defineComponent({
         </div>
 
         <v-radio
-          :value="multiOmicsForm.otherAward"
+          :value="OTHER"
```
marySalvi, line 280: Is it necessary to make this use a variable? I think I would set inline with a string instead:
```suggestion
          value="OTHER"
```
This might just come down to preference but I don't typically use a component scoped const in this case.

web/src/views/SubmissionPortal/Components/DataTypes.vue
```diff
@@ -42,9 +80,10 @@ export default defineComponent({
       value="mg"
       :disabled="templateChoiceDisabled"
       hide-details
+      @change="handleMetagenomeChange"
     />
     <div
-      v-if="multiOmicsForm.omicsProcessingTypes.includes('mg')"
+      v-if="showDataCompatibilityQuestions && multiOmicsForm.omicsProcessingTypes.includes('mg')"
```
marySalvi, line 86: Would it make sense to check `multiOmicsForm.facilityGenerated === false` here and line 170 instead of creating a new prop? It seems as though that is the condition for these extra fields to show.

web/src/views/SubmissionPortal/Components/DataTypes.vue
```diff
@@ -10,14 +10,52 @@ export default defineComponent({
       type: String,
       default: 'Data Types',
     },
+    showDataCompatibilityQuestions: {
+      type: Boolean,
+      default: true,
+    },
   },
-  setup() {
+  setup(props, { emit }) {
```
marySalvi, line 18: I don't think you need to pass props here because it doesn't look like you are accessing them in the setup function.

marySalvi (changes requested):

web/src/views/SubmissionPortal/Components/MultiOmicsDataForm.vue
pkalita-lbl: I did it that way because the function in `otherAwardValidationRules` checks against the value the `OTHER` constant. Without the variable there would be two hardcoded `"OTHER"` strings floating around; not the end of the world, but the variable made more sense to me.

web/src/views/SubmissionPortal/Components/DataTypes.vue
pkalita-lbl: No `multiOmicsForm.facilityGenerated === false` applies to both places where the `<DataTypes>` component is shown: https://github.com/microbiomedata/nmdc-server/blob/b1f8b7df96aeea9fe03217a2a7d960db2bc736a7/web/src/views/SubmissionPortal/Components/MultiOmicsDataForm.vue#L314-L325

I guess it would be possible to replace the new prop with `multiOmicsForm.doe === false`. But since we already have this separate `<DataTypes>` component, I figured that the prop helps keep the component be more encapsulated and less knowledgeable about the structure of `multiOmicsForm`.

web/src/views/SubmissionPortal/Components/DataTypes.vue
pkalita-lbl: Oh, I was just sort of blindly following a pattern I saw elsewhere. Maybe this is a dumb Vue question then, but is it possible to just get `emit` without getting `props`?

web/src/views/SubmissionPortal/Components/DataTypes.vue
marySalvi: I think I might be confused. I thought that the first `<DataTypes>` was dependent on `multiOmicsForm.dataGenerated === false` and the second (the one that shows the doe compatibility questions) is dependent on `multiOmicsForm.facilityGenerated === false`

_data_ vs _facility_ generated.

web/src/views/SubmissionPortal/Components/MultiOmicsDataForm.vue
marySalvi: Fair enough. Definitely fine to keep in.

web/src/views/SubmissionPortal/Components/DataTypes.vue
pkalita-lbl: Whoops, my bad! I wasn't reading that closely enough 🤦🏻 

I think I still prefer passing in an explicit prop to `<DataTypes>`. You could make a similar argument about the existing `legend` prop. Could that text be changed from _within_ `<DataTypes>` by inspecting the `multiOmicsForm` state? Sure, but to me the explicitness of have a separate prop makes the behavior of the component more clear from the outside.

web/src/views/SubmissionPortal/Components/DataTypes.vue
pkalita-lbl: Got it, thanks.

web/src/views/SubmissionPortal/Components/DataTypes.vue
marySalvi: Fair enough

marySalvi (approved):

PR #1600, opened 2025-04-11
 Empty Data Sheet Validation/Warning

### LINKED ISSUE #946: Shouldn't be able to submit an empty metadata sheet
Currently, as long as I press validate, even if nothing has bene entered, I can submit and empty study with no samples / metadata

criteria for completion
- Require that something valid be entered into the submission portal 
Screen Shot 2023-04-07 at 3 15 20 PM
Screen Shot 2023-04-07 at 3 15 31 PM

cristina-stonepedraza:
Closes #946 

Add to existing data sheet validation code to also check for an empty data sheet. If the data sheet is empty, a warning now flashes at the bottom of the screen warning the user that the submission sheet is empty. The 'Submit' button is also grayed-out until this is resolved. 

image

### DIFF
--- a/web/src/views/SubmissionPortal/HarmonizerView.vue
+++ b/web/src/views/SubmissionPortal/HarmonizerView.vue
@@ -136,6 +136,7 @@ export default defineComponent({
     const snackbar = ref(false);
     const importErrorSnackbar = ref(false);
     const notImportedWorksheetNames = ref([] as string[]);
+    const emptySheetSnackbar = ref(false);
 
     watch(activeTemplate, () => {
       // WARNING: It's important to do the column settings update /before/ data. Otherwise,
@@ -259,7 +260,7 @@ export default defineComponent({
 
     async function validate() {
       const data = harmonizerApi.exportJson(); // Gets data from harmonizer API
-      
+
       // Check if the spreadsheet is empty
       const isEmpty = Object.keys(data).length === 0;
       // Update invalid cells if empty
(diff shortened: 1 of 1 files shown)


ssarrafan:
@cristina-stonepedraza is this still a draft?  
I'll move it to the next sprint but usually we don't add draft issues to sprints.

pkalita-lbl (approved):
Looks good. One minor comment.

Edit: whoops, where did my comment go?

web/src/views/SubmissionPortal/HarmonizerView.vue
```diff
@@ -386,7 +387,26 @@ export default defineComponent({
...
+      if (isEmpty) {
+        invalidCells.value = {
+          ...invalidCells.value,
+          [activeTemplateKey.value]: data,
+        };
+        tabsValidated.value = {
+          ...tabsValidated.value,
+          [activeTemplateKey.value]: false,
+        };
+        snackbar.value = Object.values(tabsValidated.value).every((value) => value);
```
pkalita-lbl, line 404: There's no need to compute this here. Given the way `tabsValidated` is set in the line above, this will always result in `snackbar.value = false`. 

Since we now have several possible snackbar displays, it also might be more clear if we rename `snackbar` to something like `validationSuccessSnackbar`. Or, extra credit, is it possible to store a `ref` that contains the full snackbar state (message, color, timeout) and then only render one `v-snackbar` component?

PR #1611, opened 2025-04-17
Avoid nginx rebuild by basing nmdc-server/client image off static older image

pkalita-lbl:
These changes avoid the rebuild of nginx to add the mod_zip module. As noted in the comments, we lost the ability to rebuild nginx 1.22. That's because nginx decomissioned the Mercurial repo which was required for the build process. 

These changes implement a very short-term bandaid of basing the `nmdc-server/client` image off of the last stable `nmdc-server/client` version -- which already has nginx built with mod_zip.

Thanks for the suggestion @aclum! It's a wild one, but it works.

### DIFF
--- a/web/Dockerfile
+++ b/web/Dockerfile
@@ -1,61 +1,3 @@
-FROM nginx:1.22-alpine as nginx_builder
-
-ENV ENABLED_MODULES=modzip
-RUN set -ex \
-    && if [ "$ENABLED_MODULES" = "" ]; then \
-        echo "No additional modules enabled, exiting"; \
-        exit 1; \
-    fi
-
-COPY ./nginx/ /modules/
-
-RUN set -ex \
-    && apk update \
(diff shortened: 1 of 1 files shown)


naglepuff:
@pkalita-lbl thanks for moving quick on this. I can take a look tomorrow morning

naglepuff (approved):
Changes to the docker file all make sense to me. There's now a build stage where the new client application is built, and then that is placed into the right directory in the base image (a recent build of nmdc-server/client).

I was able to get this to build locally, with some warnings:

``` 
 - FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 1)
 - FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 18)
 - JSONArgsRecommended: JSON arguments recommended for CMD to prevent unintended behavior related to OS signals (line 26)
```
I also made some changes to the client code base, re-built, and verified that my changes carried through. Once this lands in dev we should do lots of testing, especially around (bulk) downloads.

PR #1614, opened 2025-04-22
Use zipstreamer for bulk downloads

### LINKED ISSUE #1587: make bulk download more fault tolerant
user story: If even a single file has a different file size than what is in mongo bulk download, which uses mod_zip, fails

discuss at the infra admin meeting 4/10

cc @ssarrafan @jeffbaumes @eecavanna

naglepuff:
Fix #1587 

### New service

Previously, our bulk download feature relied on an nginx plugin called [mod_zip](https://github.com/evanmiller/mod_zip). We have decided to try and move away from mod_zip, due to its need for file size information and trouble building with newer nginx releases. Due to previous experience with [zipstreamer](https://github.com/scosman/zipstreamer), we have decided to try using that as a replacement.

The default `docker-compose.yml` file has an additional service: zipstreamer. It is a microservice written in Go that can be used to build and stream a zip file given a list of URLs. Note that we will need to update the Rancher deployments to include this new service so that zip download works when this PR lands.

### Removal of `mod_zip`

Replacing mod_zip with zipstreamer was fairly plug-and-play. The major changes are that:

1. zipstreamer wants a manifest of files in a different format. Instead of a list, it wants a json blob
2. zipstreamer does not need all files to appear local. mod_zip needed the files to be from a single origin, and we had proxy configuration and string replacement logic to accommodate. this code has been removed
3. We now proxy the download through the `backend` service. Instead of returning a response with a special header that had meaning for mod_zip, we now reach to the `zipstreamer` service directly from `backend`, and stream the response to the client. Zipstreamer can also be configured to stream directly to the client, but that requires it to be exposed to the internet. The current implementation means it only needs to be accessible by the `backend` service. Since our application is running in an async manner with many workers, this proxying shouldn't be too much of an issue. If needed, we can move to the other model in the future.

The URL used by the backend, as well as the chunk size for streaming, are both configurable as environment variables. We should make sure to set:

`NMDC_ZIP_STREAMER_URL` and `NMDC_ZIP_STREAMER_CHUNK_SIZE_BYTES`

### Other changes
Since zipstreamer doesn't care about file size information, we can remove the warnings during ingest about missing file sizes. We can also remove the safeguards when building the manifest for the streaming service.

We no longer need to do an ouroborus-like build for our client image. We can go back to using a base nginx image - now updated (`1.22 -> 1.28`).

The function that actually does the request to zipstreamer during the bulk download process is mocked during testing.

Todo:

- [x] Update nginx base image version
- [x] Update ingest to no longer transform data object URLs (they don't need to be relative anymore)
- [x] Update ingest to no longer warn that data objects without a size can't be downloaded
- [x] Remove check for no size when building the zip streamer json blob (we can download those data objects now)

### DIFF
--- a/nmdc_server/api.py
+++ b/nmdc_server/api.py
@@ -684,13 +684,12 @@ async def get_data_object_aggregation(
 
 
 async def stream_zip_archive(zip_file_descriptor: Dict[str, Any]):
-    zip_streamer_url = "http://zipstreamer:4008/download"
-    chunk_size = 100
+    settings = Settings()
     async with httpx.AsyncClient() as client:
         async with client.stream(
-            "POST", zip_streamer_url, json=zip_file_descriptor
+            "POST", settings.zip_streamer_url, json=zip_file_descriptor
         ) as response:
-            async for chunk in response.aiter_bytes(chunk_size=chunk_size):
+            async for chunk in response.aiter_bytes(chunk_size=settings.zip_streamer_chunk_size):
                 yield chunk
 
 
--- a/nmdc_server/api.py
+++ b/nmdc_server/api.py
@@ -689,14 +689,12 @@ async def stream_zip_archive(zip_file_descriptor: Dict[str, Any]):
     a ZIP archive in response, which this function yields in chunks.
     """
     settings = Settings()
-    async with httpx.AsyncClient() as client:
-        async with client.stream(
-            "POST", settings.zip_streamer_url, json=zip_file_descriptor
-        ) as response:
-            async for chunk in response.aiter_bytes(
-                chunk_size=settings.zip_streamer_chunk_size_bytes
-            ):
-                yield chunk
+    async with (
+        httpx.AsyncClient() as client,
+        client.stream("POST", settings.zip_streamer_url, json=zip_file_descriptor) as response,
+    ):
+        async for chunk in response.aiter_bytes(chunk_size=settings.zip_streamer_chunk_size_bytes):
+            yield chunk
 
 
 @router.get(
--- a/nmdc_server/api.py
+++ b/nmdc_server/api.py
@@ -693,7 +693,9 @@ async def stream_zip_archive(zip_file_descriptor: Dict[str, Any]):
         async with client.stream(
             "POST", settings.zip_streamer_url, json=zip_file_descriptor
         ) as response:
-            async for chunk in response.aiter_bytes(chunk_size=settings.zip_streamer_chunk_size):
+            async for chunk in response.aiter_bytes(
+                chunk_size=settings.zip_streamer_chunk_size_bytes
+            ):
                 yield chunk
 
 
--- a/nmdc_server/api.py
+++ b/nmdc_server/api.py
@@ -684,6 +684,10 @@ async def get_data_object_aggregation(
 
 
 async def stream_zip_archive(zip_file_descriptor: Dict[str, Any]):
+    r"""
+    Sends the specified `zip_file_descriptor` to ZipStreamer and receives
+    a ZIP archive in response, which this function yields in chunks.
+    """
     settings = Settings()
     async with httpx.AsyncClient() as client:
         async with client.stream(
--- a/nmdc_server/api.py
+++ b/nmdc_server/api.py
@@ -683,6 +683,17 @@ async def get_data_object_aggregation(
     return query.aggregate(db)
 
 
+async def stream_zip_archive(zip_file_descriptor: Dict[str, Any]):
+    zip_streamer_url = "http://zipstreamer:4008/download"
+    chunk_size = 100
+    async with httpx.AsyncClient() as client:
+        async with client.stream(
+            "POST", zip_streamer_url, json=zip_file_descriptor
+        ) as response:
+            async for chunk in response.aiter_bytes(chunk_size=chunk_size):
+                yield chunk
+
+
 @router.get(
     "/bulk_download/{bulk_download_id}",
     tags=["download"],
@@ -694,18 +705,9 @@ async def download_zip_file(
     zip_file_descriptor = crud.get_zip_download(db, bulk_download_id)
     if not zip_file_descriptor:
         return
-    zip_streamer_url = "http://zipstreamer:4008/download"
-
-    async def stream_results():
-        async with httpx.AsyncClient() as client:
-            async with client.stream(
-                "POST", zip_streamer_url, json=zip_file_descriptor
-            ) as response:
-                async for chunk in response.aiter_bytes(chunk_size=100):
-                    yield chunk
 
     return StreamingResponse(
-        stream_results(),
+        stream_zip_archive(zip_file_descriptor),
         media_type="application/zip",
         headers={
             "Content-Disposition": f"attachment; filename={zip_file_descriptor['suggestedFilename']}"  # noqa: E501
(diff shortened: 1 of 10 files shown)


nmdc_server/api.py
```diff
@@ -682,6 +683,16 @@ async def get_data_object_aggregation(
     return query.aggregate(db)
 
 
+async def stream_zip_archive(zip_file_descriptor: Dict[str, Any]):
```
eecavanna, line 686: I recommend adding a docstring, where the target audience is people that are considering invoking this function or debugging this code. Here's a docstring that I distilled from a longer one that GitHub Copilot came up with:

```suggestion
async def stream_zip_archive(zip_file_descriptor: Dict[str, Any]):
    r"""
    Sends the specified `zip_file_descriptor` to ZipStreamer and receives
    a ZIP archive in response, which this function yields in chunks.
    """
```

nmdc_server/config.py
```diff
@@ -41,6 +41,10 @@ class Settings(BaseSettings):
     orcid_client_secret: str = "oauth secret key"
     orcid_authorize_scope: str = "/authenticate"
 
+    # for zipstreamer
+    zip_streamer_url: str = "http://zipstreamer:4008/download"
+    zip_streamer_chunk_size: int = 512
```
eecavanna, line 46: I recommend indicating the **units** for the size (i.e. bytes versus kilobytes), either as part of the "setting" (i.e. environment variable) name, or in the `# zip streamer` comment above this block of "settings".

nmdc_server/crud.py
```diff
@@ -486,36 +485,30 @@ def create_bulk_download(
         raise
 
 
-def get_zip_download(db: Session, id: UUID) -> Optional[str]:
+def get_zip_download(db: Session, id: UUID) -> Dict[str, Any]:
     """Return a download table compatible with mod_zip."""
```
eecavanna, line 489: I think the changes in this diff make this docstring obsolete.

nmdc_server/crud.py
```diff
@@ -486,36 +485,30 @@ def create_bulk_download(
...
     bulk_download = db.query(models.BulkDownload).get(id)
     if bulk_download is None:
         raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail="Bulk download not found")
     if bulk_download.expired:
         raise HTTPException(status_code=status.HTTP_410_GONE, detail="Bulk download expired")
-    content = []
+    zip_file_descriptor: Dict[str, Any] = {"suggestedFilename": "archive.zip"}
+    files = []
```
eecavanna, line 496: I recommend renaming this list to something like `file_descriptors`. With its current name and without a type hint, I found myself wondering whether it would eventually contain files (versus containing data structures describing files).

web/Dockerfile
```diff
@@ -6,18 +6,11 @@ RUN yarn install
...
-#
-# The short-term solution is to build the nmdc-server/client image off of the last stable version of the
-# nmdc-server/client image (1.5.0). It has nginx 1.22.1 built with mod_zip already. Then we drop in the new
-# frontend code.
-FROM ghcr.io/microbiomedata/nmdc-server/client:1.5.0 as production-stage
+
+FROM nginx:1.28-alpine as production-stage
 LABEL org.opencontainers.image.source=https://github.com/microbiomedata/nmdc-server
 
+RUN rm -v /etc/nginx/nginx.conf
```
eecavanna, line 13: I was surprised to see this line (i.e. `RUN` command) be added in this PR. Is the basic idea that you want to prevent any default configuration file that might be part of the `nginx:1.28-alpine` from coming into play? Maybe there is history related to this that I'm forgetting about.

eecavanna (approved):
Thanks a lot for implementing this, and explaining the situation so clearly in the PR description. I'm comfortable with this branch being merged into `main` once ZipStreamer has been deployed to the development stack on Spin. I will contact you via Slack about this now, @naglepuff.

eecavanna:
> Note that we will need to update the Rancher deployments to include this new service so that zip download works when this PR lands.

I deployed an instance of ZipStreamer to the `nmdc-dev` namespace today. The Kubernetes deployment's name is `portal-zipstreamer`. The Kubernetes service that sits in front of it is also named `portal-zipstreamer`; so other containers in the same Kubernetes namespace can now access that ZipStreamer instance at `http://portal-zipstreamer:4008` (the service name is used as the hostname). There is no ingress pointing to that service, so arbitrary computers on the public Internet cannot access that ZipStreamer instance.

I also set the following environment variable on the `portal-backend` deployment in the same namespace:
```
NMDC_ZIP_STREAMER_URL=http://portal-zipstreamer:4008/download
```

Now that that is done, I am comfortable with this branch being merged into `main`.

tests/conftest.py
```diff
@@ -39,6 +40,14 @@ def mock_not_implemented(*args, **kwargs):
     monkeypatch.setattr(nmdc_geoloc_tools, "landuse_dates", mock_not_implemented)
 
 
+@pytest.fixture()
+def patch_zip_stream_service(monkeypatch):
+    def mock_zip_streamer(*args, **kwargs):
+        yield b"foo"
+
+    monkeypatch.setattr(nmdc_server.api, "stream_zip_archive", mock_zip_streamer)
```
danlamanna, line 48: Is zipstreamer running in testing? If so, should it be mocked or should the test run and assert that the result is a valid zip file?

nmdc_server/api.py
```diff
@@ -682,6 +683,16 @@ async def get_data_object_aggregation(
...
+    settings = Settings()
+    async with httpx.AsyncClient() as client:
+        async with client.stream(
+            "POST", settings.zip_streamer_url, json=zip_file_descriptor
+        ) as response:
+            async for chunk in response.aiter_bytes(chunk_size=settings.zip_streamer_chunk_size):
+                yield chunk
```
danlamanna:
```suggestion
    async with (
        httpx.AsyncClient() as client,
        client.stream("POST", settings.zip_streamer_url, json=zip_file_descriptor) as response,
    ):
        async for chunk in response.aiter_bytes(chunk_size=settings.zip_streamer_chunk_size):
            yield chunk
```

nmdc_server/config.py
```diff
@@ -41,6 +41,10 @@ class Settings(BaseSettings):
     orcid_client_secret: str = "oauth secret key"
     orcid_authorize_scope: str = "/authenticate"
 
+    # for zipstreamer
+    zip_streamer_url: str = "http://zipstreamer:4008/download"
+    zip_streamer_chunk_size_bytes: int = 512
```
danlamanna, line 46: I think this should probably be somewhere between 1-5MB by default.

naglepuff: Updated (fixed up into commit e280bed)

tests/conftest.py
danlamanna: I attempted to set this up and it turned out to be a huge pain because of https://github.com/encode/starlette/issues/1102. I don't think it's worth pursuing at the moment.

web/Dockerfile
naglepuff: I actually don't know the exact origins of this line, but it existed before the changes @pkalita-lbl made to get our front edge image building for mod_zip, so I'm hesitant to remove it. (v1.5.0 below)

https://github.com/microbiomedata/nmdc-server/blob/4b7fe15d70302dab4d71dc2a601d09f01c43d76e/web/Dockerfile#L78

PR #1624, opened 2025-04-29
Allow Field Notes to submit the old submission data model until its next release

### LINKED ISSUE #1622: [HOTFIX] Add shim layer to allow Field Notes to submit data in pre-1.6.0 format
Currently Field Notes is unable to create new studies. This is because it is still submitting the old submission format from before the changes in https://github.com/microbiomedata/nmdc-server/pull/1581 went to production with release 1.6.0. 

The primary issue is the fact that it is still submitting the `alternativeNames`, `GOLDStudyId`, and `NCBIBioProjectId` fields on `multiOmicsForm` instead of on `studyForm`. To be clear the app doesn't actually collect those values, but since the backend requires them, it must send the empty strings/arrays.

Since the lead time in getting Field Notes versions out to users is much longer, I'm proposing we add a small shim layer in the backend to address this. This would be similar to what we did to ease the `packageName` transition from strings to arrays of strings: https://github.com/microbiomedata/nmdc-server/pull/1491 (in fact, that shim can now be safely removed).

pkalita-lbl:
Fixes #1622 

These changes exchange one Field Notes-related shim with another. The old one (regarding `packageName` as either a string or array of strings) is no longer needed because Field Notes has caught up with that change.

The new one allows Field Notes to continue providing the `GOLDStudyId`, `NCBIBioProjectId`, and `alternativeNames` fields in `multiOmicsForm`, as opposed to `studyForm`. Field Notes doesn't actually collect these values, but the backend enforces that at least empty strings/arrays are provided at creation time. That is why this shim (like the last one) is only needed in the `POST /api/metadata_submission` handler.

This is ready for review now. I'm keeping it as a draft to remind myself to make a hotfix tag from this branch before merging it into `main`.

### DIFF
--- a/nmdc_server/api.py
+++ b/nmdc_server/api.py
@@ -1227,13 +1227,24 @@ async def submit_metadata(
     db: Session = Depends(get_db),
     user: models.User = Depends(get_current_user),
 ):
-    # Old versions of the Field Notes app will continue to send a string for packageName
-    # for a little while. This code is to ease that transition and can be removed in the future.
-    if isinstance(body.metadata_submission.packageName, str):
-        if body.metadata_submission.packageName:
-            body.metadata_submission.packageName = [body.metadata_submission.packageName]
-        else:
-            body.metadata_submission.packageName = []
+    # Old versions of the Field Notes app will continue to use the pre-v1.6.0 submission format
+    # (before certain fields were moved from the multi-omics form to the study form) until its
+    # next release. This is a temporary shim layer that can be removed later.
+    multi_omics_form_extras = body.metadata_submission.multiOmicsForm.__pydantic_extra__ or {}
+    if body.metadata_submission.studyForm.GOLDStudyId is None:
+        body.metadata_submission.studyForm.GOLDStudyId = multi_omics_form_extras.get(
+            "GOLDStudyId", ""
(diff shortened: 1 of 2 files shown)


eecavanna (approved):
This all looks good to me. Thanks for taking care of this.

There is one detail I want to document here (in the PR, not necessarily as a code change/comment) regarding the removal of past shims, which is that users of the Field Notes app may not necessarily update to the latest version right away and I don't know that the app has the infrastructure in place yet to detect client/server version mismatches. We can discuss that as the Field Notes squad outside of this PR (I think we've touched on the topic at least once in the past).

pkalita-lbl:
> users of the Field Notes app may not necessarily update to the latest version right away

Yeah we should always keep that in mind. In the extreme case, it's not even a matter of "right away" but potentially "at all". It's certainly possible that some users could have automatic updates disabled. So at the end of the day, it's probably up to us to declare which versions of the app are no longer "supported", give people a reasonable amount of time to upgrade, but then have to say sorry to anyone trying to continue using an unsupported version.

In practical terms, for this change I know from looking at analytics that all usage in the last 30 days have been from versions that do not require the old backend shim (the one being removed here).

PR #1625, opened 2025-04-30
Make ended_at_time nullable

### LINKED ISSUE #1618: Change nullability of `ended_at_time` for to match NMDC Schema
Discovered during April 2025 release.

On `WorkflowExecution`s, the `ended_at_time` slot is not required by [NMDC schema](https://microbiomedata.github.io/nmdc-schema/WorkflowExecution/).

However, the corresponding model in `nmdc-server` has this column defined with `nullable=False`.

https://github.com/microbiomedata/nmdc-server/blob/b0875fab6f7ca39e2d2483ae6a100668f813a19f/nmdc_server/models.py#L499

This causes records with no `ended_at_time` set to be skipped during ingest.

naglepuff:
Fix #1618 

This is needed to support data from the JAWS rollout.

@aclum


naglepuff:
Before merge: please note that this branch is based off of tag `v1.6.2`, and is intended to be able to be hotfixed into production. I would like @aclum's input on whether or not this should be a hotfix.

eecavanna (approved):
LGTM. I reviewed the migrations and the changes to the corresponding classes.

marySalvi (approved):

Issue #1484, opened 2024-12-05
Data model - remove "used" columns and "best_protein" column


naglepuff:
Spun out of #1331 

This slot is absent from NMDC Schema (I believe starting with 11.0), but a nullable column in the SQL tables. It should be removed from:

- [ ] `MetabolomicsAnalysis`
- [ ] `NOMAnalysis`

naglepuff:
Additionally, remove the `best_protein` column from `metap_gene_function_annotation`.

PR #1634, opened 2025-05-08
Remove columns from data model

yxu-lanl:
Closes https://github.com/microbiomedata/nmdc-server/issues/1484

Remove "used" columns from fakes.py, models.py and schemas.py
Remove "best_protein" column from models.py


yxu-lanl:
@naglepuff How to test the changes? Thanks.

naglepuff:
Thanks for taking this on! Postgres data model changes have a bit more work to them. You need to make an [alembic](https://alembic.sqlalchemy.org/en/latest/tutorial.html) migration. These are essentially python scripts that can be run to sync up the postgres database(s) with the model definitions in [models.py](https://github.com/microbiomedata/nmdc-server/blob/main/nmdc_server/models.py). These are run automatically at startup, and on ingest.

For the most part, we stick to automatically generating migration files via:

`docker compose run backend alembic -c nmdc_server/alembic.ini revision --autogenerate`

This will create a new migration file. The `-c nmdc_server/alembic.ini` flag tells `alembic` where to look for the config. Take a look at that file for more details.

Since this command is run as `docker compose run ...`, you might want to change the owner of the new file to yourself, otherwise you might not be able to save/edit the file without `sudo` permissions. You might not have an issue, but this is a "gotcha" I've run into before.

Once you create the new file (which can be found at `nmdc_server/migrations/versions`) you'll need to rename it (following the pattern of existing migrations) and add a description to the top of the file. Refer to existing migration files for examples. You might also need to format it (`tox -e format`) for all CI tests to pass. Let me know if you have any other questions! If I left anything out it might also appear [here](https://github.com/microbiomedata/nmdc-server/blob/main/docs/development.md#generating-new-migrations).

yxu-lanl:
I formatted the new migration file with 'tox -e format'. But the server check still failed.  Could you please take a look at the new migration file? Thanks.

naglepuff:
@yxu-lanl it looks like linting and typechecking are also failing. You can inspect the output of the last CI run, or try

`tox -e lint` and `tox -e typecheck` locally.

yxu-lanl:
Fixed. Thank you.

naglepuff (commented):
Code changes look good! I do have one more request before merge, and that's to clean up the git history a little bit - generally I think its best if we try to keep it organized so that its easier to review the history when we need to come back and look at changes that were made. In this case I'd recommend doing an [interactive rebase](https://git-scm.com/docs/git-rebase) to "fixup" some of the formatting-only commits.

In case you aren't familiar with the process you can start by generating a to-do list for the rebase with `git rebase -i HEAD~6`. On this branch at the time of writing that will look like this:

```
pick 1720bdb Generate migration file
pick ab3d129 Format new migration file
pick a41a5e5 Format new migration file
pick 8049b89 Format new migration file
pick 338422f Format new migration file
pick 4c24581 Fix linting and typechecking errors
```

Each line is a commit, and the first work in the line is the action that the rebase will take. In this case, we're editing the history of all commits from 1720bdb to 4c24581.

I like to use "fixup" to handle these sorts of cases where you have multiple commits that serve one purpose. Fixup will basically take the changes in a commit, and merge them with the changes in the previous commit, and keep the message of the previous commit.

Since all of the commits with the message "Format new migration file" are really just touching up the changes in the "Generate migration file" commit, I would suggest fixing up all of those commits. To do that, you would edit the to-do list to look like this:

```
pick 1720bdb Generate migration file
fixup ab3d129 Format new migration file
fixup a41a5e5 Format new migration file
fixup 8049b89 Format new migration file
fixup 338422f Format new migration file
pick 4c24581 Fix linting and typechecking errors
```

Once you save and exit that file, git will re-run those commits with the new action, and you should be left with a cleaner git history. Give it a shot and let me know how it goes!

yxu-lanl:
How to push the new local git history to repo?

naglepuff:
> How to push the new local git history to repo?

`git push -f` should do the trick (`-f` is the "force" flag)

yxu-lanl:
Thanks for sharing your expertise and experience with me.

PR #1639, opened 2025-05-14
Fix name and subtext in submission stepper

JamesTessmer:
Currently step 4 on the stepper is labeled 'Environment package' with the subtext 'choose package type'.
This no longer matches the page's heading and subtext, so this PR changes the text in the stepper to match.
There is no related issue for this change.

Before:
Screenshot 2025-05-14 at 10 29 15 AM

After:
Screenshot 2025-05-14 at 10 26 52 AM

### DIFF
--- a/web/src/plugins/router.ts
+++ b/web/src/plugins/router.ts
@@ -77,7 +77,7 @@ const router = new VueRouter({
               component: MultiOmicsDataForm,
             },
             {
-              name: 'Environment Package',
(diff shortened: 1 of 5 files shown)


eecavanna:
I don't know about the real-world context (i.e. who wants what to say what). In terms of the code changes, those all look good to me. 👍

mslarae13 (approved):

PR #1643, opened 2025-05-15
Create filter class for searching metap workflows

naglepuff:
Looks like I goofed pretty hard here lmao


marySalvi (approved):

nmdc_server/query.py
```diff
@@ -433,7 +437,8 @@ def execute(self, db: Session) -> Query:
         """Search for entities in the target table."""
         model = self.table.model
         subquery = self.query(db).subquery().alias("id_filter")
-        return db.query(model).join(subquery, model.id == subquery.c.id)  # type: ignore
+        to_return = db.query(model).join(subquery, model.id == subquery.c.id)  # type: ignore
+        return to_return
```
marySalvi, line 441: Just curious why this change here? Were you getting a linting error for having a type ignore in a return?

naglepuff: Thanks, I probably captured the final query so I could print it out more easily (the debug SQL logging may have been extra noisy), and I forgot to simplify this when removing all that debugging code. I've gone ahead and just simply returned the query.

PR #1648, opened 2025-05-16
Force no timeout for zipstreamer

naglepuff:
This could be a workaround for some of the zipstreamer issues we've been seeing in dev.

Some error logs showed a `httpx.ReadTimeout` error. This change will stop `httpx` from throwing these errors.


eecavanna:
Thanks for implementing this. I'll merge this PR in once the checks pass (running now).

PR #1649, opened 2025-05-16
Log a message when zipstreamer seems starved

naglepuff:
I'm not exactly sure what kinds of messages would be useful to have for zipstreamer monitoring. I think logging each chunk is overkill. This change will log a message if it takes zipstreamer more than 5 seconds to send a chunk.

This is related to #1648 in that the number chosen for this change is the [default timeout for HTTPX](https://www.python-httpx.org/advanced/timeouts/). Whether or not this kind of logging proves useful is yet to be seen, so we should revisit these changes after resolving the zipstreamer issues and find more intentional fixes to the issues that we've been seeing.

### DIFF
--- a/nmdc_server/api.py
+++ b/nmdc_server/api.py
@@ -1,6 +1,7 @@
 import csv
 import json
 import logging
+import time
 from io import BytesIO, StringIO
 from typing import Any, Dict, List, Optional, Union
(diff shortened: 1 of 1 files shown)


eecavanna (approved):
This will make it easier to see slow chunks at a glance.

PR #1651, opened 2025-05-19
Allow specifying a host for NERSC data

naglepuff:
cc @shreddd 

Here is my idea for specifying an alternative host for data hosted at NERSC. We can use this to see if it makes any difference with zip download performance with zipstreamer.

### DIFF
--- a/nmdc_server/config.py
+++ b/nmdc_server/config.py
@@ -44,6 +44,7 @@ class Settings(BaseSettings):
     # for zipstreamer
     zip_streamer_url: str = "http://zipstreamer:4008/download"
     zip_streamer_chunk_size_bytes: int = 2 * 1024 * 1024
+    zip_streamer_nersc_data_host: str = "https://data.microbiomedata.org/data"
 
     @property
     def orcid_openid_config_url(self) -> str:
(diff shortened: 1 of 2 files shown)


nmdc_server/crud.py
```diff
@@ -485,6 +486,14 @@ def create_bulk_download(
         raise
 
 
+def replace_nersc_data_host(url: str):
```
eecavanna, line 489: I suggest adding a docstring. Here's an example:

```suggestion
def replace_nersc_data_host(url: str) -> str:
    """
    Updates NERSC URLs so they have the custom prefix defined in
    an environment variable. This can be used to optimize the URLs
    for HTTP clients that have direct access to the NERSC network.
    """
```

eecavanna (approved):
LGTM! I added a suggestion about a docstring (and return type). Not a deal-breaker, given the pace of things today.

nmdc_server/config.py
```diff
@@ -44,6 +44,7 @@ class Settings(BaseSettings):
     # for zipstreamer
     zip_streamer_url: str = "http://zipstreamer:4008/download"
     zip_streamer_chunk_size_bytes: int = 2 * 1024 * 1024
+    zip_streamer_nersc_data_host: str = "https://data.microbiomedata.org/data"
```
eecavanna, line 47: I suggest calling it a "base URL" since it includes more than the host portion.

```suggestion
    zip_streamer_nersc_data_base_url: str = "https://data.microbiomedata.org/data"
```

Note: I am making this suggestion without making the corresponding suggestion in the function that consumes this environment variable.

naglepuff: Works for me. I've changed it in both places with f777517

eecavanna:
Thanks — LGTM! I'm OK with merging now.

Issue #1579, opened 2025-03-26
Update `require` statements in client-side code use impot/export syntax


naglepuff:
We currently have some light mixing of ES6 style `import`/`export` syntax and common JS `require`. I'm specifically referring to the use of images for the study detail page:

https://github.com/microbiomedata/nmdc-server/blob/da8bb307b45492fe971f05c62d5d0d5ee7b60957/web/src/views/IndividualResults/StudyPage.vue#L222-L229

And the biosample detail page (see PR #1577).

I think it would be good to favor one style of imports over the other. The vast majority of our code base uses `import`/`export`, so we should update to statements like:

```ts
import gold from @/assets/gold.png
```

In order to appease typescript, we might also need to add a type declaration file so it knows how to treat these imports. Something like: https://stackoverflow.com/a/54137233

PR #1653, opened 2025-05-21
Use import/export syntax for require statements

yxu-lanl:
Closes https://github.com/microbiomedata/nmdc-server/issues/1579

Add module "*.png"  to shims-vue.d.ts
Replace require('@/assets/.png') with 'import' in AttributeList.vue and StudyPage.vue


naglepuff (approved):
LGTM! Thanks for taking care of this

PR #1659, opened 2025-05-29
Define submissionStatus in nmdc-server via schema enum

### LINKED ISSUE #1658: Define submissionStatus in nmdc-server via schema enum

samobermiller:
Closes #1658 

To expand submission status and its functionality on the submission portal, this PR takes the new submissionStatusEnum and uses it to define submissionStatus in the nmdc-server repo.  
- Replaces all definition and direct strings used in repo with reference to enum
- Creates migrator to update existing status in postgres to fit enum (`in-progress` to `In Progress`, `Submitted- Pending Review` to `Submitted - Pending Review`, `Complete` to `Released`)

Notes: Does not change any underlying functionality with `status` yet. In course of testing these changes, discovered that any unknown status is reverted to "In Progress", which might not be preferable to generating an error. Issue will be created for this but not addressed in this PR


web/src/views/SubmissionPortal/store/index.ts
```diff
@@ -47,13 +50,12 @@ const permissionLevelHierarchy: Record<PermissionLevelValues, number> = {
...
 };
 
-const submissionStatus: Record<string, SubmissionStatus> = {
-  InProgress: 'In Progress',
-  SubmittedPendingReview: 'Submitted- Pending Review',
-  Complete: 'Complete',
-};
+//use schema enum to define submission status
+const submissionStatus: Record<SubmissionStatusKey, SubmissionStatusTitle> = Object.fromEntries(
+  Object.entries(NmdcSchema.enums.SubmissionStatusEnum.permissible_values).map(([key, item]: [SubmissionStatusKey, SubmissionStatusTitle]) => [key, item.title]),
```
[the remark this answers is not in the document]
eecavanna: I don't quite follow what Copilot is pointing out here. I'll come back to this.

nmdc_server/models.py
```diff
@@ -34,6 +35,15 @@
 # described by https://microbiomedata.github.io/nmdc-schema/.
 
 
+def get_submission_status_enum():
+    """Get SubmissionStatusEnum from the NMDC schema definition."""
+    schema = get_nmdc_schema_definition()
+    return schema.enums["SubmissionStatusEnum"].permissible_values
```
eecavanna, line 41: When I open this file in my IDE (VS Code), my IDE shows a type-related warning.

image

That reminded me that there is some code in `nmdc-server`, where someone is getting all enums from the schema. That is here:

https://github.com/microbiomedata/nmdc-server/blob/ccb5caac3f3e6496ce0295c7289aee4842bddbf2/nmdc_server/api.py#L900-L902

My IDE does not show any type-related warnings on the latter code. I am wondering whether this function could be rewritten to use the latter approach (i.e. accessing the schema via a [SchemaView](https://linkml.io/linkml/developers/schemaview.html) instance instead of "directly").

From the [LinkML documentation](https://linkml.io/linkml/developers/schemaview.html):
> The `SchemaView` class in the `linkml-runtime` provides a method for dynamically introspecting and manipulating schemas.

Here's what that might look like:

```py
def get_submission_status_enum():
    """Get SubmissionStatusEnum permissible values from the NMDC schema definition."""

    permissible_values = {}
    submission_schema_files = resources.files("nmdc_submission_schema")
    schema_path = submission_schema_files / "schema/nmdc_submission_schema.yaml"
    sv = SchemaView(str(schema_path))
    for enum_name, enum_definition in sv.all_enums().items():
        if enum_name == "SubmissionStatusEnum":
            permissible_values = enum_definition.permissible_values
            break
    return permissible_values
```

tests/test_submission.py
```diff
@@ -11,6 +12,15 @@
 from nmdc_server.schemas_submission import SubmissionMetadataSchema, SubmissionMetadataSchemaPatch
 
 
+def get_submission_status_enum():
+    """Get SubmissionStatusEnum from the NMDC schema definition."""
+    schema = get_nmdc_schema_definition()
+    return schema.enums["SubmissionStatusEnum"].permissible_values
```
eecavanna, line 18: I think this is a duplicate snippet from `nmdc_server/models.py`. I recommend consolidating.

eecavanna:
Thanks for making those changes, @samobermiller! I think updating `nmdc-server` to use enum values (as you have done here) instead of string literals will contribute to making our system consistent with itself (i.e. smaller chance that a developer making a typo will lead to invalid status values in the database).

naglepuff (approved):
This looks good to me! Everything that we discussed last week has been addressed. The only thing that gives me pause is the use of the "magic" strings "InProgress" and "SubmittedPendingReview".

It would be awesome if we could get these strings in some other way. I think ideally if 
`SubmissionStatusEnum.InProgress.title` was equivalent to `SubmissionStatusEnum["InProgress"].title`, then we'd use that, but I don't know enough about what `SubmissionStatusEnum` looks like. I think its fine as-is though.

pkalita-lbl (approved):
Looks good!

Two notes:

1. Let's hold off on merging this until after Release 2025.7 is deployed next week.
2. Pretty soon (see: https://github.com/microbiomedata/nmdc-schema/pull/2582) we'll be able to import the `nmdc-schema` definition as a JSON file (as opposed to the YAML file). This will allow us to tighten up the type checking in front-end code because TypeScript can statically analyze JSON imports.

Issue #1623, opened 2025-04-29
Data Portal - map rendering for issues for arctic/antarctic samples


aclum:
We are trying to bring in some sea ice samples, @pkalita-lbl loaded these into the dev data portal and they are not rendering correctly. It introduces gray area at the top and puts samples above 85° at 85° (see image below). Patrick said this is b/c web-based Mercator projections generally cut off around ±85°. and these samples are from around 88° and that it is [mathematically impossible](https://en.wikipedia.org/wiki/Web_Mercator_projection#Formulas) for this type of projection to show the poles.

![Image](https://github.com/user-attachments/assets/ec7f7aa1-2924-4876-b92d-6db287fa6aba)

The suggestion was to remove these samples from displaying in the map and update the tool tip to indicate that arctic/antarctic samples won't be shown on the map.

Please comment on if you agree w/Patrick's assessment and suggested path forward.

cc @mslarae13 

We'd like this to be a target for the May monthly release.

pkalita-lbl:
The [SOTO tool](https://soto.podaac.earthdatacloud.nasa.gov/) provided by NASA uses this little widget that lets you switch between default ([EPSG:4326](https://epsg.io/4326)), arctic ([EPSG:3414](https://epsg.io/3413)), and antarctic ([EPSG:3031](https://epsg.io/3031)) projections:

![Image](https://github.com/user-attachments/assets/d0561b22-2d64-4d19-9c59-ed2cdb67c4cb)

So maybe that's an idea if we're really **_really_** committed to showing arctic/antactic samples on _a_ map. But you still wouldn't see them all on the _same_ map. _And_ that would not be a small amount of work; we should not block loading arctic samples on it.

I still think the best path forward for the May timeline is to filter out these samples from the map and add a note to the help popup saying that you can still search for them using the regular search facts.

naglepuff:
@aclum I think @pkalita-lbl's suggestion is a good start

marySalvi:
@aclum 
I spoke with our map expert. @pkalita-lbl  is correct, the issue is that the coordinates are 'out of scope' so to speak. The tool that we use for the map, Leaflet, does support other projections that would include the poles however the tile source we are using, OSM does not have tiles for the poles. This means that if we update the projection but not the tile source, we will have a  white 'hole' instead of mapping for the poles.

I agree that removing the artic data from the map is the correct decision for the short term. We can explore some options for tile sources and projections in the long term.

aclum:
[...]

PR #1663, opened 2025-06-02
1623 map rendering

marySalvi:
Addresses first step in resolving #1623 

This sets the bounds for the map and updates the tooltip to inform user of potentially missing samples from the map-vis

### DIFF
--- a/web/src/components/ClusterMap.vue
+++ b/web/src/components/ClusterMap.vue
@@ -174,6 +174,7 @@ export default defineComponent({
     </v-btn>
     <l-map
       ref="mapRef"
+      :max-bounds="[[[-85, -180], [85, 180]]]"
       :style="{
         height: `${height}px`,
         width: '100%',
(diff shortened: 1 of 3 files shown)


marySalvi:
While troubleshooting a bug I came across a potentially good solution to include all data on the map without 'breaking' the map. We can set the max bounds for the map which will prevent the grey bar but still include the data from the pole:


### This includes all data:

[Screencast from 06-02-2025 01:25:15 PM.webm](https://github.com/user-attachments/assets/3f417ed0-4382-4a9e-a0ab-57b67406b904)


### This has data out of scope of the projection filtered out:

[Screencast from 06-02-2025 03:46:04 PM.webm](https://github.com/user-attachments/assets/3e1d4f61-c5bf-45bb-ad31-15650ec03b7f)




@aclum and @shreddd  do you have a preference? I am personally leaning towards no longer filtering out the data because I think it is a suitable option vis-wise while including as much data as possible.

marySalvi:
### updated map tooltip:

Screenshot from 2025-06-02 14-01-43

Issue #1266, opened 2024-05-25
Update `/api/stats` endpoint to return additional metrics data


eecavanna:
#### Tasks

- [x] Determine what the database queries were that people ran manually in order to provide metrics data to the leadership team for a recent report. At least some of the data pertained to downloads. I want to know what the queries are, anyway, so that I can run them in the future and so that people can consider them when interpreting the metrics data.
- [x] Assess implications of running those queries "on demand" when a client visits the `/api/stats` endpoint. Some of the queries might take too long to run in "real time" like that. As an alternative, some of the queries could be run on a schedule and their results could be stored in the database, so that the API endpoint can serve them without actually performing the full query at that time (it would just fetch the pre-baked result of that full query).
- [x] Either add the data to the response of the `/api/stats` endpoint; or create a private counterpart to the endpoint (e.g. `/api/admin-stats`) that returns that additional data.

**Note:** The assessment (second task) may result in us deciding not to pursue the implementation (third task).

---

#### Current behavior

For reference, here's what the `/api/stats` endpoint returns today:

```json
{
  "studies": 28,
  "locations": 33,
  "habitats": 37,
  "data_size": 77352654942449,
  "metagenomes": 2927,
  "metatranscriptomes": 111,
  "proteomics": 52,
  "metabolomics": 34,
  "lipodomics": 0,
  "organic_matter_characterization": 2767
}
```

eecavanna:
Here's the SQL query we submit to the Postgres database in order to get the number of user accounts that exist:
```sql
SELECT count(*)
FROM (
    SELECT DISTINCT(orcid)
    FROM user_logins
) as distinct_orcids;

-- Note: That happens to produce the same result as... 
--   SELECT count(*)
--   FROM user_logins;
```

ssarrafan:
Removing from sprint since nobody is assigned to this. @eecavanna

eecavanna:
A PR for sub-issue https://github.com/microbiomedata/nmdc-server/issues/1509 is awaiting review. I will move this meta issue to the next sprint, during which I'll think about the other sub-issues.

PR #1665, opened 2025-06-03
Implement `/api/admin/stats` endpoint that returns number of user accounts

eecavanna:
On this branch, I implemented a new API endpoint: `GET /api/admin/stats`. The API endpoint is accessible to admins only. It returns a JSON object that currently has a single property: `num_user_accounts`.

The API endpoint was designed to be used by people that gather metrics for reports (the process they currently use involves going to Rancher to determine which Postgres database is "active," and then using a SQL client to query that database). Here is the SQL query they currently run (it is equivalent to the one that this endpoint runs):

```sql
SELECT count(*)
FROM (
    SELECT DISTINCT(orcid)
    FROM user_logins
) as distinct_orcids;
```

### DIFF
--- a/nmdc_server/aggregations.py
+++ b/nmdc_server/aggregations.py
@@ -145,6 +145,21 @@ def get_aggregation_summary(db: Session):
     )
 
 
+def get_admin_stats(
+    db: Session,
+) -> schemas.AdminStats:
+    r"""
+    Compiles statistics designed to be consumed by Data Portal/Submission Portal administrators.
+    """
+
+    distinct_orcids_subquery = db.query(func.distinct(models.User.orcid)).subquery()
+    num_distinct_orcids = db.query(func.count()).select_from(distinct_orcids_subquery).scalar()
+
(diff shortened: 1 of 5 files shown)


eecavanna:
Note to reviewers: I made up the `/api/admin` prefix for the URL path. I don't know whether that is compatible with (or in conflict with) any plans any of y'all had in mind for an admin-only endpoint prefix. I can imagine the application having some admin-only pages someday (e.g. an admin-only dashboard that displays various metrics).

eecavanna:
Per the first checkbox in the description of ticket https://github.com/microbiomedata/nmdc-server/issues/1266, one of our team members (e.g. me) may eventually pursue adding a second field to this endpoint's response; maybe something related to data downloads performed by users.

eecavanna:
I'll go ahead and merge this into `main` now, which will cause it to be deployed to dev. The new endpoint will appear in a new section on Swagger UI, called "administration."

PR #1666, opened 2025-06-10
Update `GET /api/metadata_submission/report` endpoint to include the number of samples in each submission

hesspnnl:
Update the get_metadata_submissions_report endpoint to include the number of samples for each submission. Add a column to the TSV called "Number of Samples".

### DIFF
--- a/nmdc_server/api.py
+++ b/nmdc_server/api.py
@@ -1000,9 +1000,9 @@ async def get_metadata_submissions_report(
         # Note: `metadata["sampleData"]` is a dictionary where keys are sample types
         #       and values are lists of samples of that type.
         # Reference: https://microbiomedata.github.io/submission-schema/SampleData/
-        samples = metadata["sampleData"]
-        for samp in samples:
-            sample_count += len(samples[samp])
+        sample_data = metadata["sampleData"]
+        for sample_type in sample_data:
+            sample_count += len(sample_data[sample_type])
 
         author_user = s.author  # note: `s.author` is a `models.User` instance
         study_form = metadata["studyForm"] if "studyForm" in metadata else {}


tests/test_submission.py
```diff
@@ -169,14 +169,140 @@ def test_get_metadata_submissions_report_as_admin(
...
+                            'Gardner, Walter H. "Water content." Methods of Soil Analysis: '
+                            "Part 1 Physical and Mineralogical Methods 5 (1986): 493-544."
+                        ),
+                    },
+                ],
+                "jgi_mg_data": [
+                    {"samp_name": "June2016WEW_Plot6_D2", "analysis_type": ["metagenomics"]},
+                    {"samp_name": "Aug2016WEW_Plot6_D5", "analysis_type": ["metagenomics"]},
+                ],
+            },
```
eecavanna, line 253: I like that this includes multiple sample types.

tests/test_submission.py
```diff
@@ -169,14 +169,140 @@ def test_get_metadata_submissions_report_as_admin(
         author_orcid=other_user.orcid,
         created=now + timedelta(seconds=1),
         date_last_modified=datetime.utcnow(),
-        is_test_submission=True,
         metadata_submission={
```
eecavanna, line 172: I think this test data can be distilled to only the parts that are required by the submission schema and that impact the functionality of this test. I suggest adding a `TODO` comment about it.

```suggestion
        # TODO: Omit some optional fields in order to simplify the test data.
        # See: `class MetadataSubmissionRecordCreate` in `schema_submission.py`
        # See: https://microbiomedata.github.io/submission-schema/SampleData/
        metadata_submission={
```

tests/test_submission.py
```diff
@@ -232,6 +360,7 @@ def test_get_metadata_submissions_report_as_admin(
     assert data_row["Source Client"] == ""  # upstream faker lacks `source_client` attribute
     assert data_row["Status"] == "In Progress"  # matches value in upstream faker
     assert data_row["Is Test Submission"] == "False"
+    assert data_row["Number of Samples"] == "0"
```
eecavanna, line 363: I like that both a non-zero case and a zero case are demonstrated in this test!

hesspnnl: thanks!

hesspnnl:
> Thanks for implementing this! It looks good to me and I feel comfortable with this being merged in as is.
> 
> I left a few suggestions, all of which I consider to be optional. Their subjects include comments, variable names, and distilling an example dictionary down.

I like the suggestions!

eecavanna:
LGTM! Merging into `main` now, which (like in `nmdc-runtime`) will trigger a deployment of `main` to our Spin-hosted development environment, where our team members will be able to test the new behavior before the next monthly release.

FYI: @mslarae13

PR #1667, opened 2025-06-10
Don't import worksheets if not selected

### LINKED ISSUE #947: Deleting from main tab doesn't erase EMSL and JGI tabs, but you can't see it
Deleting content from the soil tab doesn't delete it from the JGI or EMSL tabs, but you can't see the rows, so you can't remove them

Replicate
1: start a submission, data has not been generated, JGI and EMSL
2: In DH interface, import this data file: https://docs.google.com/spreadsheets/d/11kG8Y-q9MS6EPw-kqHkZf655afyDfIO3/edit?usp=share_link&ouid=113791678009149106423&rtpof=true&sd=true 
3: validate
4: check that information is in the EMSL and JGI tabs... add some information to the columns you can edit
5: go to soil tab and delete all rows
6: go to EMSL and JGI tab, nothing is ther
7: go back to Multiomics data or environmental package, see it says there's still information in the DH sheet

Resolution
- If metadata samples or rows are removed from the soils tab, they should remove from the EMLS and JGI tab. Not just become invisible

See this video : https://drive.google.com/file/d/1nCoG02Oc42jns8B12imjxl1V_AePw9TN/view?usp=sharing
~ 8:30 to 9:10
or
~11-12:40

naglepuff:
Fix #947

### Changes
(Use this spreadsheet for testing: [nmdc_sample_export.xlsx](https://github.com/user-attachments/files/20679377/nmdc_sample_export.xlsx))

While investigating the linked issue, I noticed 2 behaviors that this PR seeks to change.

#### Importing unselected template data
When a user fills out a submission, the templates (tabs) that show up on the data harmonizer page are determined by the choices made at earlier steps. If a user's submission information results in tabs `A, B, C` showing up on the data harmonizer page, and imports a spreadsheet file that has worksheets for `A, B, C, D`, all of those templates would be imported, resulting in  sample data on the submission that was hidden to the user. Since template choice is locked in when any sample data is populated, users wouldn't be able to go back and add a tab to see the data.

Now, on import, if a worksheet does not match a currently selected template, that sheet will not be imported, and the toast message will alert the user that some worksheet was not recognized.

#### Responding to deleted rows
Previously, if I use the data harmonizer/hands on table feature of deleting rows through the right click context menu, I'd get some unintuitive behavior. For example, if I import a spreadsheet, then delete all rows and immediately refresh, my data would magically return. 

Now, we can hook into the hands on table `afterRemoveRow` event to update the state of our application, and delete the rows from `sampleData` right away. This deletion also handles the case where there is "hidden" sample data, which can exist due to the workflow described in the previous section. This change means that if any existing submissions have hidden data, it should be deleted when the corresponding rows in the environment tab are deleted.

### DIFF
--- a/web/src/views/SubmissionPortal/HarmonizerView.vue
+++ b/web/src/views/SubmissionPortal/HarmonizerView.vue
@@ -205,8 +205,139 @@ export default defineComponent({
       }
     });
 
+    function rowIsVisibleForTemplate(row: Record<string, any>, templateKey: string) {
+      const environmentKeys = templateList.value.filter((t) => HARMONIZER_TEMPLATES[t].status === 'published');
+      if (environmentKeys.includes(templateKey)) {
+        return true;
+      }
+      const row_types = row[TYPE_FIELD];
+      if (!row_types) {
+        return false;
+      }
+      if (templateKey === EMSL) {
+        return row_types.includes('lipidomics')
+          || row_types.includes('metaproteomics')
+          || row_types.includes('metabolomics')
+          || row_types.includes('natural organic matter');
+      }
+      if (templateKey === JGI_MG) {
+        return row_types.includes('metagenomics');
+      }
+      if (templateKey === JGI_MG_LR) {
+        return row_types.includes('metagenomics_long_read');
+      }
+      if (templateKey === JGI_MT) {
+        return row_types.includes('metatranscriptomics');
+      }
+      if (templateKey === DATA_MG) {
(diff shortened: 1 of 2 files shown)


naglepuff:
@pkalita-lbl as of 9a8311f, the test cases you outlined above appear to work as expected when looking at the `PATCH` requests.

pkalita-lbl (approved):
Looks good! I didn't find any behavior that I thought was incorrect upon further testing.

PR #1670, opened 2025-06-17
Add Data DOI collection

JamesTessmer:
This PR adds data DOI collection to the study page.

Resolves https://github.com/microbiomedata/submission-schema/issues/329 (and part of https://github.com/microbiomedata/submission-schema/issues/314)

Implementation is similar to contributor and funding source where multiple cards can be added for multiple DOIs.

Current appearance for dataDoi collection:
Screenshot 2025-06-23 at 5 50 34 PM

current appearance for awardDoi collection:
Screenshot 2025-06-23 at 5 51 30 PM

Things left to do:
- update hints


web/src/views/SubmissionPortal/Components/StudyForm.vue
```diff
@@ -48,6 +48,8 @@ export default defineComponent({
       },
     ]);
 
+    const doiProviderValues = Object.keys(NmdcSchema.enums.DoiProviderEnum.permissible_values);
```
pkalita-lbl, line 51: > 3. Updating the values displayed for each enum. I'm not sure what the best way to do that is since they're imported from the schema.

The permissible value definitions (i.e. the _values_ of the `permissible_values` object) has a `title` field that should be used for end-user display. Combine that with the ability of the `items` [prop](https://v2.vuetifyjs.com/en/api/v-select/#props-items) on `<v-select>` to take an array of objects instead of an array of strings. So something like this:

```suggestion
    const doiProviderValues = Object.values(NmdcSchema.enums.DoiProviderEnum.permissible_values)
      .map((pv: any) => ({
        value: pv.text,
        text: pv.title || pv.text, // The permissible value definition should have a title, but in case it doesn't fallback to text (which it will always have)
      }))
      .sort((a, b) => a.text.localeCompare(b.text));
```

web/src/views/SubmissionPortal/Components/StudyForm.vue
```diff
@@ -65,6 +67,20 @@ export default defineComponent({
...
+        studyForm.dataDois = [{
+          value: '',
+          provider: '',
+        }];
+      } else {
+        studyForm.dataDois.push({
+          value: '',
+          provider: '',
+        });
+      }
```
pkalita-lbl, line 81: Minor suggestion to prevent having to repeat the new value being added:

```suggestion
      if (!Array.isArray(studyForm.dataDois)) {
        studyForm.dataDois = [];
      }
      studyForm.dataDois.push({
        value: '',
        provider: '',
      });
```

pkalita-lbl (commented):
Changes look good! 

Regarding your comments:

> 1. Excluding award_doi from the enum for the category selection.

It looks like that has already been done? Since all DOIs collected in this part of the interface are data DOIs I agree with not presenting a dropdown for the category. When this data is translated into a MongoDB record it will be implicit which category should be applied based on the structure of the submission data (i.e. these DOIs live in the `dataDois` field).

> 2. Updating hints for these components ( @mslarae13 if you have any input on this that would help)

Yes, agreed. I'd recommend working with Montana to craft some language that will be understandable to an outside user.

One last general comment is that at some point we'll need to loop back and apply something similar to the Award DOI inputs in the Multi-omics Data screen. I'm not sure if there is a separate ticket for that work, but I almost think that should be done before https://github.com/microbiomedata/submission-schema/issues/330.

web/src/views/SubmissionPortal/Components/MultiOmicsDataForm.vue
```diff
@@ -63,6 +72,9 @@ export default defineComponent({
     }];
     const doiRequiredRules = () => [(v: string) => {
       const valid = !!v || (!multiOmicsForm.facilityGenerated && multiOmicsForm.dataGenerated) || (multiOmicsForm.unknownDoi && !multiOmicsForm.dataGenerated);
+      if (!/^doi:10.\d{2,9}.*$/.test(v)) {
+        return 'Doi must be in the correct format.';
+      }
       return valid || 'An award DOI is required if data was or is going to be generated by a DOE facility.';
     }];
```
pkalita-lbl, line 79: A few issues here.

1. As this is currently implemented this function will return the "Doi must be in the correct format" message instead of the "DOI is required" when the input is blank. The purpose of allowing the `rules` prop of `v-text-field` to be an array of functions is to do multiple independent checks of a value. Currently what we have here is an array with one function that does multiple checks. I think it would be more canonical to have an array of two functions: one that does the check if the value is present and another that does the format check. Check out the example here: https://v2.vuetifyjs.com/en/components/text-fields/#validation. 
2. Let's standardize on "DOI" instead of "Doi"
3. We should make the `doi:` part of the pattern optional here. Users will be extraordinarily confused because 99.99% of the time they will just enter the value starting with "10." The code that translates submission data into `nmdc-schema`-compliant data can add the `doi:` prefix for them.
4. If there's a way to abstract out the pattern check so that it can be reused by `StudyForm.vue` that would be great.

web/src/views/SubmissionPortal/Components/MultiOmicsDataForm.vue
```diff
@@ -375,7 +412,7 @@ export default defineComponent({
         />
       </div>
       <v-btn
-        v-if="multiOmicsForm.dataGenerated || multiOmicsForm.facilities.includes('EMSL') || multiOmicsForm.facilities.includes('JGI')"
+        v-if="multiOmicsForm.facilities.includes('EMSL') || multiOmicsForm.facilities.includes('JGI')"
```
pkalita-lbl, line 415: I can't quite understand why the change in logic here

web/src/views/SubmissionPortal/Components/MultiOmicsDataForm.vue
```diff
@@ -335,35 +348,59 @@ export default defineComponent({
...
+                v-model="multiOmicsForm.awardDois[i].provider"
+                label="Award DOI Provider *"
+                :hint="Definitions.dataDoiProvider"
+                :items="doiProviderValues"
+                persistent-hint
+                outlined
+                dense
+                clearable
+                class="mb-2 mr-3"
+                :error-messages="multiOmicsForm.awardDois[i].provider ? undefined : ['A provider must be selected.']"
```
pkalita-lbl, line 386: Are we using `error-messages` elsewhere? I'm not quite sure why we would use it over `rules`?

web/src/views/SubmissionPortal/Components/StudyForm.vue
```diff
@@ -397,6 +416,78 @@ export default defineComponent({
...
+              label="Data DOI value *"
+              :hint="Definitions.dataDoiValue"
+              persistent-hint
+              outlined
+              dense
+              class="mb-2 mr-3"
+              :error-messages="studyForm.dataDois[i].value ? undefined : ['Field cannot be empty.']"
+              :rules="requiredRules('Field cannot be empty',[
+                v => /^doi:10.\d{2,9}.*$/.test(v) || 'Doi must be valid',
+              ])"
```
pkalita-lbl, line 445: Both `error-messages` and `rules`?? 😱

web/src/views/SubmissionPortal/Components/MultiOmicsDataForm.vue
JamesTessmer: Yeah, I'm not sure why I did that. I think there was a strange behavior I was running into and changed it due to that, but reverting the logic doesn't cause any issues, so I've changed it back.

web/src/views/SubmissionPortal/Components/MultiOmicsDataForm.vue
JamesTessmer: Changed this to rules, after your explanation in the previous comment the use of `rules` makes more sense

web/src/views/SubmissionPortal/Components/StudyForm.vue
JamesTessmer: I certainly made a choice here. I've made better choices now :)

pkalita-lbl (approved):
Looks good!

PR #1675, opened 2025-06-23
Update api/stats/ endpoint to sum file byte size if the DataObject is a WorkflowExecution ouput

hesspnnl:
Added a slot in the response JSON called wfe_output_data_size_bytes that is the summed value of all DataObjects that come from WorkflowExecution processes. Incorporated tests associated with this process.

### DIFF
--- a/nmdc_server/aggregations.py
+++ b/nmdc_server/aggregations.py
@@ -3,7 +3,7 @@ from typing import Any, Dict, List, Type, cast
 from sqlalchemy import Column, func, or_
 from sqlalchemy.orm import Session
 
-from nmdc_server import models, query, schemas
+from nmdc_server import models, query, schemas, table
 from nmdc_server.attribute_units import get_attribute_units
 from nmdc_server.data_object_filters import WorkflowActivityTypeEnum
 
@@ -129,6 +129,19 @@ def get_aggregation_summary(db: Session):
         )
 
         return num_non_parent_studies
+    
+    def count_wfe_output_data_size_bytes() -> int:
+        r"""Returns the total size of WFE output data objects in bytes."""
+        doj_outputs = set()
+        for t in table.workflow_execution_tables:
+            t.model()  # Ensure the model is loaded
+            doj_outputs.update(t.model.has_outputs)
+        sum = (
+            q(func.sum(func.coalesce(models.DataObject.file_size_bytes, 0)))
+            .filter(models.DataObject.id.in_(doj_outputs))
+            .scalar() or 0
+            )
+        return sum
 
     return schemas.AggregationSummary(
         studies=q(models.Study).count(),
@@ -136,6 +149,7 @@ def get_aggregation_summary(db: Session):
         locations=distinct(models.Biosample.annotations["location"]),
         habitats=distinct(models.Biosample.annotations["habitat"]),
         data_size=q(func.sum(func.coalesce(models.DataObject.file_size_bytes, 0))).scalar(),
+        wfe_output_data_size_bytes=count_wfe_output_data_size_bytes(),
         metagenomes=omics_category("metagenome"),
         metatranscriptomes=omics_category("metatranscriptome"),
         proteomics=omics_category("proteomics"),
--- a/nmdc_server/aggregations.py
+++ b/nmdc_server/aggregations.py
@@ -3,7 +3,7 @@ from typing import Any, Dict, List, Type, cast
 from sqlalchemy import Column, func, or_
 from sqlalchemy.orm import Session
 
-from nmdc_server import models, query, schemas, table
+from nmdc_server import models, query, schemas
 from nmdc_server.attribute_units import get_attribute_units
 from nmdc_server.data_object_filters import WorkflowActivityTypeEnum
 
@@ -97,6 +97,20 @@ def get_table_summary(db: Session, model: models.ModelType) -> schemas.TableSumm
     return schemas.TableSummary(total=count, attributes=attributes)
 
 
+def get_all_workflow_data_object_ids(db: Session) -> set:
+    r"""Returns a set of all DataObject IDs that are outputs of any workflow execution model."""
+    doj_outputs = set()
+
+    # Use workflow_activity_types to iterate through all workflow types
+    for workflow_model in models.workflow_activity_types:
+        # For each workflow instance, get its output data objects
+        for instance in db.query(workflow_model).all():
+            # Add each output data object ID to the set using the has_outputs relationship
+            doj_outputs.update(instance.has_outputs)
+
+    return doj_outputs
+
+
 def get_aggregation_summary(db: Session):
     q = db.query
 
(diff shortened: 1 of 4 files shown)


nmdc_server/aggregations.py
```diff
@@ -136,6 +150,12 @@ def count_non_parent_studies() -> int:
         locations=distinct(models.Biosample.annotations["location"]),
         habitats=distinct(models.Biosample.annotations["habitat"]),
         data_size=q(func.sum(func.coalesce(models.DataObject.file_size_bytes, 0))).scalar(),
+        wfe_output_data_size_bytes=(
+            q(func.sum(func.coalesce(models.DataObject.file_size_bytes, 0)))
+            .filter(models.DataObject.id.in_(get_all_workflow_data_object_ids(db)))
```
eecavanna, line 155: I recommend replacing this helper function call with a subquery, so that the overall query is run against a single state of the database (i.e. be atomic).

In theory, if the mapping between the WFEs and the DOs were to change between when the helper function is called and when the overall query uses what the function returned, the result of the overall query could be an inaccurate representation of things at the time the overall query was run.

nmdc_server/models.py
```diff
@@ -488,6 +488,7 @@ def downloads(self) -> int:
 
 # This is a base class for all workflow processing activities.
 # https://microbiomedata.github.io/nmdc-schema/WorkflowExecutionActivity.html
+# TODO : does this exist anymore?
```
eecavanna, line 491: Thanks for adding this!

tests/test_app.py
```diff
@@ -82,13 +82,32 @@ def test_api_faceting(db: Session, client: TestClient):
 
 
 def test_api_summary(db: Session, client: TestClient):
+    """
+    This test checks the /api/summary endpoint to ensure it returns the expected
```
eecavanna, line 86:
```suggestion
    This test checks the `/api/stats` endpoint to ensure it returns the expected
```

tests/test_app.py
```diff
@@ -82,13 +82,32 @@ def test_api_faceting(db: Session, client: TestClient):
...
-        fakes.MetaproteomicAnalysisFactory()
-        fakes.DataObjectFactory()
+        fake_meta_annotation = fakes.MetagenomeAnnotationFactory()
+        fake_meta_assembly = fakes.MetagenomeAssemblyFactory()
+        fake_meta_pro = fakes.MetaproteomicAnalysisFactory()
+        # Commit the objects to the database first to ensure they exist
+        db.commit()
+        fake_meta_annotation.outputs = [data_object]
+        fake_meta_assembly.outputs = [data_object]
+        fake_meta_pro.outputs = [data_object]
```
eecavanna, line 110: I'm wondering whether you can do this assignment via arguments to the constructor (as was done when setting the file size above). I'll check when I'm back at my desk (currently reviewing from phone).

tests/test_app.py
```diff
@@ -82,13 +82,32 @@ def test_api_faceting(db: Session, client: TestClient):
...
+        this tests that data object from different processes are accounted for.
+    - we check that data object are not double counted through connecting the
+        same data object to multiple processes
+    - we then check that the total size of the WFE output data is 55 bytes.
+        this is the sum of the first 10 natural numbers (1 + 2 + ... + 10 = 55).
+    """
     # TODO: This would be better queried against the real data
-    for _ in range(10):
+    for i in range(10):
+        data_object = fakes.DataObjectFactory(file_size_bytes=i + 1)
```
eecavanna, line 101: I recommend also creating one or more `DataObject`s that _aren't_ referenced as outputs of any WFEs, to ensure their file sizes are _not_ included in the sum.

tests/test_app.py
```diff
@@ -82,13 +82,32 @@ def test_api_faceting(db: Session, client: TestClient):
 
 
 def test_api_summary(db: Session, client: TestClient):
+    """
+    This test checks the /api/summary endpoint to ensure it returns the expected
+    summary statistics about the database, including the number of studies,
+
+    wfe data size bytes test case:
```
eecavanna, line 89:
```suggestion
    WFE data size bytes test case:
```

tests/test_app.py
```diff
@@ -82,13 +82,32 @@ def test_api_faceting(db: Session, client: TestClient):
...
+    - here we test that the WFE data size is calculated correctly
+    - we create 10 data objects with increasing file sizes (1 byte to 10 bytes)
+    - we associate these data objects with various workflow execution processes
```
eecavanna, line 92:
```suggestion
    - we associate these data objects with various workflow execution processes.
```

eecavanna (commented):
Looks great! Thanks for implementing this. 

I left a few comments. The only concern I have is about helper function versus subquery. One resolved, I'll approve.

As you mentioned, there is still an outstanding question about how this sum will be presented in the API endpoint's response object.

nmdc_server/aggregations.py
```diff
@@ -1,7 +1,9 @@
 from typing import Any, Dict, List, Type, cast
 
-from sqlalchemy import Column, func, or_
-from sqlalchemy.orm import Session
+from sqlalchemy import Column, func, or_, union_all
+from sqlalchemy.orm import Query, Session
+from sqlalchemy.sql import Alias
+from sqlalchemy.sql import Selectable as Subquery  # fallback if Subquery import fails
```
eecavanna, line 6: If importing `Selectable` here, I'd rather keep using its original name of `Selectable` below (since there is also a SQLAlchemy symbol originally named `Subquery` and I assume they are not 100% interchangeable).

tests/test_app.py
```diff
@@ -106,6 +122,7 @@ def test_api_summary(db: Session, client: TestClient):
     data = resp.json()
     assert data["studies"] == 13
     assert data["non_parent_studies"] == 11  # excludes studies A and B
+    assert data["wfe_output_data_size_bytes"] == 55
```
eecavanna, line 125: I recommend asserting something about the **original, related** field here for contrast (although we may eventually retire or repurpose that field... TBD).

```suggestion
    assert data["data_size"] == 110
    assert data["wfe_output_data_size_bytes"] == 55
```

eecavanna (approved):
Looks good to me! Thanks for implementing this. I think this will make it easier for team members to gather metrics for reports in the future.

I left a couple comments—one about aliasing an imported symbol and one about adding an assertion. Neither is a deal-breaker, from my perspective. 

I do still want to wait until we have heard from @aclum about how to present this metric in the API endpoint's response payload. As a reminder, I think some of our team members will be traveling and in all-day meetings this week, and that may affect response/review times.

tests/test_app.py
```diff
@@ -82,13 +82,29 @@ def test_api_faceting(db: Session, client: TestClient):
...
+    - here we test that the WFE data size is calculated correctly
+    - we create 10 data objects with increasing file sizes (1 byte to 10 bytes)
+    - we associate these data objects with various workflow execution processes.
+        this tests that data object from different processes are accounted for.
+    - we check that data object are not double counted through connecting the
+        same data object to multiple processes
+    - we then check that the total size of the WFE output data is 55 bytes.
+        this is the sum of the first 10 natural numbers (1 + 2 + ... + 10 = 55).
+    """
     # TODO: This would be better queried against the real data
```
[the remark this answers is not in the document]
eecavanna: Another bit of "scope creep" (this can all be deferred to a future PR): I want to eventually make a new test, named `test_api_stats`, and move the `/api/stats` testing out of the current test and into that new test. Then, this existing `test_api_summary` test would only involve the `/api/summary` endpoint.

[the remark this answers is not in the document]
hesspnnl: Should I make a new ticket for these comments?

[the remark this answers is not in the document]
eecavanna: Thanks for suggesting that. I am in favor of these being moved to a new ticket. I'll go ahead and create one now. I appreciate you offering.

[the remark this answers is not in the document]
eecavanna: I created a ticket here: https://github.com/microbiomedata/nmdc-server/issues/1679

I'll resolve this conversation.

eecavanna:
Here's what the API response object looks like with this new property in place. This is in my local development environment, with the latest production Postgres dump (i.e. the one from June 27) loaded into my local Postgres database.

<details>
<summary>Show/hide curl command</summary>

```shell
curl http://localhost:8080/api/stats
```

</details>

```json
{
  "studies": 44,
  "non_parent_studies": 32,
  "locations": 50,
  "habitats": 80,
  "data_size": 63371176676244,
  "metagenomes": 4648,
  "metatranscriptomes": 411,
  "proteomics": 74,
  "metabolomics": 493,
  "lipodomics": 198,
  "organic_matter_characterization": 2583,
  "wfe_output_data_size_bytes": 51703764742677
}
```

> One nitpick I have is that I think I'd prefer moving `data_size` (assuming we're keeping it) to the end of the list, right after `wfe_output_data_size_bytes`.

Emphasis on:

```js
  "data_size": 63371176676244,                  // 63.37 TB (57.63 TiB)
  "wfe_output_data_size_bytes": 51703764742677  // 51.70 TB (47.02 TiB)
```

The new property is currently named `wfe_output_data_size_bytes`, which is short for "Workflow Execution output data objects' file size (in bytes)." The `data_size` part of the name was copied from the name of the existing property, `data_size` (whose value is "[All] data objects' file size (in bytes)").

eecavanna:
Thanks, @hesspnnl. That did the trick; i.e., now, the two properties related to file size are next to one another.

```json
{
  "...": "...",
  "wfe_output_data_size_bytes": 51703764742677,
  "data_size": 63371176676244
}
```

Hi @aclum, what are your thoughts on how this information gets presented to the HTTP client? 
Options include:
1. Keep both the `data_size` field and the `wfe_output_data_size_bytes` field how they are (shown above)
2. Get rid of the `data_size` field and leave the `wfe_output_data_size_bytes` field how it is
3. Get rid of the `wfe_output_data_size_bytes` field and put its value in the `data_size` field

If we resolve this before noon-ish, we can include this in today's release. I wouldn't typically be in favor of introducing something so late in the pre-release test phase. In this case, I understand the scope and have tested the feature locally several times.

eecavanna (approved):
We can merge this in now, based on a conversation in today's infrastructure meeting.

Issue #1676, opened 2025-06-23
Docker PostgreSQL database setup results in exposed database


codytodonnell:
After setting and spinning up the initial docker stack, I received this message from LBNL security:

> Our automated network scanning has discovered a database exposed on the following computer.  Based on our records, we believe you are responsible for this computer.  

I believe this is a result of [this line](http://github.com/microbiomedata/nmdc-server/blob/main/docker-compose.yml#L26) in the `docker-compose.yml` file.

After talking with @shreddd about it, he suggested adding 127.0.0.1: to the port so that the database is explicitly and exclusively hosted on localhost:

```
    ports:
      - "127.0.0.1:5432:5432"
  ```

eecavanna:
Hi @codytodonnell, FYI:

Since you worked on this this sprint (and to give you exposure to another aspect of ticket management on the project), I'm going to add this ticket to the current sprint board and set its status.

To do that, on this Issue page, I'll click "Projects" in the right sidebar, and then–in the menu that appears—I'll click on the project whose name says the current 2-week period we are in (in this case, it's "Sprint 65"). Each sprint has its own GitHub Project, whose contents can be viewed as a kanban board (a.k.a. "the sprint board").

Then, since you have already implemented this and are ready for people to review a PR for it, I'll set this ticket's _status_ in the sprint, to "In Review." To do that, I'll click on the status menu on the project item (which, at this point, will be visible in the right sidebar) and choose "In Review".

PR #1677, opened 2025-06-23
Explicitly host the postgreSQL database on localhost

codytodonnell:
Fixes https://github.com/microbiomedata/nmdc-server/issues/1676

I believe this should fix the issue causing the PostgreSQL database to be exposed on the LBNL network when spinning up the docker stack onsite at LBNL. Can confirm the stack still works as expected with this change.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -23,7 +23,7 @@ services:
       # Our MGA ingest makes some *very* large transactions
       - "max_wal_size=2GB"
     ports:
-      - "5432:5432"
+      - "127.0.0.1:5432:5432"
 
   worker:
     image: ghcr.io/microbiomedata/nmdc-server/worker:main
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -23,6 +23,10 @@ services:
       # Our MGA ingest makes some *very* large transactions
       - "max_wal_size=2GB"
     ports:
+      # We specify `127.0.0.1` here so that other computers on the same LAN cannot
+      # access the container, even when the host's firewall is configured to allow
+      # incoming connections to Docker.
+      # Reference: https://github.com/compose-spec/compose-spec/blob/main/spec.md#short-syntax-3
       - "127.0.0.1:5432:5432"
 
   worker:


eecavanna:
Thanks for implementing this, @codytodonnell. Also, thanks for identifying this fix, @shreddd. 

I see, in the Docker Compose [spec](https://github.com/compose-spec/compose-spec/blob/37cc49e897219b2843f90f296c715339e3c1fae8/spec.md?plain=1#L1781), some information that seems consistent with this fix:

```md
- `HOST` is `[IP:](port | range)` (optional). If it is not set, it binds to all network interfaces (`0.0.0.0`). 
```

I will tag one of our teammates that develops on a Linux machine, so he can give this a try locally and confirm everything still works for him. If it does, I'd approve of this being merged in.

eecavanna:
Hi @naglepuff and/or @marySalvi, if you are using a Linux-based development machine these days, will you confirm that the addition of the `127.0.0.1` to the port mapping doesn't interfere with your setup?

eecavanna:
I did a little experiment here on my Mac (on my home LAN).

For reference, for the duration of this experiment, my macOS firewall was set up to allow incoming connections to Docker.

image

First, I tried running a web server in a Docker container, mapping ports via `-p 8989:8000`.

<details>
<summary>Show/hide details</summary>

I ran the following Docker command to spin up a web server on port 8989:

```sh
docker run --rm -p 8989:8000 -w /usr/src/app python:3.12 python -m http.server 8000
```

Then, I opened a web browser on my Mac, visited the URL `http://localhost:8989`, and saw a web page being served by that Docker container.

Next, I ran `$ ifconfig | grep inet` to get my Mac's IP address (on my home LAN), which was `192.168.1.2`.

Finally, on my iPhone (which is also on my home LAN), I opened a web browser and visited the URL `http://192.168.1.2:8989`, and saw that same web page. Indeed, the Docker container's logs would show a new line each time I refreshed the web page on my iPhone.

</details>

**Summary:** With ports having been mapped via `-p 8989:8000`, I can access the Docker container from another computer on my home LAN.

Then, I tried running a web server in a Docker container, mapping ports via `-p 127.0.0.1:8989:8000`.

<details>
<summary>Show/hide details</summary>

Next, I stopped that Docker container and ran the same command—except with `-p 127.0.0.1:8989:8000` (i.e. I specified an IP address).

```sh
docker run --rm -p 8989:8000 -w /usr/src/app python:3.12 python -m http.server 8000
```

Once again, I opened a web browser on my Mac, visited the URL `http://localhost:8989`, and saw a web page being served by that Docker container.

However, this time, on my iPhone (which is still on my home LAN), I tried refreshing the web page and got a "This site can't be reached" error message.

</details>

**Summary:** With ports having been mapped via `-p 127.0.0.1:8989:8000`, I cannot access the Docker container from another computer on my home LAN.

**Personal comments:** I found this helpful for internalizing things because, until last week, I didn't realize Docker containers running on my Mac were accessible to other computers on the same LAN. Based on past experience with Docker on Windows, I had assumed that wasn't the case (in hindsight, I'm thinking the only reason it wasn't is that I had configured the Windows firewall to block that access, as I routinely deny applications' access to open ports in the Windows firewall).

docker-compose.yml
```diff
@@ -23,7 +23,7 @@ services:
       # Our MGA ingest makes some *very* large transactions
       - "max_wal_size=2GB"
     ports:
-      - "5432:5432"
+      - "127.0.0.1:5432:5432"
```
eecavanna, line 26: I recommend adding a comment about this because I think it is the only occurrence of it in any of the NMDC repositories.

```suggestion
      # We specify `127.0.0.1` here so that other computers on the same LAN cannot
      # access the container, even when the host's firewall is configured to allow 
      # incoming connections to Docker.
      # Reference: https://github.com/compose-spec/compose-spec/blob/main/spec.md#short-syntax-3
      - "127.0.0.1:5432:5432"
```

codytodonnell: Good suggestion, I just added in that comment

codytodonnell:
Thanks all, that's correct, this just needs to be merged into `main`. Will reserve `hotfix-` for instances as you described in the future.

I don't have the permission to merge into `main` yet. Can someone grant me permission?

PR #1686, opened 2025-07-08
Fix bar chart and search results containers border cutoff

### LINKED ISSUE #1081: UI: Bar chart container border is hidden at corners
I suspect a child element does not have a border radius set, and so the (sharp) corners of its border are "covering" the (rounded) corners of the parent element.

### Screenshot

image

codytodonnell:
Fixes #1081 

This addresses an issue causing the border corners to be clipped by a pixel in the bar chart and search results. The bar chart issue is fixed by adding a wrapper element to the chart and hiding its overflow. The search result issue is fixed by ensuring the `v-list` bottom borders also have a border radius of 4px to match the wrapping panel.

### DIFF
--- a/web/src/components/Presentation/FacetBarChart.vue
+++ b/web/src/components/Presentation/FacetBarChart.vue
@@ -155,20 +155,12 @@ export default defineComponent({
 </script>
 
 <template>
-  <div class="bar-chart">
-    <GChart
-      ref="chartRef"
-      type="BarChart"
-      :data="chartData"
-      :options="barChartOptions"
-      :events="chartEvents"
-    />
-  </div>
+  <GChart
+    ref="chartRef"
+    type="BarChart"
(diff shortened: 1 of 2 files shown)


eecavanna:
Thanks, @marySalvi!

---

I'm still spinning this up locally. Installing NPM dependencies is taking a looong time on this hotel WiFi.

image

> [Source](https://medium.com/frontendweb/find-how-many-packages-we-need-to-run-a-react-hello-world-app-695fbb755af7) 😅

codytodonnell:
@marySalvi Good suggestion! I just modified the code to use the vuetify classes instead. Looks like this still solves the issue.

eecavanna:
Thanks, @codytodonnell! At this point, as far as I'm concerned, you can merge this in whenever you want (does GitHub allow you to merge PRs?).

PR #1696, opened 2025-07-16
Refactor API summary and stats test to be two independent tests

codytodonnell:
The `test_api_summary` function was testing two different endpoints together, `/api/summary` and `/api/stats`. This PR separates that test into two. The summary test is also improved now to test both the success of the response and the shape of the data by comparing it to the expected `DatabaseSummary` fields.


eecavanna (approved):
Looks great! Thanks for splitting up the test so each test is only focused on a single API endpoint. I'm OK with this being merged in as is.

Issue #1608, opened 2024-08-15
Add ability to upload an image to NMDC submissions via submission portal


mslarae13:
/study page of the submission portal should have a "upload image" spot to add a photo of the PI.

![Screenshot 2024-08-15 at 12 44 44 PM](https://github.com/user-attachments/assets/90f99176-4f80-429e-9d7a-35d47b1b6862)

- shrink the PI ORCID box
- add a upload image box
  - How do we store these? Where should it go?
- info grey text should read "An image of the PI for a [research study](https://data.microbiomedata.org/details/study/nmdc:sty-11-5tgfr349) or a logo for a [consortia](https://data.microbiomedata.org/details/study/nmdc:sty-11-nxrz9m96) to share on the study landing page.

pkalita-lbl:
A few miscellaneous thoughts:
 
1. I think we have a path forward on the technical details of storing and retrieving media. See #1612 for details. 
2. The annotation in the screenshot says “A image of the PI for a research study or logo for a consortia”. Those two types of images go to different places in the schema (`Study.principal_investigator.profile_image_url` and  `Study.study_image.url` respectively). So in order to direct the image to the correct place, we’d need to add something to the Submission Portal UI that asks the user whether the study is a research study or a consortia. Not a big deal, but we’d need to decide where that would go and what the language around it would be.
3. Part of the Data Portal ingest process currently tries to avoid storing duplicate image data. Say we already have a study with a PI named “Morgan Jones”. If a second study comes in with a PI named “Morgan Jones”, the ingest will always reuse the existing PI image (provided by the first study) no matter what else the second study contains.
    1. This behavior is slightly questionable because matching by name is a weird choice when we have the PI email. Plus, the PI name isn’t even technically required in the Submission Portal.
    2. If we keep this behavior while allowing user image uploads, this could be confusing to users. If they have one submission accepted and ingested, then come back later and make a second submission with a different PI image, they might rightfully wonder why that second image they uploaded isn’t displayed on the data portal after ingest.
    3. There is something to be said, though, for keeping PI images consistent across studies. It would be nice if one PI creates a number of submissions that they don’t have to keep uploading the same photo over and over. That gets into the idea of attaching an image to the user's portal profile (i.e. the `user_logins` table in Postgres). That might simplify some use cases, but it also might complicate the cases where the user creating the submission is _not_ the PI.

mslarae13:
[...]

PR #1706, opened 2025-07-22
Add API endpoints for handling submission-related image uploads

### LINKED ISSUE #1612: Set up infrastructure for storing submission media in object store
This task is to set up the media upload, storage, and retrieval infrastructure to support users providing [PI profile images](https://github.com/microbiomedata/nmdc-server/issues/1608) via the Submission Portal and [study/sample images](https://github.com/microbiomedata/nmdc-field-notes/issues/95) via Field Notes.

On the 17 April 2025 infrastructure call there was general agreement to use an object store for this media, as opposed to using direct NERSC file system access or blobs in Postgres. See [here](https://app.diagrams.net/#G1tLuYXZhSno22zuDBAM15gSBM8YMQKZ5r#%7B%22pageId%22%3A%22dvvz-SrvIYoVY2cKc7al%22%7D) for more details on the proposal.

This task is to get the backend/database pieces in place so that the above linked issues can proceed. This will include working on the finer points of:

* Choose an object store provider. On the infrastructure call it seemed like there was general support for Google Cloud Storage. But I don't know if we fully ruled out spinning up our own MinIO service. 
* Ensure that relevant team members have access to the object store admin interface and understand how to create and manage buckets
* Take into account dev and production systems. For example, we should allow internal testers to upload/replace media on the dev Submission Portal without them having to worry about affecting production data.
* Take into account local development. If using a cloud provider, should local development stacks spin up an [emulator](https://github.com/fsouza/fake-gcs-server) instead of talking to the real object store?
* Ensure a baseline level of privacy. For example, a user should not be able to list or retrieve media for submissions that they do not have permission to view. A user should not be able to upload media for a submission that they do not have permission to edit.

pkalita-lbl:
Fixes #1612 

### Summary

These changes set up API endpoints for managing various types of submission-related images. The images themselves are ultimately stored in a Google Cloud Storage (GCS) bucket and we only store information about object names in Postgres. These changes **do not** introduce any usage of these endpoints. That will be done as part of other work:

* https://github.com/microbiomedata/nmdc-server/issues/1608
* https://github.com/microbiomedata/nmdc-field-notes/issues/95

The new endpoints are:

* `POST /api/metadata_submission/{id}/image/signed_upload_url`
    
    This endpoint is used request a time-limited [signed upload URL](https://cloud.google.com/storage/docs/access-control/signed-urls). Once the client receives this URL they can use it to upload a file directly to GCS.
* `POST /api/metadata_submission/{id}/image/{image_type}`

    This endpoint is expected to be called by the client after successfully uploading a file to GCS via a signed URL. There are three ways that an image can be associated with a submission: a single-valued PI headshot image, a single-valued primary study image, and a multivalued collection of other study images. The `image_type` parameter determines how the new storage object's name is associated with the submission. 
* `DELETE /api/metadata_submission/{id}/image/{image_type}`

    This endpoint is used to remove an image object association from a submission. The `image_type` parameter functions similarly to the above `POST` endpoint. When removing an image from the multivalued collection of study images, the particular image to remove must be indicated by also providing the `image_name` query parameter.

#### Database changes

References to all images in GCS associated with submissions are stored in a new table named `submission_images_object`. The slightly awkward name reflects the fact that it holds information about [objects](https://cloud.google.com/storage/docs/objects) stored in the `nmdc-submission-images` [bucket](https://cloud.google.com/storage/docs/buckets).

Two new foreign key columns (`pi_image_name`) and (`primary_study_image_name`) on the `submission_metadata` table enable the two single-valued associations. A new association table (`submission_study_image_association`) enables the multi-valued "other study images" association.

image

Why am I not just stuffing all of these image references into the venerable `metadata_submission` JSONB column? 

1. These fields will be updated by clients using dedicated API endpoints (by necessity because of the interaction with the GCS server). So having the information mixed into the JSONB object didn't seem advantageous in any way.
2. More broadly speaking, I think the big JSONB object wasn't the best design decision from the get-go. There's nothing about the submission data that requires that much flexibility. So the first step towards moving away from that model is not adding to it.

#### Fake GCS server

In order to not force local development systems to talk to the real GCS server and incur the (admittedly small) cost of storing and transferring data, a GCS emulator is also included. The Docker image is called [`fake-gcs-server`](https://github.com/fsouza/fake-gcs-server), and the code generally uses the terminology "fake" when referring to it as well.

The application will use the fake GCS server by default, but you can switch to the real one with an environment variable (`NMDC_GCS_USE_FAKE`). Even with the fake GCS server, you must provide Google Cloud credentials in the same way that you'd provide them to the real GCS server. I can explain why, but it gets quite in-the-weeds. Regardless, there are instructions in `development.md` for how to get that set up.

#### The `Storage` class

The new `Storage` class (`nmdc_server/storage.py`) provides the main interface to GCS operations. This class is responsible for setting up a `Client` and using it to perform operations on buckets and objects (referred to as "blobs" by the `google-cloud-storage` Python library methods; I tried to stick with the "objects" nomenclature but they're synonymous). 

The `nmdc_server/storage.py` module also exports an instance (`storage`) of the `Storage` class which is intended to be used as a singleton elsewhere in the code.

#### CLI commands

Two new `nmdc-server` CLI commands have been added:

* `nmdc-server storage init` -- this command is responsible for ensuring that all the buckets (currently only one) that the code knows about actually exist in GCS. If a bucket does not exist and you're using the fake GCS server, the bucket will automatically be created. If a bucket does not exist and you're using the real GCS server, an exception will be raised. 

    Why are we not automatically creating buckets on the real GCS server? Creating a bucket and configuring its CORS settings automatically on the real GCS server would involve giving more permissions to the [service account](https://cloud.google.com/iam/docs/service-account-overview) used to access GCS. Since creating new buckets will be relatively rare, it seemed like a better tradeoff to keep the permissions limited and create buckets manually when needed.

    This command is called during application startup (`start.sh`) and prior to testing with `tox`.

* `nmdc-server storage vacuum` -- this command will delete any object from GCS that is not referenced in the `submission_images_object` table. This command is not currently automatically called anywhere. Under normal operation the new endpoints will remove de-referenced GCS objects automatically. But I did find it useful during debugging and development. At some point we could consider running this command periodically in production just in case.

#### Pydantic models

Because all of the content of the `nmdc-submission-images` bucket is private, a user needs a time-limited signed URL in order to download the image. This generation is managed in the `SubmissionMetadataSchema` Pydantic model so that it happens as one of the last steps of the object going "out the door" (i.e. being serialized to JSON). At that point, upstream code (e.g. in API request handlers) should have already determined that the requesting user is allowed to see images associated with the given submission.

I also took this opportunity to create a new Pydantic model (`SubmissionMetadataSchemaListItem`) which represents a "slim" view of a submission. It contains only the fields needed to render a submission in the `SubmissionList.vue` component. In particular none of the information from the `metadata_submission` field is needed. This makes the response to `GET /api/metadata_submission` drastically smaller. And, coincidentally, signed URL generation can be skipped because those fields are also not included in `SubmissionMetadataSchemaListItem`.

Finally, I fixed a handful of places where we were using the old-style way of converting a SQLAlchemy model to a Pydantic model:

```python
some_pydantic_instance = SomePydanticModel(**some_sqla_instance.__dict__)
```

By replacing it with the [newer style](https://docs.pydantic.dev/latest/concepts/models/#arbitrary-class-instances):

```python
some_pydantic_instance = SomePydanticModel.model_validate(some_sqla_instance)
```


tests/test_submission.py
```diff
@@ -517,20 +514,21 @@ def test_try_edit_expired_locked_submission(db: Session, client: TestClient, log
...
+        lock_updated=datetime.now(tz=UTC) - timedelta(hours=1),
     )
     fakes.SubmissionRoleFactory(
         submission=submission,
         submission_id=submission.id,
         user_orcid=logged_in_user.orcid,
         role=SubmissionEditorRole.owner,
     )
-    payload = SubmissionMetadataSchema(**submission.__dict__).json(exclude_unset=True)
+    payload = SubmissionMetadataSchema.model_validate(submission)
```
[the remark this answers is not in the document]
pkalita-lbl: > may trigger additional database queries for related objects (like images with signed URL generation).

There is no concern about signed URL generation here because actual signed URL generation is stubbed out in the testing environment.

nmdc_server/schemas_submission.py
```diff
@@ -170,6 +195,39 @@ def populate_roles(cls, metadata_submission, info: ValidationInfo):
...
+    @computed_field  # type: ignore
+    @property
+    def study_images(self) -> List[str]:
+        """Returns a list of signed URLs for all study images."""
+        if not self.study_images_objects:
+            return []
+        return [
+            storage.get_signed_download_url(BucketName.SUBMISSION_IMAGES, img.name).url
+            for img in self.study_images_objects
+        ]
```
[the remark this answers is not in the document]
pkalita-lbl: > For list views or bulk operations, this could result in many API calls

This is already handled elsewhere by the introduction of the `SubmissionMetadataSchemaListItem` Pydantic model, which does not include these fields.

> Consider caching or implementing lazy loading for these computed fields

The caching suggestion provided here will not work. Because the signed URLs have an expiration date any caching needs to be time-based, otherwise we run the risk of providing a cached -- but expired -- URL. Python's standard `@cached_property` doesn't offer that functionality. There are [third-party options](https://cachetools.readthedocs.io/en/latest/#cachetools.TTLCache), but for now I don't think the volume of image usage is going to be high enough to worry about it.

nmdc_server/utils.py
```diff
@@ -34,3 +36,48 @@ def log_extras(req: Request) -> Dict[str, Any]:
...
+        # Truncate (from the start) to fit within 512 bytes, ensuring we don't cut a multibyte
+        # character
+        start_byte = total_bytes - max_total_bytes
+        # (b & 0xC0) == 0x80 checks if the byte b is a continuation byte in UTF-8
+        # https://en.wikipedia.org/wiki/UTF-8#Description
+        while start_byte <= total_bytes and (utf8_bytes[start_byte] & 0xC0) == 0x80:
+            start_byte += 1
+        combined = utf8_bytes[start_byte:].decode("utf-8", errors="ignore")
+
+    return combined
```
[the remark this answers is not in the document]
pkalita-lbl: I don't think that adds much.

nmdc_server/api.py
```diff
@@ -1464,8 +1455,140 @@ async def suggest_metadata(
...
+    if body.file_size > settings.max_submission_image_file_size:
+        raise HTTPException(
+            status_code=status.HTTP_422_UNPROCESSABLE_ENTITY, detail="File size exceeds limit"
+        )
+
+    # Ensure the requested submission exists and the user has permission to edit it
+    submission = get_submission_for_user(db, id, user.orcid, allowed_roles=context_edit_roles)
+
+    # Ensure that the incoming image will not push the submission over its quota
+    potential_max_size = submission.study_images_total_size + body.file_size
```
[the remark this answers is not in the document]
pkalita-lbl: Possible future optimization, not needed for now.

.env.example
```diff
@@ -38,6 +38,14 @@ NMDC_API_JWT_SECRET=generateme
 # Change this value to "testing" to run tests outside tox
 NMDC_ENVIRONMENT=development
 
+# == Cloud Storage ==
+NMDC_GCS_USE_FAKE=True
+# Get this from the Google Cloud Console, only needed if NMDC_GCS_USE_FAKE=False
+# NMDC_GCS_PROJECT_ID=changeme
+# Use `openssl rand -hex 4` to generate the random string once. Keep the `local_` prefix for local development.
+# NMDC_GCS_OBJECT_NAME_PREFIX=local_[random_string]
```
eecavanna, line 46: I'm thinking about suggesting that developers also incorporate their GitHub handle (if they're comfortable) because it could help them and GCS admin-ish team members trace down issues (e.g. "someone's flooding the bucket with 10K images a day by running a test without realizing they haven't switched back to the fake bucket").

Example: ` local_eecavanna_{randomness}...`

eecavanna: On second thought, we can survey team members as needed (e.g. "who is using the prefix `FvjjSd1234`?").

nmdc_server/api.py
```diff
@@ -1052,26 +1055,48 @@ async def get_metadata_submissions_report(
...
+
+
+@router.get(
+    "/metadata_submission/slim",
+    tags=["metadata_submission"],
+    responses=login_required_responses,
+    response_model=query.Paginated[schemas_submission.SubmissionMetadataSchemaSlim],
+)
+async def list_submissions_slim(submissions=Depends(get_paginated_submission_list)):
+    """Return a paginated list of submissions in slim format."""
```
eecavanna, line 1096: I like the word, "slim," for this (I have used "lite" in other projects). I think "slim" will be easier to use in spoken conversation.

nmdc_server/cli.py
```diff
@@ -352,5 +349,46 @@ def generate_static_files(remove_existing):
     click.echo("Done generating static files.")
 
 
+@cli.group(name="storage")
+def storage_group():
+    """Commands for managing storage buckets and objects."""
+    pass
```
eecavanna, line 355: Consider raising https://docs.python.org/3/library/exceptions.html#NotImplementedError

eecavanna (commented):
Reviewed the first few files from my iPhone while out for a walk. Submitting those comments before my battery runs out. I'll resume reviewing this PR on Monday.

Looks great so far! Great PR description! 🙏

nmdc_server/config.py
```diff
@@ -46,6 +46,14 @@ class Settings(BaseSettings):
     zip_streamer_chunk_size_bytes: int = 2 * 1024 * 1024
     zip_streamer_nersc_data_base_url: str = "https://data.microbiomedata.org/data"
 
+    # for cloud storage
+    gcs_use_fake: bool = True
+    gcs_project_id: str | None = None
+    gcs_object_name_prefix: str  # no default, must be set in .env file
+
+    max_submission_image_file_size: int = 25 * 1000 * 1000  # 25 MB
+    max_submission_image_total_size: int = 1 * 1000 * 1000 * 1000  # 1 GB
```
eecavanna, line 55: The "`# 25 MB`" and "`# 1 GB`" comments are the only things I have seen so far that have indicated to me the units of these values. I noticed the `zip_streamer_chunk_size_bytes` variable above has the units embedded in its name. As long as the comments remain intact, I'm OK with this (I assume variable name length factored into the decision to use the current name).

.env.example
```diff
@@ -38,6 +38,14 @@ NMDC_API_JWT_SECRET=generateme
 # Change this value to "testing" to run tests outside tox
 NMDC_ENVIRONMENT=development
 
+# == Cloud Storage ==
+NMDC_GCS_USE_FAKE=True
+# Get this from the Google Cloud Console, only needed if NMDC_GCS_USE_FAKE=False
```
eecavanna, line 43: I suggest adding a semicolon or period, as a way of emphasizing that the second half of the statement is referring to the variable, not to the Google Cloud Console.

```suggestion
# Get this from the Google Cloud Console; only needed if NMDC_GCS_USE_FAKE=False
```

docker-compose.yml
```diff
@@ -63,6 +64,7 @@ services:
     volumes:
       - ./data/ingest:/data/ingest
       - ./nmdc_server:/app/nmdc_server
+      - $HOME/.config/gcloud/application_default_credentials.json:/gcp/credentials.json
```
eecavanna, line 67:
```suggestion
      # Note: On Linux and macOS, `$HOME` contains the path to the user's home directory.
      - $HOME/.config/gcloud/application_default_credentials.json:/gcp/credentials.json
```

nmdc_server/fakes.py
```diff
@@ -400,6 +401,10 @@ class Meta:
     }
     locked_by = None
     lock_updated = None
+    is_test_submission = False
```
eecavanna, line 404: I noticed the other newly-added variables have a type hint (maybe only because those types aren't built-in Python types and aren't as implicit from their initial values).

```suggestion
    is_test_submission: bool = False
```

nmdc_server/migrations/versions/09a705b0bc00_submission_images.py
```diff
@@ -0,0 +1,103 @@
+"""Add submission_images_object table.
+
+The name of the table indicates that stores references to objects in the nmdc-submission-images
```
eecavanna, line 3: I think a word is missing.

```suggestion
The name of the table indicates that it stores references to objects in the nmdc-submission-images
```

nmdc_server/models.py
```diff
@@ -984,6 +1052,29 @@ def owners(self) -> list[str]:
             if role.role == SubmissionEditorRole.owner
         ]
 
+    @property
+    def study_images_total_size(self) -> int:
+        """Calculate the total size of all study images associated with this submission."""
```
eecavanna, line 1057:
```suggestion
        """Calculate the total size (in bytes) of all study images associated with this submission."""
```

nmdc_server/storage.py
```diff
@@ -0,0 +1,170 @@
...
+from google.cloud.storage import Blob, Bucket, Client
+
+from nmdc_server.config import settings
+from nmdc_server.schemas import SignedUrl
+
+
+class BucketName(StrEnum):
+    """Enum for GCS bucket names"""
+
+    SUBMISSION_IMAGES = "nmdc-submission-images"
```
naglepuff, line 17: Does each deployment of `nmdc-server` need more than 1 bucket? Should the bucket name be configurable via env var instead of hard-coded here?

nmdc_server/utils.py
```diff
@@ -34,3 +36,48 @@ def log_extras(req: Request) -> Dict[str, Any]:
         "method": req.method,
         "url": req.url,
     }
+
+
+def sanitize_filename(filename: str) -> str:
```
naglepuff, line 41: should this be moved into the `storage.py` module? at first glance it seems that this is only used to adhere to GSC requirements.

naglepuff: also too bad that the `google-cloud-storage` library doesn't seem to provide this functionality

.env.example
pkalita-lbl: I was going back and forth between using `local_$USER` and `local_[random]`. I ended up choosing the latter because of the [recommendation](https://cloud.google.com/storage/docs/objects#naming) to avoid personally identifiable information in object names -- even if it is just for development purposes.

nmdc_server/cli.py
pkalita-lbl: Right, I believe `pass` is the correct way to have a command group that doesn't need to do anything common to the group's subcommands. At least, that's what the docs show ([for example](https://click.palletsprojects.com/en/stable/commands/#command-chaining)).

nmdc_server/config.py
pkalita-lbl: > I assume variable name length factored into the decision to use the current name

You're giving me too much credit 😂 I just didn't think about it. But I like the idea of adding `_bytes` to the name. I will do that.

docker-compose.yml
pkalita-lbl: As far as I know (and I freely admit I could be wrong), it's not up to us. 

[Google Cloud authentication](https://cloud.google.com/iam/docs/overview) is mind-bending-ly complicated, but I'll try to step through my reasoning here:
1. Based on this [decision tree](https://cloud.google.com/docs/authentication#auth-decision-tree) I reckon that the `nmdc-server` code should authenticate using a service account. That's also good because a service account is the [recommended way](https://cloud.google.com/storage/docs/access-control/signed-urls#overview) of authenticating when generated signed URLs (which we need to do).
2. In environments deployed on Rancher, we can securely manage the service accounts JSON credential file via secrets.
3. But for local development environments, I'd rather not have to be ad-hoc passing around said JSON credential files to each developer who needs them. And we don't have to, either. Because there is [service account impersonation](https://cloud.google.com/docs/authentication/use-service-account-impersonation).
4. When accessing Google Cloud through a client library (as `nmdc-server` does), the [recommendation](https://cloud.google.com/docs/authentication/use-service-account-impersonation#adc) is to use service account impersonation through the Application Default Credentials (ADC) mechanism.
5. At long last, I don't think you get to choose where your ADC file lives. It just goes to `$HOME/.config/gcloud/application_default_credentials.json` -- again, unless I'm missing something.

nmdc_server/storage.py
pkalita-lbl: All deployments share one bucket for submission images.

Setting up a new bucket on the real GCS server requires elevated permissions to create it and set the CORS configuration. That is why the `nmdc-server` code _doesn't_ do it automatically. For now, setting up a bucket is a manual operation; it will be done very rarely. Later, it could be automated via helm charts. Either way, it's outside the scope of what `nmdc-server` should worry about.

Since all deployments share one bucket, objects coming into the bucket from different environments are differentiated by a prefix on the object name. That is configured via an environment variable (namely `NMDC_GCS_OBJECT_NAME_PREFIX`).

It's possible there will be other buckets for other purposes (other than submission images, that is) in the future. The enum is maybe a bit overkill for now, but it allows for future expansion.

nmdc_server/utils.py
pkalita-lbl: > should this be moved into the storage.py module? 

Yes, that's a good idea. Will do.

> too bad that the google-cloud-storage library doesn't seem to provide this functionality

Couldn't agree more.

nmdc_server/api.py
pkalita-lbl: I just realized that I think I subconsciously picked that word because the Gene Ontology (the thing I'm usually working on when I'm not working on NMDC) has subsets that are colloquially called "slims": https://geneontology.org/docs/go-subset-guide/

nmdc_server/storage.py
```diff
@@ -0,0 +1,170 @@
...
+        """Google Cloud Storage client."""
+        client_args: dict[str, Any] = {
+            "project": self.project_id,
+        }
+
+        if self._is_testing:
+            client_args["credentials"] = AnonymousCredentials()
+            client_args["client_options"] = {"api_endpoint": "http://localhost:4443"}
+        elif self.use_fake_server:
+            client_args["client_options"] = {"api_endpoint": "http://storage:4443"}
```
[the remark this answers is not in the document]
naglepuff: > It's determined by the Python client library itself. I believe part of the logic for how its determined involves reading what's know as the "universe domain" from your Application Default Credentials file.

Cool. Lots of magic going on with the credentials file.


> My thought was that it's better to let the library make the determination itself and only override it in certain special scenarios.

Makes sense to me. What if the api endpoint was an environment variable? Could we get rid of the `self.use_fake_server` check entirely?
```
if settings.environment == "testing":
    client_args["credentials"] = AnonymousCredentials()
if settings.gcs_api_endpoint:
    client_args["client_options"] = { "api_endpoint": settings.gcs_api_endpoint }
```

I'm trying to avoid the hard-coded strings here.

eecavanna:
I'm planning on reviewing this more this evening (hopefully, to completion). Sorry about the delay.

docs/development.md
```diff
@@ -72,6 +72,32 @@ NMDC_MONGO_USER=changeme
 NMDC_MONGO_PASSWORD=changeme
 ```
 
+### Google Cloud Storage
+
+Google Cloud Storage (GCS) is used to store images associated with Submission Portal submissions. By default, local development uses a local mock GCS server. This is controlled by the `NMDC_GCS_USE_FAKE` variable in `.env`. If you want to use the real GCS server, set this variable to `false`.
```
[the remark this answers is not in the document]
naglepuff: Yeah I think I wrote this comment here before looking at the extent of the usage of this environment variable, which is only in one file (I think). My main concern is keeping development and production environments as close as possible.

My experience with projects that use some kind of bucket storage has pretty much been with S3 as the storage service. These projects typically use MINIO as the development "fake" (s3-compatible) service, and the boto3 library, which knows how to talk to both services based entirely on configuration (i.e. there's no boolean toggle needed). So the use of an explicit flag gave me pause.

As long as the use of this setting is entirely confined to the `storage` module, its probably fine.

nmdc_server/storage.py
naglepuff: Thanks for clarifying. I think its probably a _good_ thing that the server code doesn't do any bucket creation itself.

My concern is more with keeping deployment configuration separate from application code. Even if all deployments share a bucket now, and its not likely to change in the future, I think its cleaner to make this (at least the string value) configurable through an environment variable. Totally fine to keep the enum, but maybe change to:

```suggestion
class BucketName(StrEnum):
    """Enum for GCS bucket names"""

    SUBMISSION_IMAGES = settings.gcs_submission_images_bucket
```

nmdc_server/schemas_submission.py
```diff
@@ -170,6 +196,39 @@ def populate_roles(cls, metadata_submission, info: ValidationInfo):
                     contributor["role"] = SubmissionEditorRole.viewer.value
         return metadata_submission
 
+    # Mypy doesn't understand the combined use of `@computed_field` and `@property`
+    # https://docs.pydantic.dev/latest/api/fields/#pydantic.fields.computed_field
+    @computed_field  # type: ignore
+    @property
+    def pi_image(self) -> Optional[str]:
```
eecavanna, line 203: Suggestion: rename the function to indicate that the value is a URL (which can be used to _get_ the image) as opposed to the image, itself.

nmdc_server/storage.py
```diff
@@ -0,0 +1,217 @@
...
+class BucketName(StrEnum):
+    """Enum for GCS bucket names"""
+
+    SUBMISSION_IMAGES = "nmdc-submission-images"
+
+
+class Storage:
+    """A class to manage Google Cloud Storage interactions."""
+
+    def __init__(self, project_id: str | None, use_fake_server: bool):
```
eecavanna, line 25: I suggest adding a docstring (caution: I drafted this one on my iPhone screen 🔬 )

```suggestion
    def __init__(self, project_id: str | None, use_fake_server: bool):
    """
    :param project_id: Google Cloud Project ID
    :param use_fake_server: Whether you want to use a fake GCS service or the real one
    """
```

nmdc_server/schemas.py
```diff
@@ -692,3 +692,34 @@ class VersionInfo(BaseModel):
...
+
+
+class SignedUploadUrlRequest(BaseModel):
+    """Request to generate a signed upload URL for a file.
+
+    This model is used to generate a signed URL for uploading files to the object store.
+    """
+
+    file_name: str
+    file_size: int
```
eecavanna, line 704: May be an opportunity to indicate units. (Property name may be dictated by a third party. Could indicate units via comment.)

tests/conftest.py
```diff
@@ -107,3 +110,23 @@ def logged_in_admin_user(db, client) -> schemas.User:
     client.headers["Authorization"] = f"Bearer {token_response.access_token.decode()}"
 
     return user
+
+
+@pytest.fixture
+def temp_storage_object() -> Generator[Callable[[BucketName, str], Blob], None, None]:
```
eecavanna, line 116: Nice! I like this "setup, yield, clean up" pattern for fixtures.

tests/test_submission.py
```diff
@@ -930,3 +928,455 @@ def test_metadata_suggest_invalid_type(client: TestClient, suggest_payload, logg
...
+    }
+
+    response = client.post(
+        f"/api/metadata_submission/{submission.id}/image/pi_image", json=upload_data
+    )
+
+    assert response.status_code == 403
+
+
+def test_set_submission_image_viewer_role_unauthorized(
```
eecavanna, line 1123: Thanks for testing this stuff so thoroughly (and automated-ly).

eecavanna (approved):
LGTM! Thanks for waiting for me to finish reviewing it, setting the stage so thoroughly in the PR description, demonstrating so many scenarios in the automated tests, and... for working through all the GCS setup/ramp up.

docs/development.md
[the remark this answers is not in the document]
naglepuff: I think you've done a good job here of keeping storage logic confined to its own module, so I don't imagine switching backends or adding support for additional backends would be too cumbersome. If we want to try and keep everything simple by going all-in on Google Cloud at this point, that sounds reasonable to me.

nmdc_server/schemas_submission.py
```diff
@@ -125,30 +126,55 @@ class SubmissionMetadataSchemaCreate(BaseModel):
...
+    created: datetime
+    is_test_submission: bool = False
+    sample_count: int = 0
+
+
+class SubmissionImagesObject(BaseModel):
+    model_config = ConfigDict(from_attributes=True)
+
+    name: str
+    size: int
```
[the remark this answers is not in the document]
pkalita-lbl: This is a Pydantic model that needs to be directly interoperable (`from_attributes=True`) with the equivalent [SQLAlchemy model](https://github.com/microbiomedata/nmdc-server/blob/9ddb351d9ffe4c972b2720507f725f61dac3d642/nmdc_server/models.py#L918-L932). That means they need to have the same attribute names. So changing the name here would necessitate changing a lot of other names to match and I'm a little hesitant to do that. Ultimately these values are going to come from File objects on the frontend which have [`size` properties](https://developer.mozilla.org/en-US/docs/Web/API/Blob/size) and I like that those names line up. But I'm happy to sprinkle a few comments around.

nmdc_server/schemas_submission.py
pkalita-lbl: I went back and forth on that myself. Because these properties get serialized to JSON, I think I was erring on the side of keeping the JSON objects succinct. I just did a random sampling of the documentation of some public APIs. Some do use a `url` suffix on similarly-purposed fields; some don't. I guess knowing that I don't have a problem with making it more verbose but explicit.

pkalita-lbl:
Haha no I appreciate the feedback! 

I'm going to merge `main` into this branch, re-linearlize the Alembic migrations, update the dev Rancher configuration, and then merge.

PR #1709, opened 2025-07-24
Fix column access during gene search

naglepuff:
The association tables used for joining during gene function search are not SQLAlchemy models, so the columns are not accessible as properties.


pkalita-lbl (approved):
Changes look fine. Looks like `black` would reformat something.

Issue #1390, opened 2024-09-13
Data Portal - map feature after clicking on Environment resets to Algeria


aclum:
If go to the data portal and click on "Environment" to the the sankey plot to view the distribution of the GOLD pathways, then click back on 'Omics' to see the bar chart the map feature shows part of North Africa instead of resetting to the default zoomed out view which shows the whole glob.

This is happening on prod in chrome and appears to be reproducible.
![Screenshot 2024-09-13 at 4 15 35 PM](https://github.com/user-attachments/assets/d7f130d9-aa3d-4c28-9280-c164fc3db731)

aclum:
This is behaving a bit differently now, it zooms on parts of Europe instead of showing the full map.

PR #1714, opened 2025-07-30
Fix map reset issue

yxu-lanl:
Closes https://github.com/microbiomedata/nmdc-server/issues/1390

1. Pass vistab to ClusterMap
2. Don't update map data if vistab ===1 (ENVIRONMENT tab is clicked)


marySalvi (approved):
Looks good.

Issue #1673, opened 2025-06-20
Add search/sort to the users table


naglepuff:
We have almost 1000 users in the data portal, which makes granting/revoking admin status a pain (the max page size is 50). I believe we can do this [client side](https://v2.vuetifyjs.com/en/components/data-tables/#filterable).

mslarae13:
@JamesTessmer something you could work on if you have some extra time :)

marySalvi:
@JamesTessmer @naglepuff 
Apologies for unsolicited advice, but this caught my eye. 
We populate the user table server side so any search/sort being done client side would be done on that subset of the users. So this should extend the current server call instead. 

https://github.com/microbiomedata/nmdc-server/blob/16ab724068eb0d72dacae0f37be7ae9002e10d55/web/src/views/User/UserPage.vue#L32-L33

JamesTessmer:
Merged PR, should be done now

PR #1718, opened 2025-08-04
Add search functionality for users table

JamesTessmer:
This PR resolves https://github.com/microbiomedata/nmdc-server/issues/1673

New code here adds a search box to the Users page to allow filtering based on name or ORCID.
If something matching the ORCID RE pattern is entered into the box then it will search for that, otherwise it defaults to user name.

@naglepuff I haven't done any real formatting on the search box to make it look nice, I was hoping for an initial bit of feedback on functionality before that. Mostly I was wondering if we wanted something like a toggle button to filter based on admin status.

Examples:
Screenshot 2025-08-04 at 12 28 53 PM
Screenshot 2025-08-04 at 12 28 37 PM
Screenshot 2025-08-04 at 12 28 28 PM

### DIFF
--- a/nmdc_server/api.py
+++ b/nmdc_server/api.py
@@ -1475,7 +1475,7 @@ async def get_users(
     search_filter: Optional[str] = None,
 ):
     users = db.query(User)
-    if type(search_filter) is str:
+    if search_filter:
         users = users.filter(
             (
                 models.User.name.ilike(f"%{search_filter}%")
--- a/nmdc_server/api.py
+++ b/nmdc_server/api.py
@@ -1,6 +1,7 @@
 import csv
 import json
 import logging
+import re
 import time
 from importlib import resources
 from io import BytesIO, StringIO
@@ -1472,8 +1473,17 @@ async def get_users(
     db: Session = Depends(get_db),
     user: models.User = Depends(admin_required),
     pagination: Pagination = Depends(),
+    search_filter: Optional[str] = None,
 ):
     users = db.query(User)
+    if type(search_filter) is str:
+        if re.match(
+            r"^\d{4}-\d{4}-\d{4}-\d{3}[\dX]$", search_filter
+        ):  # if search_filter is patterned like an ORCID then filter on that
+            users = users.filter(models.User.orcid == search_filter)
+        else:  # otherwise filter by name
+            users = users.filter(models.User.name.ilike(f"%{search_filter}%"))
+
     return pagination.response(users)
 
 
--- a/nmdc_server/api.py
+++ b/nmdc_server/api.py
@@ -1,7 +1,6 @@
 import csv
 import json
 import logging
-import re
 import time
 from importlib import resources
 from io import BytesIO, StringIO
@@ -1477,12 +1476,12 @@ async def get_users(
 ):
     users = db.query(User)
     if type(search_filter) is str:
-        if re.match(
-            r"^\d{4}-\d{4}-\d{4}-\d{3}[\dX]$", search_filter
-        ):  # if search_filter is patterned like an ORCID then filter on that
-            users = users.filter(models.User.orcid == search_filter)
-        else:  # otherwise filter by name
-            users = users.filter(models.User.name.ilike(f"%{search_filter}%"))
+        users = users.filter(
+            (
+                models.User.name.ilike(f"%{search_filter}%")
+                | models.User.orcid.ilike(f"%{search_filter}%")
+            )
+        )
 
     return pagination.response(users)
 
--- a/web/src/data/api.ts
+++ b/web/src/data/api.ts
@@ -685,11 +685,12 @@ async function me(): Promise<User | null> {
   }
 }
 
-async function getAllUsers(params: SearchParams) {
+async function getAllUsers(params: SearchParams, input_search_filter: String | null) {
   const { data } = await client.get<SearchResponse<User>>('users', {
     params: {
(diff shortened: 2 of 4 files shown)


web/src/data/api.ts
```diff
@@ -685,11 +685,12 @@ async function me(): Promise<User | null> {
   }
 }
 
-async function getAllUsers(params: SearchParams) {
+async function getAllUsers(params: SearchParams, input_search_filter: String | null) {
```
naglepuff, line 688:
```suggestion
async function getAllUsers(params: SearchParams, inputSearchFilter: string | null) {
```

If you _wanted_ to, you could probably add something to the `SearchParams` interface to capture a filter string. It looks like that interface is currently just a bunch of optional stuff anyway, and can be used with various types of "search" endpoints. Not necessary, but might be useful.

If you want to keep `input_search_filter`, I think it would be best to switch to camelCase (i.e. `inputSearchFilter`) for the parameter, since we're in typescript land. I also think the type should be `string` instead of `String`. The [typescript docs](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#number-string-boolean-symbol-and-object) recommend against using the object wrapper classes for primitive types.

web/src/views/User/UserPage.vue
```diff
@@ -18,6 +18,7 @@ export default defineComponent({
       page: 1,
       itemsPerPage,
     });
+    const search_filter = ref(null);
```
naglepuff, line 21: Would it be more appropriate to type this as a `Ref<string>` i.e.
```suggestion
    const searchFilter = ref('');
```
(also note the camel case suggestion)

web/src/views/User/UserPage.vue
```diff
@@ -29,8 +30,13 @@ export default defineComponent({
...
     ];
 
-    const users = usePaginatedResults(ref([]), api.getAllUsers, ref([]), itemsPerPage);
+    async function getUsers(params: SearchParams): Promise<SearchResponse<User>> {
+      return api.getAllUsers(params, search_filter.value);
+    }
+
+    const users = usePaginatedResults(ref([]), getUsers, ref([]), itemsPerPage);
     watch(options, () => users.setPage(options.value.page), { deep: true });
+    watch(search_filter, () => users.setPage(options.value.page), { deep: true });
```
naglepuff, line 39: I'm a little bit confused about what this watcher is doing. Please let me know your thoughts on this.

As an aside, in Vue 3 we'd be able to use [one watcher for both of these refs](https://vuejs.org/guide/essentials/watchers.html#watch-source-types), but for now we're still on Vue 2 so this is the cleanest way to do it.

nmdc_server/api.py
```diff
@@ -1472,8 +1473,17 @@ async def get_users(
...
+    search_filter: Optional[str] = None,
 ):
     users = db.query(User)
+    if type(search_filter) is str:
+        if re.match(
+            r"^\d{4}-\d{4}-\d{4}-\d{3}[\dX]$", search_filter
+        ):  # if search_filter is patterned like an ORCID then filter on that
+            users = users.filter(models.User.orcid == search_filter)
+        else:  # otherwise filter by name
+            users = users.filter(models.User.name.ilike(f"%{search_filter}%"))
```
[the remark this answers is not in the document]
JamesTessmer: My thought was that since it would only search ORCID if it matched the RE exactly then it would look for an exact match to the ORCID and then look for partial matching names otherwise.
I hadn't thought of putting both into the same filter case, but that's a good idea and I'll add that.

web/src/views/User/UserPage.vue
JamesTessmer: I didn't have that in at first, but without it the `users` var wouldn't see changes made to the textbox and update itself.

naglepuff: Should we boot the user back to page 1 when the filter changes? I'm wondering what the expecting behavior would be if someone was like on page 5 and then started typing in some filter.
```suggestion
    watch(search_filter, () => users.setPage(1), { deep: true });
```
(I actually don't know if the code indexes pages from 0 or 1, so this suggestion might be wrong)

JamesTessmer: Good catch, I was curious what would happen if you were on page 5. It displays all the correct results but keeps you on that page and still lets you change pages even if there's only 1 page worth of results which is kind of funky. 

I added the change (it does index from 0)

web/src/views/User/UserPage.vue
JamesTessmer: Good to know, I was familiar with the python string behavior but not typescript. I'm happy they evaluate the same way.

web/src/data/api.ts
JamesTessmer: I started working on this, and it was causing some strange issues on front end things like the number of rows per page being maxed out no matter what and the next page skipping around strangely. 

For now I think I'll leave this portion as is, but it might be worth making that change in the future if we have more things we want a filter string for.

I did change the casing for the variable name and changed `String` to `string`

JamesTessmer:
@naglepuff Pushed up your requested changes

naglepuff (approved):
LGTM! I had a couple additional suggestions, feel free to apply them if you'd like.

JamesTessmer:
@naglepuff I think both of your last comments were good and added those. Pushed the changes and will merge once tests pass.

PR #1719, opened 2025-08-04
Capture sampled_portion of processed samples

### LINKED ISSUE #1681: UI Changes for additional NOM metadata
For more information, see https://docs.google.com/presentation/d/1w6orzfuDeL3ysS0p-Ag-pmqWO-UNUdwn/edit?slide=id.g35489caefa8_2_97#slide=id.g35489caefa8_2_97

naglepuff:
Fix #1681 

## Changes

Adds `eluent_introduction_category`, `sampled_portion`, and `polarity_mode` to the data object table for Natural Organic Matter workflow outputs.

### Ingest Changes
Two changes have been made to data generation ingest.
1. During the recursive fetching of biosample inputs, we now also track `sampled_portion` from all processed samples in the chain and store this in the `annotations` column (which is a JSONB column) for each data generation. After ingest you can see the results with `SELECT DISTINCT(annotations->>'sampled_portions') FROM omics_processing;`
2. We store the `polarity_mode` of the mass spectrometry configuration in a similar manner. 

### Frontend Changes
Logic similar to changes made for lipidomics has been added to the `DataObjectTable` component to render this additional metadata in the appropriate headers.
image

### DIFF
--- a/nmdc_server/ingest/omics_processing.py
+++ b/nmdc_server/ingest/omics_processing.py
@@ -66,7 +66,9 @@ def find_parent_process(output_id: str, mongodb: Database) -> Optional[dict[str,
     return None
 
 
-def get_biosample_input_ids(input_id: str, mongodb: Database, results: set) -> set[Any]:
+def get_biosample_input_ids(
+    input_id: str, mongodb: Database, results: set[str], sampled_portions: set[str]
+) -> set[Any]:
     """
     Given an input ID return all biosample objects that are included in the input resource.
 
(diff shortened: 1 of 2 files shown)


samobermiller:
Thanks @naglepuff ! There are two updates I'm hoping we can make? 

1 ) I know you asked about displaying all `sampled_portion` values since its an array, and I said yes. But I didn't realize you were collecting all the processed samples upstream of the one used as input to data generation. We would only want the values for `sampled_portion` on the final processed sample that was used as input. Does that make sense? 
2) Can we display the title for sampled_portion so it reads "methanol layer" instead of "methanol_layer"?

Otherwise this looks great!

naglepuff:
> 1 ) I know you asked about displaying all `sampled_portion` values since its an array, and I said yes. But I didn't realize you were collecting all the processed samples upstream of the one used as input to data generation. We would only want the values for `sampled_portion` on the final processed sample that was used as input. Does that make sense? 
> 2) Can we display the title for sampled_portion so it reads "methanol layer" instead of "methanol_layer"?

Thanks for the feedback @samobermiller! I'll start working on the changes

naglepuff:
@samobermiller I've addressed your comments and updated the PR description and screenshot to reflect the changes.

samobermiller (approved):

PR #1720, opened 2025-08-05
Fix wrong sample list results from SQLAlchemy orm query

yxu-lanl:
Closes https://github.com/microbiomedata/issues/issues/1134

The problem is in Biosample query ORDER BY clause (order_by(desc(self.table.model.multiomics))). The 'multiomics' is not a unique column.

Solution:
Order by a unique column or a combination of columns that guarantees uniqueness.

Not sure ordering by name is better?
order_by(self.table.model.name, self.table.model.id)

### DIFF
--- a/nmdc_server/query.py
+++ b/nmdc_server/query.py
@@ -846,7 +846,7 @@ class BiosampleQuerySchema(BaseQuerySchema):
         biosample_query = (
             db.query(model)
             .join(subquery, model.id == subquery.c.id)  # type: ignore
-            .order_by(desc(self.table.model.multiomics))  # type: ignore
+            .order_by(self.table.model.name, self.table.model.id)  # type: ignore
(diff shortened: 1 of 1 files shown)


aclum:
I thought we were previously sorting on completed workflow results?

yxu-lanl:
The name is not unique. Will use id.
.order_by(self.table.model.multiomics.desc(), self.table.model.id)  # type: ignore

nmdc_a=# select count(*) from biosample;
 count 
-------
 13248
(1 row)

nmdc_a=# select count(DISTINCT name) from biosample;
 count 
-------
 13049
(1 row)

nmdc_a=# select count(DISTINCT (multiomics, name)) from biosample;
 count 
-------
 13051
(1 row)

naglepuff (approved):

PR #1723, opened 2025-08-06
Upgrade to nmdc-submission-schema v11.10.0-rc.1

### LINKED ISSUE #1722: Upgrade to `nmdc-submission-schema` `v11.10.0-rc.1` and migrate existing submission data
Version `v11.10.0-rc.1` of `nmdc-submission-schema` contains changes (https://github.com/microbiomedata/submission-schema/pull/347) which rename a number of slots in the JGI data classes. This task is to upgrade `nmdc-server` to that release candidate, including a migration of existing submission data to use the new slot names.

cc: @mslarae13

pkalita-lbl:
Fixes #1722 

These changes upgrade to `nmdc-submission-schema` v11.10.0-rc.1. That version introduces a number of slot name changes (summarized in https://github.com/microbiomedata/submission-schema/pull/347). The accompanying migrator changes existing submission data to match. Finally, there was one place where a few of the renamed/removed slots were referenced in front-end code, and those have been updated as well.

### DIFF
--- a/nmdc_server/migrations/versions/0cb677fa8d5e_update_jgi_submission-schema_slot_names.py
+++ b/nmdc_server/migrations/versions/0cb677fa8d5e_update_jgi_submission-schema_slot_names.py
@@ -6,6 +6,7 @@ Create Date: 2025-08-06 19:53:09.430148
 
 """
 
+from collections import namedtuple
 from typing import Optional
 from uuid import uuid4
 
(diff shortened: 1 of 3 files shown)


nmdc_server/migrations/versions/0cb677fa8d5e_update_jgi_submission-schema_slot_names.py
```diff
@@ -0,0 +1,101 @@
...
+    session.bulk_update_mappings(SubmissionMetadata, mappings)
+    session.commit()
+
+
+def upgrade():
+    do_rename({old: new for old, new in RENAMED_SLOTS})
+
+
+def downgrade():
+    do_rename({new: old for old, new in RENAMED_SLOTS})
```
eecavanna, line 101: I think this `downgrade` function has the potential to restore a field to a _different_ "old" value (compared to its _original_ value), since there are some duplicate "new" values in the mapping.

For example:
1. Original value: `dnase_rna`
2. "New" value: `dnase`
3. Restored "old" value: `dna_dnase`

(I don't mean to imply that I consider this to be a dealbreaker.)

eecavanna (approved):

PR #1729, opened 2025-08-11
Make Krona plots viewable in data portal

### LINKED ISSUE #1707: Data Portal - Embed readqc html files
Based on our quarterly planning meeting we will support embedding of html reports that are the output of ReadQcAnalysis.

Specifically we will support embedding the following three `data_object_type` values

```
Kraken2 Krona Plot
GOTTCHA2 Krona Plot
Centrifuge Krona Plot
```

Suggest getting input from @faiza-a on how this would look in the Data Portal. I was thinking that clicking on the data_object_type value would either open a new window with the html or would open it below the text for that data object type.

naglepuff:
Fix #1707 

#### Backend
A new API endpoint has been added to allow a client to obtain the URL for specific types of data objects. This is necessary because the existing endpoint is only accessible to authenticated (i.e. _logged in_) users. The Krona plot's HTML needs to be accessible for all users. Being logged in should not be required to view these data objects. 

This endpoint accepts a `data_object_id` and returns the URL for the HTML for that data object only if the data object has a specific `file_type`. If this endpoint is used to try and obtain the download URL for a non-Krona plot data object, it return a `400`. Tests have been added for this endpoint.

#### Frontend
Krona plots are now viewable on the data portal without the need to log in or click the download button. They are accessible by clicking the magnifying glass icon in the corresponding row:
image

Clicking will bring up a modal (implemented with a `v-dialog` component) containing an [iframe](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/iframe).
image

### DIFF
--- a/web/src/components/DataObjectTable.vue
+++ b/web/src/components/DataObjectTable.vue
@@ -2,7 +2,7 @@
 // @ts-ignore
 import NmdcSchema from 'nmdc-schema/nmdc_schema/nmdc_materialized_patterns.yaml';
 import {
-  computed, defineComponent, PropType, reactive, ref, Ref,
+  computed, defineComponent, PropType, reactive, ref, Ref, watch,
 } from '@vue/composition-api';
 import { flattenDeep } from 'lodash';
 
@@ -95,6 +95,7 @@ export default defineComponent({
     const selectedHtmlDataObject: Ref<any | null> = ref(null);
     const dataModal = ref(false);
     const iframeDataSource = ref('');
+    const iframeLoading = ref(false);
     function hasHtmlData(fileType: string) {
       return [
         'Kraken2 Krona Plot',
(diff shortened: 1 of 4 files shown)


web/src/components/DataObjectTable.vue
```diff
@@ -87,6 +92,33 @@ export default defineComponent({
       },
     ];
 
+    const selectedHtmlDataObject: Ref<any | null> = ref(null);
+    const dataModal = ref(false);
+    const iframeDataSource = ref('');
+    const iframeLoading = ref(false);
+    function hasHtmlData(fileType: string) {
+      return [
```
codytodonnell, line 100: We may eventually want to have a variable for this array of strings like `supportedEmbedTypes`. It would be nice if we didn't have to repeat it on the backend too and could have a single source of truth.

naglepuff: #1736

PR #1737, opened 2025-08-14
Add google event tracking for active query items

### LINKED ISSUE #1698: Add ability to report statistics on user search behavior
I would be useful to answer questions like "what is the most used search facet?"

Potential solutions include:
- Adding google analytics to the front end (sidebar) components
- Keeping a record in postgres of use searches?

codytodonnell:
Resolves #1698 

Creates a custom `useGtag` hook from which we can track custom events inside any component. The first place this is being implemented is in the `SearchSidebar` and an event is triggered every time a new filter is added or updated. On the initial load, an event will be triggered for each active filter.

Note that we also created two Google Analytics "properties" for the NMDC Data Portal. One is for the production site. One is for the dev site as well as local development. Each GA property has its own GA ID associated with it. This ID gets pulled into the vue app from an environment variable called `VUE_APP_NMDC_GOOGLE_ANALYTICS_ID`.

### DIFF
--- a/web/src/use/useGtag.ts
+++ b/web/src/use/useGtag.ts
@@ -4,6 +4,7 @@ import { getCurrentInstance } from '@vue/composition-api';
  * Custom hook to access Google Analytics gtag instance.
  * returns $gtag instance if available, otherwise undefined.
  * In development, this will return undefined.
+ * Note that this must be used within the setup() function of a component.
  */
 export default function useGtag() {
   const instance = getCurrentInstance();
--- a/web/src/views/Search/SearchSidebar.vue
+++ b/web/src/views/Search/SearchSidebar.vue
@@ -185,34 +185,34 @@ export default defineComponent({
     }
     watch(filterText, updateSearch);
 
-    function trackFilterConditions(val: Condition[], oldVal: Condition[]) {
+    function trackFilterConditions(newConditionList: Condition[], oldConditionList: Condition[]) {
       // Do nothing if Google Analytics is not available. This is expected in development mode.
       if (!gtag) {
         return;
       }
       // On initial load, track each filter condition that exists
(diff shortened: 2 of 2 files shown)


web/src/use/useGtag.ts
```diff
@@ -0,0 +1,12 @@
+import { getCurrentInstance } from '@vue/composition-api';
+
+/**
+ * Custom hook to access Google Analytics gtag instance.
+ * returns $gtag instance if available, otherwise undefined.
+ * In development, this will return undefined.
+ */
```
naglepuff, line 7: I think it would be helpful to add to the comment that this function can [only be used inside of a `setup()` function](https://github.com/vuejs/composition-api/issues/455#issuecomment-667702942) in a vue component.

web/src/views/Search/SearchSidebar.vue
```diff
@@ -183,6 +185,41 @@ export default defineComponent({
     }
     watch(filterText, updateSearch);
 
+    function trackFilterConditions(val: Condition[], oldVal: Condition[]) {
```
naglepuff, line 188: What do you think about changing `val` and `oldVal` to be something like `newConditionList` and `oldConditionList`? I think that might make it easier to read through this function.

naglepuff (approved):
LGTM! I had a couple of minor suggestions. I also want to point out that I think this might need to change eventually as part of the migration to Vue 3. It seems like the `getCurrentInstance` function is poorly documented for Vue 3 and may be intended for internal use. 

We'll have to follow the latest `vue-gtag` documentation when upgrading.

web/src/use/useGtag.ts
codytodonnell: Good call I will add that to the comment.

PR #1738, opened 2025-08-19
Resize visualization components

### LINKED ISSUE #1524: upset plot is looking squished
Is there a way to make the entire upset box area slightly bigger?

Image

 The numbers are starting to run into each other now that we also have lipidomics data.

naglepuff:
Fix #1524 

### Changes
This is a simple and straightforward change that adjusts the height and width of the visualizations that appear underneath the map and data type bar chart on the data portal. The addition of new data types caused the upset plot to be difficult to read:
image

Additionally, much of the collection date histogram is effectively whitespace (looking at pre-2014). To remedy the readability of the upset plot, the heights of these components has been increased (they also now match), and each component now takes up an equal amount of horizontal space:
image

Now each bar in the upset plot is given more space.

### DIFF
--- a/web/src/components/Presentation/DateHistogram.vue
+++ b/web/src/components/Presentation/DateHistogram.vue
@@ -36,6 +36,10 @@ export default Vue.extend({
       type: String,
       required: true,
     },
+    height: {
+      type: Number,
+      default: 160,
+    },
   },
 
   data() {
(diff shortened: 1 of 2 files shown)


naglepuff:
@aclum please let me know if this sufficiently addresses the issue. I can provide additional screenshots if needed.

@codytodonnell please take a look at the changes and let me know if anything seems off to you.

codytodonnell (approved):
The resized plot looks great to me. and code changes make sense.

I believe I found another upset plot issue (exists in `main` not just this branch). The issue is with the applied multiomics filters not matching my expectation when I click on a bar from the upset plot. For example, in my data the first bar on the top looks to be associated with "MP" but when I click that bar, the applied filter is "MT." There are several mismatches like this when I click through the bars.

PR #1740, opened 2025-08-19
Fix multiomics bitmask

### LINKED ISSUE #1739: Multiomics bitmasks calculated incorrectly
Reported by @simroux. This affects searching in the data portal via the upset plot. For example, clicking the "MG"-only bar will start a search for "Multiomics [has] MP"

naglepuff:
Fix #1739 

### Changes
Updates the `MultiomicsValue` object, which maps different data types (e.g. `MB` for metabolomics) to the corresponding bit in the `multiomics` bitmask used by the data portal for searching and displaying the upset plot.

This PR adds "Amplicon" and adjusts the bitmasks for the other data types accordingly.

### DIFF
--- a/web/src/encoding.ts
+++ b/web/src/encoding.ts
@@ -674,12 +674,13 @@ function getField(name: string, table?: entityType): FieldsData {
 }
 
 const MultiomicsValue = {
-  MB: 0b100000,
-  MG: 0b010000,
-  MP: 0b001000,
-  MT: 0b000100,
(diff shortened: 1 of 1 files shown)


naglepuff:
Thanks for the review. On this branch in my local environment, I see this upset plot:
image

I see MB, so it could be a mismatch of data between our environments. As far as adding a comment goes, I agree, its a little bit arcane. I opened up #1741 to track.

Issue #1721, opened 2025-08-05
restrict editing to certain submission statuses


samobermiller:
Edits can only be made to submissions if status is `InProgress` or `UpdatesRequired`

PR #1742, opened 2025-08-21
1721 restrict editing to certain submission statuses

samobermiller:
this PR closes issue #1721. NOTE: to be merged in alongside PR #[tbd] that will enable drop down menu for status changes

Currently, a submission can continue to be edited after hitting the Submit button and there is no functionality to the `UpdatesRequired` submission status. 

Updates:
- [x] Edits are now only allowed if a submission's status is `InProgress` or `UpdatesRequired`
   - This means you can no longer edit a submission after hitting the Submit button (when the status is `SubmittedPendingReview`)
   - This is communicated by a banner at the top of each page (superseding all other banners) saying edits aren't permitted due to status
- [x] Submissions are view-only if status is anything other than `InProgress` or `UpdatesRequired`
   - `Resume` button now says `View` with an eye symbol
- [x] If a submission's status is `UpdatesRequired` and edits are made, the status becomes `InProgress` again
   - Montana will manually assign a status of `UpdatesRequired` after review via a yet to be implemented drop down
- [x] Github issues are now only created if there is no other issue in the repo with the same submission id. If there is an existing issue, a comment is added saying it was resubmitted on x date and the issue is reopened if closed (test created in test_submission.py)
- [x] Restrict edits via API based on submission status (test created in test_submission.py)

Note: status does not change to `InProgress` from `UpdatesRequired` if admin makes edits, this is intentional since `UpdatesRequired` indicates we need something from the submitter

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -62,3 +62,8 @@ NMDC_GCP_SECONDARY_POSTGRES_URI_SECRET_ID=changeme
 # See the invocations of the `replace_nersc_data_url_prefix` function for additional context.
 #NMDC_ZIP_STREAMER_NERSC_DATA_BASE_URL="https://data.microbiomedata.org/data"
 #NMDC_NERSC_DATA_URL_EXTERNAL_REPLACEMENT_PREFIX="https://data.microbiomedata.org/data"
+
+# Github issue creation, populate to enable automated github issue creation
+#NMDC_GITHUB_ISSUE_URL = https://api.github.com/repos/microbiomedata/issues/issues
+#NMDC_GITHUB_AUTHENTICATION_TOKEN = << personal github token >>
+#NMDC_GITHUB_ISSUE_ASSIGNEE = << github username >>


naglepuff (commented):
Looks good! I was not able to do any functional testing for the github endpoint, but tests of the submission portal components looked good. I had a few minor code comments is all.

pkalita-lbl (commented):
Changes look good overall.

As I was testing I noticed that I wasn't seeing the status automatically change from `UpdatesRequired` to `InProgress`. I could see that the `InProgress` status was part of the `PATCH /api/metadata_submission/{id}` request, but the response still had `UpdatesRequired` in it. I think the source of that bug is in the `update_submission` route handler function gates the status change on checking if the logged-in user has role of owner on the submission. In my case I was editing someone else's submission as an admin. It's an edge case, but it might be good to clean that up now as we start making more extensive use of statuses.

mslarae13 (commented):
lgtm from a totally non technical perspective. 
The description's expected behaviors are accurate.

pkalita-lbl (approved):
Overall looks good!

I have a slight concern about how existing GitHub issues are being found. As it stands it's only going to look at the 100 most recently created issues in the `issues` repo: https://github.com/microbiomedata/nmdc-server/blob/c230b4c8980f241d46fe8e6e9a04ef69491c4e61/nmdc_server/api.py#L1407-L1413

I have a few ideas of how we could do better. I'm going to make an issue about it, and we can handle it separately. This PR already has enough going on.

Edit: issue created https://github.com/microbiomedata/nmdc-server/issues/1821

PR #1748, opened 2025-08-27
Use `datetime.now(UTC)`

naglepuff:
This clears up >100 warnings emitted while running pytest. It moves our code away from datetime usage that was deprecated in python 3.12.


marySalvi (approved):
LGTM

PR #1749, opened 2025-09-02
Remove `worker` and `redis` from stack

eecavanna:
On this branch, I removed the Celery worker and its Redis server from the application stack. 

Nowadays, for local development and production use, the ingest script is invoked via an instance of the `nmdc-server` CLI application running within an instance of the `backend` container (in production, that container is spun up via a Kubernetes CronJob).


naglepuff (approved):
Looks pretty thorough to me!

pkalita-lbl (approved):

PR #1752, opened 2025-09-04
Change created field on auth code to be timezone aware

marySalvi:
Fixes minor typing bug when comparing times.


pkalita-lbl (approved):
I tested locally and login/logout worked as expected. We should make an issue about making other `DateTime` columns timezone aware.

Issue #1768, opened 2025-09-24
Update navigation menu to match WordPress website


eecavanna:
- [ ] Rename the top-level menu, "Partner with us", to "Get involved"
- [ ] In the top-level menu named "Resources", add a menu item named "Learning Resources" that links to https://microbiomedata.org/certificate-program/

FYI: @pkalita-lbl and @codytodonnell

PR #1769, opened 2025-09-24
Update navigation menu to match WordPress website

eecavanna:
On this branch, I made the two changes listed in #1768; i.e. I renamed a top-level menu item and introduced a new second-level menu item.

### DIFF
--- a/web/src/menus.ts
+++ b/web/src/menus.ts
@@ -84,10 +84,14 @@ const Menus = [
         label: 'Data Integration',
         href: 'https://microbiomedata.org/data-integration/',
       },
+      {
+        label: 'Learning Resources',
+        href: 'https://microbiomedata.org/certificate-program/',
+      },
     ],
   },
   {
-    label: 'Partner with Us',
+    label: 'Get Involved',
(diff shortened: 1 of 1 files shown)


eecavanna:
The GHA workflow that runs the server-side tests is failing. I think it's failing because it can't download the `postgres:13` Docker container image. I visited https://hub.docker.com/_/postgres just now and got an error page. I think Docker Hub is having some issues.

eecavanna:
FYI, @codytodonnell, this is what it looks like to make changes to the navigation menu that appears at the top of the Data Portal. Every now and then, a member of our engagement team will reach out to me and/or @pkalita-lbl via Slack and request that we make a change to that menu on the WordPress site (such a request came in earlier today). Whenever we make a change to it on the WordPress site (which we did this afternoon), we make an accompanying change on the Data Portal and on NMDC EDGE, since those web apps have their own local "copies" of that menu.

eecavanna:
The GHA workflow is passing now. I think the earlier failure was due to an issue on Docker Hub's end (I learned this evening that other people were having issues accessing Docker Hub today also). Merging.

PR #1773, opened 2025-09-25
Upgrade vue2.7

marySalvi:
This upgrades the frontend to Vue2.7. I tried to keep a clean, logical commit history so it should be possible to review the changes in commit order. 

Please review the UI extensively. Look for style changes, routing issues and prop passing issues in particular. All type, build and lint errors should be resolved.


eecavanna:
Thanks, @marySalvi!

Note to reviewers: I don't expect to look at this before the end of the quarter, due to existing workload.

PR #1777, opened 2025-10-01
Add frontend support for uploading PI and primary study images

### LINKED ISSUE #1608: Add ability to upload an image to NMDC submissions via submission portal

pkalita-lbl:
Fixes #1608 

### Summary

This PR implement frontend changes to allow users to upload a PI headshot and/or a primary study image via the Submission Portal. The upload process uses the Google Cloud Storage-based endpoints implemented in #1706.

### Details

* A new component (`web/src/views/SubmissionPortal/Components/ImageUpload.vue`) handles the user interactions of letting users choose an image, display a preview of the chosen image, upload the chosen image, or change a previously uploaded image.
* The Submission Portal study information form (`web/src/views/SubmissionPortal/Components/StudyForm.vue`) has been slightly rearranged to accommodate two usages of the `ImageUpload` component: one for the primary study image and one for the PI headshot image. This is also where the read/write interactions with the Submission Portal store (`web/src/views/SubmissionPortal/store/index.ts`) takes place.
* Functions for calling the API endpoints added in #1706 have been added to `web/src/views/SubmissionPortal/store/api.ts`, along with the necessary type definitions in `web/src/views/SubmissionPortal/types.ts`.
* The one backend change here is to add a `POST /metadata_submission/{id}/image/make_public` endpoint. This is an _admin-only_ endpoint that will copy submission images into the `nmdc-public-images` bucket. This endpoint will be called as part of the process which translates a submission into `nmdc-schema`-compatible objects and inserts them into MongoDB. This is because the images must be publicly accessible at that point.


nmdc_server/api.py
```diff
@@ -1492,6 +1492,52 @@ async def generate_signed_upload_url(
     )
 
 
+@router.post(
+    "/metadata_submission/{id}/image/make_public",
+    response_model=schemas.SubmissionImagesMakePublicResponse,
+)
+async def make_submission_images_public(
```
eecavanna, line 1499: After I read this function, I wondered whether adding a test that targets this function (endpoint) would be feasible. One test could try accessing it as a non-admin. Another test could exercise the happy path. It's the second test that I'm not sure about, as I don't remember the limitations of the "local" GCS server that is used during testing.

eecavanna (approved):
The code, itself, looks good to me. I have not tried spinning it up and viewing the new components or testing them out. If you happen to have screenshots of the image upload widgets handy, I'd be interested in seeing them. No need to hold up merge for that, though, from my perspective. I left a comment about adding a test that targets the API endpoint.

codytodonnell (approved):
This all looks good to me. I pulled down your branch and can confirm that uploading/removing images works locally. It's not clear to me when the `.../make_public` endpoint gets hit so that part I don't think I tested in my local workflow.

pkalita-lbl:
> It's not clear to me when the .../make_public endpoint gets hit 

It will get called in the process of translating a submission into MongoDB records. The core of that process is [this class](https://github.com/microbiomedata/nmdc-runtime/blob/8eada37fb28b0b91aa1073c6f3ae38e5b61817db/nmdc_runtime/site/translation/submission_portal_translator.py#L128) in `nmdc-runtime`. I say "will" because the `nmdc-runtime` changes still need to be made.

nmdc_server/api.py
pkalita-lbl: Tests added!

eecavanna: Thanks. The tests look great—I found it easy to keep track of what they were doing, while I read their code just now.

PR #1813, opened 2025-10-15
Add API endpoint that returns URLs of all `DataObject`s output by any `WorkflowExecution`

eecavanna:
On this branch, I implemented an API endpoint that admins can use to get a list of the URLs of all the `DataObjects` that are referenced by any `WorkflowExecution`s. The endpoint is: `GET /api/admin/data_object_report`.

image

#### Input (request)

The endpoint accepts one parameter: `variant`. That parameter can be used to choose between two types of reports: "normal" and "URLs only". The "normal" report includes three columns—the data object's "id", "url", and "size"—and can be expanded in a future PR to include more columns based upon team member feedback. The "URLs only" variant includes only the "url" column.

#### Output (response)

The actual column names are more specific—they are (literally):

| data_object.id | data_object.url | data_object.file_size_bytes |
| --- | --- | --- |

The endpoint returns a TSV file. The file consists of a table that lists those 1 or 3 things about each `DataObject` that is an output of any `WorkflowExecution`.

#### Authorization

The endpoint is only accessible to users who are admins. Other users will get an HTTP 403 Forbidden response.

### Background

The real-world use case for this endpoint is: A list of the URLs of `DataObject`s that are outputs of any `WorkflowExecution` can be used to locate to corresponding files on the NERSC filesystem. Once located, those files can be processed in bulk (e.g. copied to a Google Cloud Storage bucket). We did exactly that a few days ago, although the list of URLs was generated via a different means (and was derived from the upstream MongoDB database; not this downstream Postgres database that is dependent upon an ingest having taken place).

FYI: @shreddd

### DIFF
--- a/nmdc_server/aggregations.py
+++ b/nmdc_server/aggregations.py
@@ -197,10 +197,7 @@ def get_wfe_output_data_objects(db: Session) -> List[models.DataObject]:
     wfe_outputs_subquery = make_all_wfe_outputs_subquery(db)
     wfe_outputs_inner_query = select(wfe_outputs_subquery.c.id)  # type: ignore
     q = db.query(models.DataObject).filter(models.DataObject.id.in_(wfe_outputs_inner_query))
-    data_objects = []
-    for data_object in q.all():
-        data_objects.append(data_object)
-    return data_objects
+    return q.all()
 
 
 def get_sankey_aggregation(
--- a/nmdc_server/api.py
+++ b/nmdc_server/api.py
@@ -191,7 +191,7 @@ async def get_data_object_report(
     of any `WorkflowExecution`.
     """
 
-    [header_row, data_rows] = crud.get_data_object_report(db, variant=variant)
+    header_row, data_rows = crud.get_data_object_report(db, variant=variant)
 
     # Build the report as an in-memory TSV "file" (buffer).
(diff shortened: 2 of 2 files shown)


eecavanna:
Thanks for reviewing the changes. I expect this endpoint to be used as part of the following "semi-automated" process. The process involves passing a (path to a file containing) a list of filesystem paths of NERSC-hosted data files, to a [standalone script](https://github.com/microbiomedata/nmdc-cloud-deployment/blob/main/scripts/nmdc_data_management/README.md), which will generate a tree of symlinks to those files; and then using `$ gcloud storage rsync ...` to sync that symlink tree to a GCS bucket. 

I chose to make the endpoint respond with a file because I expect to always want to store the URL list as a file. Having a file containing a list of URLs gets me close to having the first input I mentioned above; i.e. having a file containing a list of filesystem paths (I can get from one to the other with search-and-replace).

I chose the TSV format because that's the format used by a couple other report-generation endpoints.

eecavanna:
Thanks for reviewing the code and testing out the endpoint! Yes, ~30 seconds is roughly what I expect for this, based on my own local testing with a recent production dump.

JamesTessmer:
Tacking on a little later, but I tested earlier and didn’t run into any problems.

PR #1823, opened 2025-10-27
Update the call to geoloc-tools `elevation` function

hesspnnl:
The updated function now requires a Google Map API key with access to the elevation endpoints. Update the server code to load the the api key and pass it to the updated function.

### DIFF
--- a/tests/test_submission.py
+++ b/tests/test_submission.py
@@ -896,8 +896,8 @@ def test_metadata_suggest(client: TestClient, suggest_payload, logged_in_user):
     )
     assert response.status_code == 200
     assert response.json() == [
-        {"type": "add", "row": 1, "slot": "elev", "value": "16", "current_value": None},
-        {"type": "replace", "row": 3, "slot": "elev", "value": "16", "current_value": "0"},
+        {"type": "add", "row": 1, "slot": "elev", "value": "16.00", "current_value": None},
+        {"type": "replace", "row": 3, "slot": "elev", "value": "16.00", "current_value": "0"},
     ]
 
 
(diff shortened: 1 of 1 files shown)


.env.example
```diff
@@ -50,6 +50,10 @@ NMDC_GCS_USE_FAKE=True
 NMDC_GCP_PRIMARY_POSTGRES_URI_SECRET_ID=changeme
 NMDC_GCP_SECONDARY_POSTGRES_URI_SECRET_ID=changeme
 
+# == Google Map API Key ==
+# the google map api key is currently used to pass to geoloc tool to get elevation data. This key must have some value in it to run tests
+GOOGLE_MAP_API_KEY=thiscanbeanythingfortesting
```
pkalita-lbl: Can you add the `NMDC_` prefix to this variable name? It works as-is because we have `extra="allow"` in the settings config: https://github.com/microbiomedata/nmdc-server/blob/e1a29e1dbbfb16d75c856151b115b85690fd0704/nmdc_server/config.py#L201-L205

But having the prefix is a good indicator that it's a variable used in the `nmdc-server` codebase.

.github/workflows/server.yml
```diff
@@ -68,6 +68,7 @@ jobs:
         echo 'export NMDC_GCS_USE_FAKE="true"' >> .env
         echo 'export NMDC_GCS_OBJECT_NAME_PREFIX="testing"' >> .env
         echo 'export NMDC_GCS_FAKE_API_ENDPOINT="http://localhost:4443"' >> .env
+        echo 'export GOOGLE_MAP_API_KEY=testing"' >> .env
```
pkalita-lbl: Ditto re: adding `NMDC_` prefix here.

pkalita-lbl (commented):
Looking good! Two more minor details.

PR #1856, opened 2025-11-21
Fix EMSL shipping date inconsistency

### LINKED ISSUE #1237: Date is not consistent between EMSL shipping form and shipping info summary
To reproduce, create a new submission. 

On the context form select "No", then "EMSL"

On the shipping form, pick a date. Close the menu by clicking "save"

Expand the "Shipper info" section of the summary

The date will not match the one you selected for the form (it will be off by one day).

cristina-stonepedraza:
Closes #1237 

To resolve this, I specified using the UTC time for the date saved when retrieving it, and then formatted the date to YYYY-MM-DD. The date selected in the shipping form and the date displayed with shipping details now match.

### DIFF
--- a/web/src/views/SubmissionPortal/utils.ts
+++ b/web/src/views/SubmissionPortal/utils.ts
@@ -1,4 +1,3 @@
-// @ts-ignore
 import moment from 'moment';
 
 export function formatShippingDate (date: string | Date | null | undefined): string {


cristina-stonepedraza:
@mslarae13 Is YYYY-MM-DD the preferred date format?

mslarae13 (commented):
YYYY-MM-DD is good
@pkalita-lbl for technical check

pkalita-lbl (changes requested):
Can you create a new `web/src/views/SubmissionPortal/utils.ts` file and define `formatShippingDate` there, so that the definition is not duplicated in the two components?

cristina-stonepedraza:
> Can you create a new `web/src/views/SubmissionPortal/utils.ts` file and define `formatShippingDate` there, so that the definition is not duplicated in the two components?

Yes, I'll make that change now!

pkalita-lbl (approved):

PR #1873, opened 2025-12-02
Add SingleM krona plot rendering

### LINKED ISSUE #1865: add singleM krona plot to get_data_object_html_content function
We will be updating the reads based taxonomy workflow to use a new tool so we need to add that permissible value to the list of data objects to embed the html.

`SingleM Krona Plot`

codytodonnell:
Resolves #1865 

This adds the `SingleM Krona Plot` file type to the list of file types to be rendered as HTML in a clickable dialog from the data objects table.

Also fixes the sizing of the krona plot dialogs that were affected by vue upgrade.


eecavanna (approved):
I read the diff and everything looks intentional to me, and consistent with the PR description. I requested a review from Copilot just now also.

PR #1876, opened 2025-12-05
Change submission status to In Progress regardless of who makes updates

samobermiller:
Closes https://github.com/microbiomedata/issues/issues/1497
- `UpdatesRequired` status is updated to `InProgress` when submission is saved, regardless of if user has a current role on the submission
- The status change from `UpdatesRequired` to `InProgress` was being captured in postgres but required the user to refresh the page so the Submit button was available. I added a simple endpoint to get submission status from postgres so `status` can be refreshed on Harmonizer.vue. 
- The logic for the `Submit` button on Harmonizer.vue was getting more confusing, so I reformatted the various factors (owner/admin, submission status, validated tabs, already submitted) are all addressed in one variable SubmissionState

### DIFF
--- a/nmdc_server/api.py
+++ b/nmdc_server/api.py
@@ -1348,21 +1348,6 @@ async def update_submission_status(
     return submission
 
 
-@router.get(
-    "/metadata_submission/{id}/status",
-    tags=["metadata_submission"],
-    responses=login_required_responses,
-)
-async def get_submission_status(
-    id: str,
-    db: Session = Depends(get_db),
-    user: models.User = Depends(get_current_user),
-):
-    """Get current submission status"""
-    submission = get_submission_for_user(db, id, user)
-    return {"status": submission.status}
-
-
 @router.post(
     "/metadata_submission/{id}/role",
     tags=["metadata_submission"],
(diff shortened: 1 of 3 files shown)


pkalita-lbl:
@samobermiller I pushed a few commits that do the following:

* Update the `incrementalSaveRecord` and `submit` functions to take the response body of API requests and feed it back into frontend state. This allows the frontend to stay in sync with modifications that the backend makes. The advantage here is that we can expand the types of automatic backend updates we do in the future without (hopefully) having to make additional frontend changes. This did require updating a few `computed` and `watch` callbacks to be a little more picky about when we update reactive state. There is still a bit of a pre-existing reactive rat's nest going on here that I'm not prepared to fully detangle  in this PR.
* Expand the `submissionState` computed ref that you created so that specific reasons can be shown in the tooltip displayed with the Submit button is disabled.
* Remove the now-unused `GET /api/metadata_submission/{id}/status` endpoint 😢 

Please give that a test and let me know if you see any issues.

samobermiller:
Looks good to me! Are you ready for it to be merged in @pkalita-lbl ?

PR #1878, opened 2025-12-05
Account for AppBanner size when computing DataHarmonizer container height

### LINKED ISSUE #1875: Unable to see bottom-most rows of DataHarmonizer
I think this might be fallout from the Vuetify 3 upgrade.

On prod I can scroll all the way to see the last (100 by default) row in DataHarmonizer:
Image

On dev, scrolling all the way to the bottom only gets you to about row 96. It appears that the footer (with the "add more rows" control and color key may be overlaid atop the remaining rows:
Image

pkalita-lbl:
Fixes #1875 

DataHarmonizer needs a container (not necessarily the direct parent, but some ancestor element) with a fixed height in order to work properly. We set the height on the element that contains the stepper header, the validation controls, DataHarmonizer itself and the help sidebar, and the footer. Currently we set the height of that element as the window height (`100vh`) minus the main application bar height (`APP_HEADER_HEIGHT`). The problem is that this doesn't take into account the height of the site banner which is sometimes shown. This element is outside of the fixed-height container, so it needs to be taken into account as well.

These changes are a little more complex than I would have hoped, but other options I looked at would involve touching the layout of _all_ pages. Here's the summary of what's happening here.

1. The site banner (`AppBanner`) now computes its own height and exposes that as a value that the component that renders `AppBanner` can access.
2. `SubmissionView` renders some common elements for all Submission Portal pages, including `AppBanner`. It uses a `ref` to get the height of the `AppBanner` component and uses [`provide`](https://vuejs.org/guide/components/provide-inject.html#provide-inject) to make that value available to its component tree.
3. `HarmonizerView` is where the `AppBanner` height is needed, so it uses `inject` to get the value. It uses the value when setting the height of the fixed-height container.

### DIFF
--- a/web/src/components/AppBanner.vue
+++ b/web/src/components/AppBanner.vue
@@ -1,5 +1,5 @@
 <script lang="ts">
-import { computed, defineComponent } from 'vue';
+import { computed, defineComponent, onUpdated, ref, useTemplateRef } from 'vue';
 import { stateRefs } from '@/store';
 
 export default defineComponent({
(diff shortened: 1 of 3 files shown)


eecavanna:
(I like the way the banner looks — I just mean that we may be able to overhaul how it's implemented in order to make it easier to integrate into the submission portal.)

PR #1879, opened 2025-12-06
Add submitter email field to form and add modal to collect user email upon login

cristina-stonepedraza:
Addresses https://github.com/microbiomedata/issues/issues/1106

1) Make the submitter email required in the Submission Portal form

- Added a new field for this in the form (StudyForm.vue), and made it required
- Added appropriate definitions for new field (index.ts, definitions.ts)
- The field is auto-populated using the email saved in the logged-in user's details (saved in db) 

2) Re-worked logic in auth.py

- Re-ordered email check (now checks if email is missing first, then checks for ORCID email instead of the other way around)
- Changed API return from XML to JSON

3) Added modal to get the user's email if it is still missing after login (this is required to create a portal submission)

- Created a new file at /web/src/views/SubmissionPortal/Components/UserEmailModal.vue
- Added check (for user login and saved email) and modal call to App.vue
- Expected behavior: After login, a check is made from App.vue to see if the user is logged in and if they have an email saved. If not, then the modal will pop up, and the user will need to enter an email to close it.

### DIFF
--- a/web/src/views/User/UserDetailPage.vue
+++ b/web/src/views/User/UserDetailPage.vue
@@ -24,6 +24,8 @@ if (refreshToken != null) {
 const isTokenVisible = ref(false);
 const isCopyRefreshTokenSnackbarVisible = ref(false);
 
+const origin = window.location.origin;
+
 const handleRefreshTokenCopyButtonClick = async () => {
   if (refreshToken != null) {
     await navigator.clipboard.writeText(refreshToken);


web/src/views/SubmissionPortal/Components/StudyForm.vue
```diff
@@ -122,6 +125,48 @@ export default defineComponent({
...
+    //       editEmail.value = !editEmail.value;
+    //     }
+    //   } else {
+    //     editEmail.value = !editEmail.value;
+    //   }
+    // };
+
+    const submitterEmail = ref(user.value?.email || '');
+
+    studyForm.submitterEmail = submitterEmail.value;
```
pkalita-lbl: I think we need to discuss this from a requirements perspective. As it stands now, this means that if User A creates the submission then `submitterEmail` will be `user.a@example.com`. But if they add User B as an editor of the submission, when User B logs in and open the submission, `submitterEmail` will change to `user.b@example.com`. I don't think that's what we want. 

I think updating the email address of the logged in user's user profile based on this form could also lead to confusion when considering submissions that are editable by more than one user.

web/src/views/SubmissionPortal/Components/UserEmailModal.vue
```diff
@@ -0,0 +1,159 @@
...
+        is_admin: user.value?.is_admin as boolean,
+      };
+      await api.updateUser(user.value?.id as string, update);
+    };
+
+    const editEmail = ref(false);
+    const isEmailValid = ref(false);
+
+    const submitterEmail = ref(user.value?.email || '');
+    studyForm.submitterEmail = submitterEmail.value;
```
pkalita-lbl: Since this component is used directly by `App.vue`, it's unclear which submission state is being modified here. I guess just a copy of the default form object? I think that's not necessary to do here. That can be done (if needed at all) in Submission Portal-specific code.

web/src/views/SubmissionPortal/Components/UserEmailModal.vue
```diff
@@ -0,0 +1,159 @@
...
+
+<template>
+  <teleport to="body">
+    <v-dialog
+      v-model="dialog"
+      persistent
+      max-width="600"
+    >
+      <v-card>
+        <v-card-title class="headline">
```
pkalita-lbl: I don't see where the `headline` CSS class is defined and I don't believe it comes from Vuetify. Since the style of the title doesn't really need to deviate from the default (IMO), I would recommend just using the `title` prop on `v-card` instead.

web/src/App.vue
```diff
@@ -1,10 +1,21 @@
-<script lang="ts" setup>
-import { onMounted, onUnmounted } from 'vue';
+<script lang="ts">
```
pkalita-lbl: Is there a reason to change this from the `<script setup>` style back to the `defineComponent` style?

pkalita-lbl (changes requested):
The basic workflow of logging in, populating the `user_logins.email` column from the ORCID API response, and prompting the user with a modal as a fallback works well.

I have some concerns about the requirements around populating `user_logins.email` from a submission. I think that might have unintended consequences when you have submissions editable by more than one user.

web/src/views/SubmissionPortal/Components/StudyForm.vue
cristina-stonepedraza: Requirements discussed: 

- Use "author" instead of "submitter" 
- "Author email" will not be editable from the submission portal
- Add "Author" section below "Principal Investigator"
- If the author is logged in and editing the submission, a note will appear by the email info - "If this does not look correct, click here [author's user profile] to change email"
- If the user logged in is NOT the author, then author section/related fields will be read-only, with note "only the author can edit this field"

web/src/components/UserEmailModal.vue
```diff
@@ -0,0 +1,159 @@
...
+      updateUser,
+      updateEmail,
+      dialog,
+    };
+  },
+});
+</script>
+
+<template>
+  <teleport to="body">
```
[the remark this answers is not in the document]
pkalita-lbl: Sorry I wasn't questioning the `<v-dialog>` usage; that looks correct to me. I was just wondering if it really needs to be wrapped in `<teleport>`.

pkalita-lbl:
I'll work on resolving the post-merge issues here.

Issue #1889, opened 2025-12-16
Implement health check API endpoint that checks database connection


eecavanna:
Implement an API endpoint that can be used for health checks. Have the endpoint check whether the FastAPI app can access the Postgres database.

Here's a link to the `nmdc-runtime` counterpart of this ticket: https://github.com/microbiomedata/nmdc-runtime/issues/1348

This endpoint will be used by a Kubernetes `readinessProbe`. More info: https://github.com/microbiomedata/nmdc-cloud-deployment/issues/202

PR #1890, opened 2025-12-16
Implement `/api/health` API endpoint that checks database connection

eecavanna:
On this branch, I implemented a dedicated health check API endpoint, which is at `/api/health`. Unlike the `/version` API endpoint, this new endpoint checks whether the FastAPI app can talk to the Postgres server. I plan to use this new endpoint for Cloudflare health checks and for Kubernetes `readinessProbe`s.

The reason the `HealthResponse` (Pydantic model) class's definition is in the `api.py` file is that the `models.py` file seemed to me to contain only SQLAlchemy models—not Pydantic models—and the Pydantic models I saw elsewhere were in the same files as the functions that use them. The only function that uses _this_ Pydantic model is in `api.py`.

An analogous endpoint in `nmdc-runtime` was introduced into that application [here](https://github.com/microbiomedata/nmdc-runtime/pull/1349).

### Screenshots

When database connection is healthy, the endpoint returns an HTTP 200 response with both properties set to `true`.

image

When database connection is unhealthy, the endpoint returns an HTTP 503 response with the database property set to `false`.

image

### Related issues

Fixes: #1889

### DIFF
--- a/nmdc_server/api.py
+++ b/nmdc_server/api.py
@@ -14,7 +14,6 @@ from fastapi import APIRouter, Depends, Header, HTTPException, Query, Response, 
 from fastapi.responses import JSONResponse
 from linkml_runtime.utils.schemaview import SchemaView
 from nmdc_schema.nmdc import SubmissionStatusEnum
-from pydantic import BaseModel, ConfigDict, Field
 from sqlalchemy import text
 from sqlalchemy.orm import Session
 from starlette.responses import StreamingResponse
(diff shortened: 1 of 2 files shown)


nmdc_server/api.py
```diff
@@ -49,6 +51,58 @@
 logger = get_logger(__name__)
 
 
+class HealthResponse(BaseModel):
```
pkalita-lbl: I would recommend defining this in the `schemas.py` module. That's where other Pydantic models used as API response models are defined.

eecavanna: Or, rather, I forgot about that file and didn't notice it when I did a search today.

Moved in commit https://github.com/microbiomedata/nmdc-server/pull/1890/commits/c0d81a5247a779ad21cfcbb2e32fa9a842d0e5a4! (I'll merge this in when the GHA checks pass.)

PR #1907, opened 2025-12-23
Add dynamic max-height for study image

### LINKED ISSUE #1895: Data portal - limit consortium logo size
env: development

There is a large whitespace between  'data summary' and 'team' for 1000 soils, see below. Per Patrick. we should set a maximum height for these.

Image

to reproduce: https://data-dev.microbiomedata.org/details/study/nmdc:sty-11-28tm5d36

pkalita-lbl:
Fixes #1895 

These changes make the maximum height of the study image on the study details page dynamic. There are two cases:

1. On small screens, the "main info column" (which contains the study name, description, and data summary) and the "image column" are stacked (both `width: 100%`). The maximum height of the image is capped at 300px by setting the `max-height` prop on the `<v-img>` component.
2. On medium and larger screens, the main info column and image column are displayed side-by-side. In this case we want the height of both to _only_ be dictated by the content of the main info column. This is accomplished by making the image column absolutely positioned and therefore doesn't contribute to the height of the row containing both columns. In this case the `max-height` prop of the `<v-img>` component is _not_ set so that it can fill the maximum allowed space.

The [`useDisplay` composable](https://vuetifyjs.com/en/features/display-and-platform/#usage) from Vuetify is used to make the small vs medium or larger screen size determination. Previously we used Vuetify's `<v-row>` and `<v-col>` components, but with the amount of customization needed here it was easier to directly implement what was needed with CSS classes instead of working around the default behavior of those components.

### DIFF
--- a/web/src/views/IndividualResults/StudyPage.vue
+++ b/web/src/views/IndividualResults/StudyPage.vue
@@ -203,10 +203,7 @@ export default defineComponent({
       </div>
 
       <div :class="['mainInfoRow', smAndDown ? 'smAndDown' : 'mdAndUp']">
-        <div
-          ref="main-info-col"
-          :class="['mainInfoCol', study.image_url ? 'withImage' : '']"
-        >
+        <div :class="['mainInfoCol', study.image_url ? 'withImage' : '']">
           <IndividualTitle :item="study">
             <template
               v-if="study.description"


eecavanna (approved):
Disclaimer: I only skimmed it (on my iPhone). I'm OK with it being merged into `main`. Thanks for implementing this.

codytodonnell (approved):
Cool solution using `position: absolute` for the image column on larger screens. Also good to know about the `useDisplay` hook. This all makes sense to me and seems to be working on my end.

PR #1908, opened 2025-12-26
Track Github issue for submission via slot on submission_metadata instead of API search

samobermiller:
- update github issue search in api.py to use slot on submission_metadata
- update models.py to include github issue slot on SubmissionMetadata
   - code updates tested via https://github.com/microbiomedata/issues/issues/1524
- add migrator updating github issue number for existing submisssions

*Notes on the migrator:
- It uses a static CSV (`nmdc_server/migrations/versions/submission_github_issues_20260121.csv`) that I generated using a local python script on 1/21/26. That script searched the Github API for issues that matched each submission id. Some of those matches had to be manually verified because there was more than one (so I selected one to update going forward) or because none existed (and I confirmed with Bea that those were okay as is). 
- The script found Github issues for 119 out of 703 submissions (using a postgres dump from 1/20/2026). This makes sense because only 113 of the 703 submissions have a status other than "InProgress". Those 6 submissions that have a GH issue but are "InProgress" were likely restarted. 
- To ensure there are no stragglers at the time of migration, I added a check to the migrator finding any submissions that are 1) missing a GH issue, 2) aren't "InProgress", 3) were made after 2023 and 4) weren't already verified by me. If any of those are found (likely because someone submitted sometime between me making the static CSV and this PR going into prod) a print statement will be made. A note to check the output of this migrator will be made on approval of this PR to the [release issue](https://github.com/microbiomedata/issues/issues/1510)


pkalita-lbl (approved):
Looks good!

PR #1938, opened 2026-01-28
Enable filtering environmental filters by ID

### LINKED ISSUE #1026: data portal - environmental context faceted search should allow to search by term id
Similar to what happens with the KEGG search you should be able to search by text or by a term. I should be able to do a faceted search for broad scale environment with 'terrestrial biome' OR 'ENVO:00000446' which is the corresponding term ID. In mongo the term identifiers are stored in env_local_scale.term.id, env_broad_scale.term.id, env_medium.term.id.

Similar to what happens with KEGG search, the terms should be hyperlinked to their ontology page.

codytodonnell:
Resolves #1026 

Replace the default search logic in the `treeselect` component in the `FilterTree` wrapper so that the items in the tree can be searched by label or by ID. This applies to Broad-scale Environmental Context, Local Environmental Context, and Environmental Medium at the moment, but will apply to any facet that uses the `FilterTree` component going forward.

### DIFF
--- a/web/src/components/FilterTree.vue
+++ b/web/src/components/FilterTree.vue
@@ -185,6 +185,7 @@ export default defineComponent({
       always-open
       async
       :load-options="loadOptions"
+      placeholder="Select or search by ID or label"
       class="ma-2"
     >
       <template #option-label="{ node }">


eecavanna:
I spun up my local `nmdc-server` environment and tried playing with the search field, and got an error in the `backend` container log (and no search results on the UI). I don't have any prior experience using this search modal, and may local database may be in a wonky state as a result of working on some other things recently. I will refresh my local database (from the latest production dump) and try again. This might be user error on my part.

image

eecavanna:
One possible cause of my local woes:

```console
$ node --version
v20.19.2
```

I'm using Node version 20, but the docs say to use version 22 or newer:

https://github.com/microbiomedata/nmdc-server/blob/10f4a174800a5d3eee269dc917cce911a39c4e1c/docs/development.md?plain=1#L230

Although another part of the docs says 20 or newer:

https://github.com/microbiomedata/nmdc-server/blob/10f4a174800a5d3eee269dc917cce911a39c4e1c/docs/development.md?plain=1#L9

Either way, I will upgrade to Node 22 and re-test.

eecavanna:
Upgrading to Node 22 did not fix the issue. In other words, the strange checkbox issue I described above still occurs. I also, subsequently, upgraded to Node 24 and re-tested there. That did not fix the issue either. Neither does running `corepack enable` and running `yarn` directly (instead of via `npx`). Anyway, I'll stop posting additional things that _don't_ fix the issue, here in this merged PR.

PR #1946, opened 2026-01-31
Update Alembic migrations to be compatible with SQLAlchemy 2

### LINKED ISSUE #1808: Past Alembic migration is incompatible with SQLAlchemy 2
I'm creating this ticket while trying to get the portal backend to spin up on GKE.

When I spun up the backend, its console showed the following error, involving this migration script:

https://github.com/microbiomedata/nmdc-server/blob/6b3059b1acab2b71eb591322e92e5922fc387a26/nmdc_server/migrations/versions/0ff690fb929d_add_study_name_and_template_columns_to_.py#L40

```py
# ...
  File "/usr/local/lib/python3.12/site-packages/alembic/runtime/migration.py", line 628, in run_migrations
    step.migration_fn(**kw)
  File "/app/nmdc_server/migrations/versions/0ff690fb929d_add_study_name_and_template_columns_to_.py", line 40, in upgrade
    sa.select([submission_metadata.c.id, submission_metadata.c.metadata_submission])
  File "/usr/local/lib/python3.12/site-packages/sqlalchemy/sql/_selectable_constructors.py", line 537, in select
    return Select(*entities)
           ^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/sqlalchemy/sql/selectable.py", line 5391, in __init__
    coercions.expect(
  File "/usr/local/lib/python3.12/site-packages/sqlalchemy/sql/coercions.py", line 395, in expect
    resolved = impl._literal_coercion(
               ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/sqlalchemy/sql/coercions.py", line 634, in _literal_coercion
    self._raise_for_expected(element, argname)
  File "/usr/local/lib/python3.12/site-packages/sqlalchemy/sql/coercions.py", line 1132, in _raise_for_expected
    return super()._raise_for_expected(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/sqlalchemy/sql/coercions.py", line 695, in _raise_for_expected
    super()._raise_for_expected(
  File "/usr/local/lib/python3.12/site-packages/sqlalchemy/sql/coercions.py", line 518, in _raise_for_expected
    raise exc.ArgumentError(msg, code=code) from err
sqlalchemy.exc.ArgumentError: Column expression, FROM clause, or other columns clause element expected, got [<sqlalchemy.sql.elements.ColumnClause at 0x7cb8242980e0; id>, <sqlalchemy.sql.elements.ColumnClause at 0x7cb8242997f0; metadata_submission>]. Did you mean to say select(<sqlalchemy.sql.elements.ColumnClause at 0x7cb8242980e0; id>, <sqlalchemy.sql.elements.ColumnClause at 0x7cb8242997f0; metadata_submission>)?
```

Looks to me like the `select` function expects the columns to be passed in as positional arguments, not as a list.

I checked the SQLAlchemy 2 docs and saw the following:

Docs: https://docs.sqlalchemy.org/en/20/core/selectable.html#sqlalchemy.sql.expression.select

Image

It's not clear to me whether the flexibility described in the "Added in version 1.4" note is still present in SQLAlchemy 2.

Anyway, I am noting this for reference. I don't know whether we will go back and update migrations. I also am only seeing this in a newly-deployed environment on GKE, and do not trust the rest of the environment yet to the point where I am confident the error isn't stemming from something else.

eecavanna:
### Background

We migrated from SQLAlchemy 1 to 2 a few months ago. Four Alembic migrations were not migrated accordingly, and so they would not run anymore. They would fail with this kind of error:

```py
sqlalchemy.exc.ArgumentError: Column expression, FROM clause, or other columns clause element expected, got [<sqlalchemy.sql.elements.ColumnClause at 0xffff97a24890; id>, <sqlalchemy.sql.elements.ColumnClause at 0xffff97a03200; metadata_submission>]. Did you mean to say select(<sqlalchemy.sql.elements.ColumnClause at 0xffff97a24890; id>, <sqlalchemy.sql.elements.ColumnClause at 0xffff97a03200; metadata_submission>)?
```

This didn't affect our Spin or Cloud-hosted environments, since the Postgres databases there had already been migrated. However, it did affect local development. Developers would work around this by restoring already-migrated dumps from production.

### Changes

On this branch, I updated those Alembic migrations to be compatible with SQLAlchemy 2.

Updating them involved converting function arguments to positional ones instead of list items. 

However, once I did that conversion, `$ tox -e typecheck` (which is a build-time thing, not a run-time thing) would fail with this kind of error:

```py
nmdc_server/migrations/versions/ff4e651c3007_migrate_templates.py:33: error: Argument 1 to "Select" has incompatible type "Column[Any]"; expected "Iterable[ColumnElement[Any] | FromClause | int] | None"  [arg-type]
```

I interpreted that to mean that our type checker was still checking things based on the rules of SQLAlchemy 1, not 2. 

Indeed, when I looked into the configuration of our type checker, I found that it is using some deprecated things. I added a `TODO` comment to `mypy.ini` about this. I initially tried _removing_ the deprecated thing already; but, when I did that, `$ tox -e typecheck` reported lots more errors (involving lots of lines of code), so I think we are (i.e. the codebase is) still relying on the deprecated thing for some things.

Meanwhile, to suppress the above errors (specific to the changed lines of code) so the GHA checks would pass, I added a `# type: ignore[arg-type]` comment to each of the converted function calls.

Fixes #1808

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,4 +1,11 @@
 [mypy]
+# TODO: Remove the deprecated `sqlmypy` plugin, now that we're using SQLAlchemy 2.0.
+#
+#       From the SQLAlchemy 1.4 documentation (updated in 2024):
+#       > - "Deprecated since version 2.0: The SQLAlchemy Mypy Plugin is DEPRECATED [...]"
+#       > - "SQLAlchemy 2.0 [...] removes the need for the Mypy plugin [...]"
+#       Docs: https://docs.sqlalchemy.org/en/14/orm/extensions/mypy.html
+#
 plugins = pydantic.mypy, sqlmypy
 allow_redefinition = false
 check_untyped_defs = true
--- a/nmdc_server/migrations/versions/0ff690fb929d_add_study_name_and_template_columns_to_.py
+++ b/nmdc_server/migrations/versions/0ff690fb929d_add_study_name_and_template_columns_to_.py
@@ -37,7 +37,7 @@ def upgrade():
 
     connection = op.get_bind()
     submissions = connection.execute(
-        sa.select(submission_metadata.c.id, submission_metadata.c.metadata_submission)
+        sa.select(submission_metadata.c.id, submission_metadata.c.metadata_submission)  # type: ignore[arg-type]
     )
 
     for submission in submissions:
--- a/nmdc_server/migrations/versions/5403c7fe0b33_migrate_study_names.py
+++ b/nmdc_server/migrations/versions/5403c7fe0b33_migrate_study_names.py
@@ -31,7 +31,7 @@ def upgrade():
 
     connection = op.get_bind()
     submissions = connection.execute(
-        sa.select(submission_metadata.c.id, submission_metadata.c.metadata_submission)
+        sa.select(submission_metadata.c.id, submission_metadata.c.metadata_submission)  # type: ignore[arg-type]
     )
 
     for submission in submissions:
--- a/nmdc_server/migrations/versions/afa1ff687968_convert_packagename_to_an_array.py
+++ b/nmdc_server/migrations/versions/afa1ff687968_convert_packagename_to_an_array.py
@@ -32,7 +32,7 @@ def upgrade():
 
     connection = op.get_bind()
     submissions = connection.execute(
-        sa.select(submission_metadata.c.id, submission_metadata.c.metadata_submission)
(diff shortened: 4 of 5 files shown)


eecavanna:
The GHA checks are failing because of a type issue involving these changed lines.

PR #1957, opened 2026-02-04
Fix synchronization of DataHarmonizer validation state with backend

### LINKED ISSUE #1904: Data Harmonizer validation state does not persist in submission navigation sidebar
If I fill out some sample metadata in DataHarmonizer and click the validate button, the validation icon in the submission navigation sidebar briefly changes to a checkmark and then goes back to an x. 

In addition to fixing this issue, this task is to:

* Consider whether the submission validation state should be all string arrays, all booleans, or the current mix of both
* Consider whether there are ways to reduce the manual `revalidate` calls that are required to push updates to the validation state.

pkalita-lbl:
Fixes #1904 

Sorry for the large diff here. I'm happy to talk through any of the changes in more detail, if needed. Here's a rough outline of what's going on:

* Modeling changes
  * Renamed `validForms` to `validationState` to reflect that not everything it stores is form-based. It has been moved from `MetadataSubmissionRecordCreate` to `MetadataSubmissionRecord` so that it is not required when creaeting a new submission. On `MetadataSubmissionRecord` it has a `default_factory` so that existing submissions will get a default value when loaded for the first time.
  * The form-related fields (everything other than `sampleMetadata`) is either `None` (if the form hasn't been touched) or a list of error message strings. This means a fully valid form is represented by an empty list.
  * `sampleMetadata` is either `None` (if the interface hasn't been touched) or an object with `invalidCells` and `tabsValidated` fields. These were previously two pieces of validation state tracked only by the frontend.
* New `SubmissionForm` component
  * The three form-based pages now use a common `SubmissionForm` component. It provides an event hook for uniformly tracking validation state changes in the `null | string[]` format.
  * The new component also handles disabling itself based on user permissions.
  * The `StepperView` component is now responsible for displaying the permission-related banner messages instead of each individual form page.
* Navigation sidebar and summary page
  * I tried to slim down the design of both the sidebar and summary page in an attempt to make them fit in a little better with the rest of the site.
  * I removed the validation messages from the summary page since they are somewhat redundant with the sidebar and a lot of the messages are hard to understand out of context. I still have mixed feeling about this, so I'm open to taking another pass at it later.
  * Both the sidebar and summary page now use a common shared state for the list of pages and their validation states.
* DataHarmonizer
  * The diff of `HarmonizerView.vue` looks scary but a lot of it is because of changing some indentation levels in the template.
  * The main changes involve replacing the local `tabsValidated` and `invalidCells` state with the new shared state from the central store.

### DIFF
--- a/web/src/views/SubmissionPortal/Components/TemplateChooser.vue
+++ b/web/src/views/SubmissionPortal/Components/TemplateChooser.vue
@@ -56,7 +56,7 @@ export default defineComponent({
       @valid-state-changed="(state) => validationState.sampleEnvironmentForm = state"
     >
       <v-input
-        v-model="packageName"
+        :model-value="packageName"
         validate-on="input eager"
         :rules="[(v) => (!!v && v.length > 0) || 'Please select at least one template.']"
       >


JamesTessmer (approved):
I think all the changes look good and make sense.
Tested locally and didn't find any issues, and I think the changes to the sidebar and summary page are good.

PR #1960, opened 2026-02-05
Fix `FilterTree` component for environmental filters

### LINKED ISSUE #1953: Checkbox selection is delayed and incorrect in FilterTree when searching
There are two related bugs in the `FilterTree` component:

1. When selecting a child checkbox after searching, the parent is considered checked, even if it has other children that are unchecked (and hidden by the search).
Image

2. At times, the checkbox selection is delayed if you are selecting after having searched the tree

codytodonnell:
Resolves #1953 

In solving issue #1026, I introduced a bug that caused odd behavior in the `FilterTree` component. This was mostly fixed by switching to use the built-in `match-keys` prop in the [vue3-treeselect](https://zanmato.github.io/vue3-treeselect/) component. In implementing that feature, I discovered a bug in the underlying treeselect library which has just been fixed in the latest version `0.4.3` of the library. This PR upgrades to the latest version and implements the `match-keys` prop. Users can now search through tree options by label or ID without any unexpected behavior.

Reference: https://github.com/zanmato/vue3-treeselect/issues/11

P.S. I was having issues with `.DS_Store` files not being ignored so I added this to the top level gitignore which fixed the issue for me.

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -19,7 +19,7 @@
     "@types/lodash": "^4.17.20",
     "@vue-leaflet/vue-leaflet": "^0.10.1",
     "@vueuse/core": "^13.9.0",
-    "@zanmato/vue3-treeselect": "github:zanmato/vue3-treeselect",
+    "@zanmato/vue3-treeselect": "^0.4.3",
     "axios": "^0.28.0",
     "axios-cache-adapter": "^2.7.3",
     "bootstrap": "4.3.1",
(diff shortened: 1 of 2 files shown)


eecavanna (approved):
I didn't spin this up locally. I reviewed the diff and all changes looked intentional to me and consistent with the PR description. Props on reporting the bug in the upstream repo and keeping track of the status of it being fixed there!

Issue #1993, opened 2026-02-24
Messages logged in `do_ingest` function do not appear on console


eecavanna:
I just noticed that the ingest logs (whether I'm viewing them via Grafana or directly via GKE) do not include any of the log entries generated inside the [do_ingest](https://github.com/microbiomedata/nmdc-server/blob/fa40c3289e715ce09259d247a5637f47671b00cb/nmdc_server/jobs.py#L57) function, although the logs do include the log entries generated inside the [load](https://github.com/microbiomedata/nmdc-server/blob/fa40c3289e715ce09259d247a5637f47671b00cb/nmdc_server/ingest/all.py#L42) function _called from_ the `do_ingest` function. I suspect the `logger` variable used in the `do_ingest` function is inadequate.

Here's what the log of the latest dev ingest contains (notice the module paths within the square brackets):

```py
▶️ Ingest is starting.
• Environment: `development`
• Start time: `2026-02-24T10:15:46+00:00`
• Ingest database: `nmdc_b`
Sending Slack message having text: ▶️ Ingest is starting.
• Environment: `development`
• Start time: `2026-02-24T10:15:46+00:00`
• Ingest database: `nmdc_b`
Starting new HTTPS connection (1): hooks.slack.com:443
https://hooks.slack.com:443 "POST /services/__REDACTED__/__REDACTED__/__REDACTED__ HTTP/1.1" 200 2
Sent Slack message.
Updating NMDC functions for the active database.
Updating NMDC functions for the ingest database.
INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
INFO  [alembic.runtime.migration] Will assume transactional DDL.
INFO  [nmdc_server.ingest.all] Loading ENVO terms: Starting.
INFO  [nmdc_server.ingest.all] Loading ENVO terms: Finished in 445 seconds.
INFO  [nmdc_server.ingest.all] Loading KEGG orthology: Starting.
INFO  [nmdc_server.ingest.all] Loading KEGG orthology: Finished in 28 seconds.
INFO  [nmdc_server.ingest.all] Loading studies: Starting.
INFO  [nmdc_server.ingest.all] Studies: extracted 84, loaded 84.
INFO  [nmdc_server.ingest.all] Loading studies: Finished in 67 seconds.
INFO  [nmdc_server.ingest.all] Loading data objects: Starting.
INFO  [nmdc_server.ingest.all] DataObjects: extracted 248418, loaded 248418.
[...]

PR #1994, opened 2026-02-24
Invoke `get_logger` later in an attempt to fix production ingest logging

eecavanna:
On this branch, I made a change that I think will fix the logging issue described in #1993. Both the code on `main` and the code on this branch result in the logs appearing in my local environment; but the code on `main` does not result in the logs appearing in dev/prod. Once this has been merged into `main`, I'll trigger an ingest and monitor its logs to see whether this, indeed, ~fixed the issue~ eliminated the symptom.

### DIFF
--- a/nmdc_server/jobs.py
+++ b/nmdc_server/jobs.py
@@ -20,11 +20,14 @@ from nmdc_server.logger import get_logger
 
 HERE = Path(__file__).parent
 
-logger = get_logger(__name__)
-
 
 def update_nmdc_functions():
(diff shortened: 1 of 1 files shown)


eecavanna:
Hi @codytodonnell, this PR is a little unusual, in my opinion. I don't yet know whether the changes will fix the issue, since I haven't been able to reproduce the issue locally. I consider the changes to be narrowly-scoped and easily-reversible enough to merge them into `main` for testing in dev without having the usual "high" level of confidence that they'll fix the issue.

PR #2011, opened 2026-03-06
Collect publication DOIs in Study Information page

### LINKED ISSUE #2007: Add optional publication DOI input box to Study Information page
Currently we do not asked user if they have a publication DOI, add an optional input for publication DOIs. 

Question - would this require changes to the translator as the translator right now does not support the publication category?

pkalita-lbl:
Fixes #2007 

The primary purpose of these changes is to add a Publication DOIs section to the Submission Portal's Study Information page. This section allows users to enter an arbitrary number of Publication DOIs. Note that this _only_ collects what the schema refers to as the [`doi_value`](https://microbiomedata.github.io/nmdc-schema/doi_value/) and _not_ the [`doi_provider`](https://microbiomedata.github.io/nmdc-schema/doi_provider/) because the enum values for that slot are more geared towards data DOIs.

image

Secondary changes:

* Update Python and TypeScript tying so make `Doi.provider` optional -- in agreement with the schema's [`Doi` class](https://microbiomedata.github.io/nmdc-schema/Doi/).
* Clean up DOI format validation code to provide more useful error messages.

### DIFF
--- a/nmdc_server/schemas_submission.py
+++ b/nmdc_server/schemas_submission.py
@@ -20,7 +20,7 @@ class Contributor(BaseModel):
 
 class Doi(BaseModel):
     value: str
-    provider: str
+    provider: Optional[str] = None
 
 
 class StudyFormCreate(BaseModel):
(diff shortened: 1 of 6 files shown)


JamesTessmer (approved):
Looks good

PR #2065, opened 2026-04-09
Update TSV-formatted submission report to include "Award" column

eecavanna:
On this branch, I updated the `GET /metadata_submission/report` endpoint so that the TSV file it returns includes an "Award" column whose value is the value the submitter specified on the submission portal form shown here:

image

"Award" may not be the label stakeholders expect for that column (I derived it from the field names, not the UI). The column's label can be updated via this PR before merging, or via a future PR.

Here's a screenshot of the upper right corner of a TSV file generated by this endpoint in my local environment:

image

This addition to the TSV file was requested by a teammate preparing to provide some submission metrics data to someone else.

Fixes #2064

### DIFF
--- a/nmdc_server/api.py
+++ b/nmdc_server/api.py
@@ -1372,9 +1372,9 @@ async def get_metadata_submissions_report(
         #       the user can enter a custom string, which gets stored in the `otherAward` field.
         #
         sentinel_value_for_other = "OTHER"
-        multi_omics_form = metadata["multiOmicsForm"] if "multiOmicsForm" in metadata else {}
-        predefined_award = multi_omics_form["award"] if "award" in multi_omics_form else ""
-        custom_award = multi_omics_form["otherAward"] if "otherAward" in multi_omics_form else ""
+        multi_omics_form = metadata.get("multiOmicsForm", {})
+        predefined_award = multi_omics_form.get("award", "")
+        custom_award = multi_omics_form.get("otherAward", "")
         award = ""
         if isinstance(predefined_award, str) and predefined_award != sentinel_value_for_other:
             award = predefined_award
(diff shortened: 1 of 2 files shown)


eecavanna:
The aspect of this change that I most want feedback about is: **What do you recommend the new TSV column be named?** From the GUI, I was expecting the stored value to represent a "kind of project", so maybe "`Project type`"? However, the JSON property names in the database (which were `award` and `otherAward`) led me to expect the stored value to be an award (or some kind of identifier of an award). Working alone at the time, I just went with "`Award`" for the column name (to align with the field names). I'm curious what y'all suggest (especially those of you familiar with this GUI and/or the JSON data structure). I can update it before merging, or someone can update it via a follow-on PR.

Maybe "Award" is sufficient. 🤷

nmdc_server/api.py
```diff
@@ -1363,6 +1364,23 @@ async def get_metadata_submissions_report(
...
+        #
+        # Note: On the submission portal, this value is solicited from the user by prompting
+        #       them for the "kind of project you have been awarded". When the user marks the
+        #       "Other" radio button (in which case, "OTHER" gets stored in the `award` field),
+        #       the user can enter a custom string, which gets stored in the `otherAward` field.
+        #
+        sentinel_value_for_other = "OTHER"
+        multi_omics_form = metadata["multiOmicsForm"] if "multiOmicsForm" in metadata else {}
+        predefined_award = multi_omics_form["award"] if "award" in multi_omics_form else ""
+        custom_award = multi_omics_form["otherAward"] if "otherAward" in multi_omics_form else ""
```
pkalita-lbl:
```suggestion
        multi_omics_form = metadata.get("multiOmicsForm", {})
        predefined_award = multi_omics_form.get("award", "")
        custom_award = multi_omics_form.get("otherAward", "")
```

pkalita-lbl (approved):
Looks good. Minor stylistic comment.

No opinion on the column header. I'll leave that up to others to weigh in on.

nmdc_server/api.py
eecavanna: Thanks. I like that. Committed!

PR #2080, opened 2026-04-15
Update study index and query logic

### LINKED ISSUE #2074: Bug(s) when using faceted search by ID using a parent study ID
In the data portal dev I tried a search by a parent study ID, it says no samples are returned and shows no samples in the map but it shows 15 metabolomics results in the bar chart.

to reproduce: https://data-dev.microbiomedata.org/?q=Ch4ICBAAGAMiFiJubWRjOnN0eS0xMS1tc2V4c3kyOSI

On prod this returns no results.

Expected results: The same as selecting the parent study via [clicking the checkbox for the parent study](https://data.microbiomedata.org/?q=Ch4IABABGAMiFiJubWRjOnN0eS0xMS1tc2V4c3kyOSIKHggAEAEYAyIWIm5tZGM6c3R5LTExLTA3NmM5OTgwIg).

Note: I believe this is caused by a difference between an active search query of `ID [like] nmdc:sty-11-msexsy29` (undesirable behhavior) versus ```
Study ID [is] 
nmdc:sty-11-msexsy29
nmdc:sty-11-076c9980

```
(desirable behavior).

### LINKED ISSUE #2075: full text search with a parent study ID returns not result
env: data portal dev
to reproduce: https://data-dev.microbiomedata.org/?q=Ch4ICBAaGAoiFiJubWRjOnN0eS0xMS1tc2V4c3kyOSI

expected behavior: I expect this to apply a search filter for the microbes persist science focus area (study ID nmdc:sty-11-msexsy29) and children of that study.

### LINKED ISSUE #2076: bug with parent studies with full text search
env: data portal dev
to reproduce: https://data-dev.microbiomedata.org/?q=ChEICBAaGAoiCSJwZXJzaXN0Ig

expected behavior: A full text search on `persist` should return the parent study nmdc:sty-11-msexsy29 (& any child studies) based on a title match of 'Microbes Persist Science Focus Area'

Observed behavior: The child studies matches but the parent study is not present.

### LINKED ISSUE #2077: full text search should include alternative identifier slots
env: data portal dev
to reproduce: https://data-dev.microbiomedata.org/?q=ChMICBAaGAoiCyJHYjAxNTcxOTci

expectation: I'd expect a full text search for Gb0157197 to return NMDC biosample ID nmdc:bsm-11-cpekyy11
observation: zero samples are returned

Ideally external identifier slots are treated in a similar way as NMDC identifiers, I'm not sure this is possible since a user likely wouldn't know to provide the curie prefix. e.g. a user isn't going to know to search for `gold:Gb0157197` instead of `Gb0157197`

[Alternative_identifiers](https://microbiomedata.github.io/nmdc-schema/alternative_identifiers/) and slots which inherit from alternative_identifiers should at be included as fields in full text search.
[source_mat_id](https://microbiomedata.github.io/nmdc-schema/source_mat_id/) should also be included.

### PR DESCRIPTION
**Pull request by codytodonnell**
Resolves #2074; resolves #2075; resolves #2076; resolves #2077

Fixes a number of issues found when searching for studies and their biosamples using FTS.

Full text searches will now include studies that match the search on its own (doesn't need any matched biosamples). This is done by adding an extra step to the `StudyQuerySchema` `query()` function.

Studies (and their associated biosamples) can also now be matched by their parent or children IDs. This is achieved by including the `children` and `part_of` fields in the Study model's `__ts_vector__`. This includes searches that use an explicit study ID (this required a separate fix in the `if typecode == NmdcTypecode.study` section of the biosample FTS function.

Biosamples can now be searched by values in their `alternate_identifiers` field.

The FTS helper SQL functions creation script is no longer specified in `models.py` and no longer needs to be included in any migration script (however it is still present in `b7d4b19db410_add_fts_gin_indexes.py` because I didn't want to change an existing migration script). Now, all SQL functions are defined in `database.py` inside the `update_nmdc_functions_sql` variable. This script gets executed every time we run `nmdc-server migrate`, which means those functions will update any time the backend container is restarted.

Note: The JSONB columns referenced in the SQL functions needed to be coalesced to empty vectors to handle situations where the field was null.

### CODE REVIEW

File: nmdc_server/migrations/versions/34a0480316fa_update_study_fts_index.py [hunk 1]
```diff
@@ -0,0 +1,48 @@
...
+
+
+def upgrade():
+    op.drop_index("ix_study_fts", table_name="study")
+    op.create_index(
+        "ix_study_fts",
+        "study",
+        [
+            sa.text(
+                "nmdc_study_fts(id, name, description, gold_name, gold_description, scientific_objective, annotations, part_of, children)"
```
eecavanna, line 28: I noticed the type-casting is not happening in this new version. I assume that was an intentional change.

eecavanna (approved):
Defining the SQL functions in this way (i.e. "piggybacking" on the existing setup) makes sense to me.

pkalita-lbl (approved):

nmdc_server/migrations/versions/34a0480316fa_update_study_fts_index.py
codytodonnell: Yes that should be the appropriate way to define this index now. I don't think the casting was ever strictly necessary. Note that these statements were generated by alembic's `autogenerate` command.

PR #2092, opened 2026-04-17
Use uv for Python project management

### LINKED ISSUE #1775: Improve python project dependency management
Currently the dependency list in [pyproject.toml](https://github.com/microbiomedata/nmdc-server/blob/main/pyproject.toml) is unorganized, and includes indirect dependencies instead of just direct dependencies.

This was the result of attempting to update multiple dependencies at once, which proved difficult until running the command `uv pip compile pyproject.toml` was able to generate a list of all dependencies at a tolerable speed. See commit 7fb779c

Here is what I believe the list of direct dependencies to be. I looked at `pyproject.toml` just before the commit, and all the occasions that it has been edited since:
```
alembic
authlib
click
cryptography
dnspython
fastapi
fastapi-debug-toolbar
factory-boy
google-cloud-storage
httpx
ipython
itsdangerous
mypy
nmdc-schema
nmdc-submission-schema
nmdc-geoloc-tools
pint
psycopg2
pydantic
pymongo
python-dateutil
python-dotenv
requests
sentry-sdk[celery,sqlalchemy]
sqlalchemy
sqlparse
starlette
typing-extensions
importlib-metadata
```

At @eecavanna's suggestion, I used a tool called [deptry](https://deptry.com/) to analyze this new set of dependencies. Here are the exact steps I followed to see what this tool had to say about the code base:

1. I copy/pasted the above into `pyproject.toml` (replacing the list of all pinned dependencies)
2. I ran `uvx deptry .` in the `nmdc-server` directory (root of this repo)

Here is the report that was generated:
```
nmdc_server/api.py:15:1: DEP001 'linkml_runtime' imported but missing from the dependency definitions                                                                                 
nmdc_server/app.py:5:1: DEP001 'debug_toolbar' imported but missing from the dependency definitions                                                                                   
nmdc_server/binning.py:6:1: DEP001 'dateutil' imported but missing from the dependency definitions                                                                                    
nmdc_server/cli.py:215:5: DEP001 'IPython' imported but missing from the dependency definitions                                                                                       
nmdc_server/cli.py:216:5: DEP001 'traitlets' imported but missing from the dependency definitions                                                                                     
nmdc_server/config.py:6:1: DEP001 'pydantic_settings' imported but missing from the dependency definitions                                                                            
nmdc_server/database.py:7:1: DEP001 'debug_toolbar' imported but missing from the dependency definitions                                                                              
nmdc_server/database.py:8:1: DEP001 'pygments' imported but missing from the dependency definitions                                                                                   
nmdc_server/database.py:9:1: DEP001 'pygments' imported but missing from the dependency definitions                                                                                   
nmdc_server/database.py:10:1: DEP001 'pygments' imported but missing from the dependency definitions
nmdc_server/fakes.py:6:1: DEP001 'factory' imported but missing from the dependency definitions
nmdc_server/fakes.py:7:1: DEP001 'factory' imported but missing from the dependency definitions
nmdc_server/fakes.py:8:1: DEP001 'faker' imported but missing from the dependency definitions
nmdc_server/ingest/doi.py:8:1: DEP001 'urllib3' imported but missing from the dependency definitions
nmdc_server/static_files.py:5:1: DEP001 'linkml_runtime' imported but missing from the dependency definitions
nmdc_server/static_files.py:6:1: DEP001 'linkml_runtime' imported but missing from the dependency definitions
nmdc_server/storage.py:8:1: DEP001 'google' imported but missing from the dependency definitions
nmdc_server/storage.py:9:1: DEP001 'google' imported but missing from the dependency definitions
nmdc_server/storage.py:10:1: DEP001 'google' imported but missing from the dependency definitions
pyproject.toml: DEP002 'cryptography' defined as a dependency but not used in the codebase
pyproject.toml: DEP002 'dnspython' defined as a dependency but not used in the codebase
pyproject.toml: DEP002 'fastapi-debug-toolbar' defined as a dependency but not used in the codebase
pyproject.toml: DEP002 'factory-boy' defined as a dependency but not used in the codebase
pyproject.toml: DEP002 'google-cloud-storage' defined as a dependency but not used in the codebase
pyproject.toml: DEP002 'ipython' defined as a dependency but not used in the codebase
pyproject.toml: DEP002 'itsdangerous' defined as a dependency but not used in the codebase
pyproject.toml: DEP002 'mypy' defined as a dependency but not used in the codebase
pyproject.toml: DEP002 'nmdc-submission-schema' defined as a dependency but not used in the codebase
pyproject.toml: DEP002 'psycopg2' defined as a dependency but not used in the codebase
pyproject.toml: DEP002 'python-dateutil' defined as a dependency but not used in the codebase
pyproject.toml: DEP002 'python-dotenv' defined as a dependency but not used in the codebase 
pyproject.toml: DEP002 'importlib-metadata' defined as a dependency but not used in the codebase
Found 32 dependency issues.
```

At first this looks like a lot, but it appears that at least some of these issues are due to mismatched names between the dependency as specified in `pyproject.toml` and as imported by code. For example, `deptry` reports that `debug_toolbar` is imported but missing from the dependency definitions, but the dependency _is_ specified as `fastapi-debug-toolbar`. I see several other instances of this pattern:
```
google <-> google-cloud-storage
dateutil <-> python-dateutil
factory <-> factory-boy
IPython <-> ipython
```

Other issues are valid. For example, it looks like we import `pygments` directly in the `nmdc-server` code, but don't specify this as a dependency. This happens to be included indirectly from `ipython`.

Additionally, some of these dependencies could likely be moved to a `dev` group (mostly thinking about `fastapi-debug-toolbar`).

FYI @codytodonnell @eecavanna I'm writing this issue to describe the current state of  backend (python) dependencies for this repo. I hope this can be used for reference when migrating to uv/poetry to slim down the `pyproject.toml` file.

pkalita-lbl:
Fixes #1775 

### Summary

These changes replace the ad-hoc `pyproject.toml` with _all_ (even transitive) dependnecies and `pip install` setup with [uv](https://docs.astral.sh/uv/) for more maintainable Python project management.

In uv's terms, `nmdc-server` is an [application](https://docs.astral.sh/uv/concepts/projects/init/#applications) (as opposed to a [library](https://docs.astral.sh/uv/concepts/projects/init/#libraries) or even a [packaged application](https://docs.astral.sh/uv/concepts/projects/init/#packaged-applications)). In other words, it only needs to run under one version of Python with a specific set of dependencies (with known versions). That distinction drives a lot of the decisions here.

### Details

* Add a `.python-version` specifying `3.12`. This is the one Python version we really need to care about, and this is the Python version we have been using in the backend Docker image. Having this file keeps non-Docker environments in sync because uv will use this when setting up new environments by default. These changes also remove Python version matrix from the `server.yml` GitHub Actions workflow file. Instead, it also now defers to the `.python-version` file.
* List only direct dependencies in `pyproject.toml`. These changes remove all the explicitly listed transitive dependencies and a number of totally unused dependencies. Most of these dependencies do not have any version specifier. As an application, the `uv.lock` file is the source of truth about which exact versions will be used. I made an exception for a few dependencies (like `nmdc-schema`) where we _really_ care about the version. I figured people might feel more comfortable seeing a version number explicitly listed in `pyproject.toml`. **I recognize this might be a controversial choice** and ultimately I don't have extremely strong feelings on it. If we want to list versions in `pyproject.toml` (explicit, but slightly redundant with `uv.lock`) I'm open to it.
* Use `tox-uv` and `uv-venv-lock-runner` in `tox.ini`. With uv and support for `dependency-groups`, it really feels like we don't get a ton of value out of tox, but I'm leaving it in for now.
* Add [deptry](https://deptry.com/) as a dependency checker. It helped identify a number of unused dependencies that were previously listed in `pyproject.toml`. I figured why not leave it in and add it to our automated checks. One thing it flagged was that we had one module (`fakes.py`) that was really only for testing purposes. I moved it into the `tests` directory and made that directory a package so that the `fakes.py` module could be imported by tests. Deptry also flagged that we were still using `typing_extensions` which was only needed in older versions of Python.

### Out of scope

Some things I thought about doing but decided they would be better as follow-on tasks:

* As noted above, we could consider removing tox. A few `Makefile` (or [`justfile`](https://just.systems/man/en/)) targets would probably suffice.
* I did a very basic migration of the backend `Dockerfile`. I'm sure that it could be looked at closer and optimized a bit.
* We are still using `setuptools-scm` for determining a dynamic version number. It seems to work fine, but it is maybe a little wonky. We could consider switching to [`uv-dynamic-versioning`](https://github.com/ninoseki/uv-dynamic-versioning).
* A lot of our code quality tools are pretty out-of-date. We could replace black, flake8, and isort with [ruff](https://docs.astral.sh/ruff/).


Dockerfile
```diff
@@ -9,11 +9,23 @@ RUN apt-get update
...
+# Install uv
+COPY --from=ghcr.io/astral-sh/uv:0.11.7 /uv /uvx /bin/
+
+# Install the project's dependencies from the lockfile and caching
+# the environment in a dedicated layer.
+WORKDIR /app
+RUN --mount=type=cache,target=/root/.cache/uv \
+    --mount=type=bind,source=uv.lock,target=uv.lock \
+    --mount=type=bind,source=pyproject.toml,target=pyproject.toml \
+    uv sync --frozen --no-install-project --no-dev
```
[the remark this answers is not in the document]
pkalita-lbl: I'll lump this in with my second "Out of scope" bullet point.

.github/workflows/server.yml
```diff
@@ -56,11 +51,20 @@ jobs:
           -p 4443:4443 \
           fsouza/fake-gcs-server \
           -scheme http
-    - uses: actions/checkout@v4
-    - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v5
+
+    - uses: actions/checkout@v6
```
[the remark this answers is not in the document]
pkalita-lbl: Updated other workflows to use `v6`.

pkalita-lbl:
Note: this repo is configured to require a check (`server (3.12)`) that no longer exists because of the changes to `server.yml`. If/when we're ready to move forward with these changes I'll make the necessary repo settings change.

Dockerfile
```diff
@@ -9,11 +9,23 @@ RUN apt-get update
 #   - docker-compose.yml
 RUN apt-get install -y postgresql-client-15 git libpq-dev libc6-dev gcc
 
-RUN pip install -U pip setuptools wheel
+# Install uv
+COPY --from=ghcr.io/astral-sh/uv:0.11.7 /uv /uvx /bin/
+
+# Install the project's dependencies from the lockfile and caching
+# the environment in a dedicated layer.
```
eecavanna:
```suggestion
# Install the project's dependencies from the lockfile and cache
# the environment in a dedicated layer.
```

tox.ini
```diff
@@ -69,9 +55,21 @@ commands = black --check nmdc_server tests --diff --color
...
-deps = black
+runner = uv-venv-lock-runner
+dependency_groups =
+    lint
 commands = black nmdc_server tests
 
+[testenv:dependencies]
+runner = uv-venv-lock-runner
+dependency_groups =
+    # Install all dependency groups so that deptry gets the full mapping of packages to module names
```
eecavanna, line 66: I am assuming the "main" (a.k.a. production) dependencies will be installed also, so that deptry processes those also.

Dockerfile
eecavanna: I recommend including [--compile-bytecode](https://docs.astral.sh/uv/reference/cli/#uv-sync--compile-bytecode), so that the `.pyc` files get generated at image _build_ time instead of _when the dependency package is first imported_ (the latter can increase how long it takes for the containerized app to become ready to receive HTTP requests).

```suggestion
    uv sync --frozen --no-install-project --no-dev --compile-bytecode
```

Dockerfile
```diff
@@ -9,11 +9,23 @@ RUN apt-get update
...
+RUN --mount=type=cache,target=/root/.cache/uv \
+    --mount=type=bind,source=uv.lock,target=uv.lock \
+    --mount=type=bind,source=pyproject.toml,target=pyproject.toml \
+    uv sync --frozen --no-install-project --no-dev
+
+# Install the project itself.
 COPY . /app/
-RUN pip install -e /app
+RUN --mount=type=cache,target=/root/.cache/uv \
+    uv sync --frozen --no-dev
```
eecavanna: I recommend including [--compile-bytecode](https://docs.astral.sh/uv/reference/cli/#uv-sync--compile-bytecode), so that the `.pyc` files get generated at image _build_ time instead of _when the dependency package is first imported_ (the latter can increase how long it takes for the containerized app to become ready to receive HTTP requests).


```suggestion
    uv sync --frozen --no-dev --compile-bytecode
```

> Note: This comment was mostly copy/pasted from my comment on line R[ight]21.

eecavanna (approved):
Thank you for implementing this!

The diff looks good to me. There are no deal breakers, from my perspective. I left some comments that I'd consider to be "optimization opportunities" and "things that caught my eye."

One last comment is that, on `nmdc-runtime`, FWIW, I opted to use [--locked](https://docs.astral.sh/uv/reference/cli/#uv-sync--locked) instead of [--frozen](https://docs.astral.sh/uv/reference/cli/#uv-sync--frozen) when running `$ uv sync ...`. The former includes a check of whether `pyproject.toml` and the lockfile agree, whereas the latter does not. I don't remember whether I considered, or was aware of, `--frozen` at the time.

eecavanna:
One more note:

When updating the Runtime to use `uv`, we opted to locate the virtual environment outside of the directory that gets synced with developers' host machines. Here's the explanation I put in the `nmdc-runtime`'s Dockerfile about that:

```py
#       Q: Why don't we use `./.venv` in the repository file tree?
#       A: If we were to do that, then, whenever a developer would mount (via our Docker Compose file)
#          the repository file tree from their host machine (which may include a `.venv/` directory
#          created by their host machine) into the container, it would overwrite the Python virtual
#          environment that the `uv` program inside the container is using.
```

Source: https://github.com/microbiomedata/nmdc-runtime/blob/67770b8ad7d25d6d301b148b07ee4feb2faed42f/nmdc_runtime/Dockerfile#L57-L61

A consequence of that is that we ended up having to include `--active` with all subsequent `$ uv ...` commands (so, less convenience downstream). ☹️

---

_Edit:_ That may not be necessary for `nmdc-server`, though, because the `backend` container does not mount the root directory of the repo.

https://github.com/microbiomedata/nmdc-server/blob/a0a5f27c9fcbe77239b6cf3111e995a15b1c7fde/docker-compose.yml#L50-L54

pkalita-lbl:
I think having the virtual environment located inside the project (in uv's default `.venv` directory) isn't causing an issue here because that directory is listed in `.dockerignore`: https://github.com/microbiomedata/nmdc-server/blob/a0a5f27c9fcbe77239b6cf3111e995a15b1c7fde/.dockerignore#L5

This is consistent with the suggestion in the [uv docs](https://docs.astral.sh/uv/guides/integration/docker/#installing-a-project):

> It is best practice to add .venv to a [.dockerignore file](https://docs.docker.com/build/concepts/context/#dockerignore-files) in your repository to prevent it from being included in image builds. The project virtual environment is dependent on your local platform and should be created from scratch in the image.

Even in the pre-uv setup I kept my `nmdc-server` virtual environment in the in-project `.venv` directory.

---

Edit: oh, yeah, you're right. The selective mounting might also be at play here.

Dockerfile
pkalita-lbl: Enabled via the `UV_COMPILE_BYTECODE` environment variable

Dockerfile
pkalita-lbl: Enabled via the `UV_COMPILE_BYTECODE` environment variable

tox.ini
pkalita-lbl: Evidently, yeah 😁. I tried not to expend too much energy understanding `tox-uv` because I'd really like to get rid of `tox`.

pkalita-lbl:
> One last comment is that, on nmdc-runtime, FWIW, I opted to use [--locked](https://docs.astral.sh/uv/reference/cli/#uv-sync--locked) instead of [--frozen](https://docs.astral.sh/uv/reference/cli/#uv-sync--frozen)

I don't recall what made me use `--frozen` in the first place. The uv docs section on [intermediate layers](https://docs.astral.sh/uv/guides/integration/docker/#intermediate-layers) shows an example that's almost identical to the `nmdc-server` setup, and it uses `--locked`. So that's good enough for me. I've switched to `--locked`.

tox.ini
eecavanna: OK, cool. Thanks for checking. I neglected to say that I was in favor of merging even without knowing whether that specific assumption of mine was correct. I'll resolve this thread now.

eecavanna (approved):
All changes since my last review look good to me. I'm in favor of this being merged in as long before the next release as possible, in the hopes of anything we might have overlooked emerging through normal use as soon as possible.

PR #2095, opened 2026-04-20
Add accurate timestamp to metadata json files

### LINKED ISSUE #2079: Downloaded metadata JSON files are dated incorrectly
All metadata files bulk downloaded from the Download --> Metadata UI are dated as being created on Jan 1 1980. This is a result of the `zipfile` module applying its default date.

codytodonnell:
Resolves #2079 

Metadata JSON files added to the zip folder were defaulting to Jan 1 1980 because no date was being specified. Now, metadata JSON files should be dated with the time of download.


pkalita-lbl (approved):

nmdc_server/api.py
```diff
@@ -496,7 +497,12 @@ async def generate_zip():
                 if filename is None:
                     continue
 
-                with zf.open(f"{filename}.json", "w") as jf:
+                zinfo = zipfile.ZipInfo(
+                    filename=f"{filename}.json",
+                    date_time=datetime.datetime.now().timetuple()[:6],  # default is 1980-01-01
```
eecavanna, line 502: Thinking out loud: I'm wondering whether time zones come into play here.

eecavanna (approved):

nmdc_server/api.py
codytodonnell: I thought about that. I don't think we can know the user's timezone without pulling that data from the client. So this will always use the system's timezone, which I think is okay for this situation.

PR #2100, opened 2026-04-22
Add link to text search documentation

### LINKED ISSUE #2085: Add documentation page that explains how text search works
We need a new section of our docs site that explains in more detail how the new text search feature works. This page should list all the fields that are checked by the search function and it should link to where these fields are defined in the code to prevent drift between the lists.

We should also add a link to this page inside the Text Search menu item.

codytodonnell:
Resolves #2085 

Adds a help icon with a tooltip and link to the Data Portal User Guide documentation page. It links specifically to the new section called `Text Search`. This section will be added via this PR: https://github.com/microbiomedata/docs/pull/346.


eecavanna (approved):
LGTM! Thanks for adding this. I suggest adding Python comments to the parts of the code that are upstream of the "docs" section that you wrote, in order to remind developers to update that specific "docs" section if they make a change to the code (e.g. adding/removing a column from the set of columns considered by FTS).

PR #2104, opened 2026-04-23
Activate uv virtual environment in Dockerfile

### LINKED ISSUE #2103: Ingest fails to start following switch to `uv`
When I run ingest via...

https://github.com/microbiomedata/nmdc-server/blob/af84525d926dec8e668b6054c22354cfaa0335e5/docs/development.md?plain=1#L282-L285

...I get the error shown here:

```console
$ docker-compose run backend nmdc-server ingest -vv --function-limit 100
[+]  1/1t 1/1
 ✔ Container nmdc-server-db-1 Started                                                                                    0.2s
Container nmdc-server-backend-run-707f114538c5 Creating 
Container nmdc-server-backend-run-707f114538c5 Created 
Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: exec: "nmdc-server": executable file not found in $PATH
```

I first noticed this when the nightly dev ingest did not start successfully. The logs on Argo showed that error. I then reproduced it locally (the above snippet is from my local development environment).

Due to the timing (the nightly dev ingest ran OK the previous day), I suspect it is related to Wednesday's switch to `uv`.

pkalita-lbl:
Fixes #2103 

These changes activate the uv virtual environment before executing the container startup command (whether the default via `CMD` over overridden at runtime). Activation is done by prepending the path as suggested by the [uv docs](https://docs.astral.sh/uv/guides/integration/docker/#using-the-environment). Having the virtual environment activated automatically makes containers running the backend image behave (from an outside view) equivalently to the old `pip`-based approach which modified the system Python environment. I thought that would be more convenient than having to sprinkle `uv run`s everywhere (and potentially missing some).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,7 +30,10 @@ COPY . /app/
 RUN --mount=type=cache,target=/root/.cache/uv \
     uv sync --locked
 
+# Activate the virtual environment
+ENV PATH="/app/.venv/bin:$PATH"
+
 COPY .env.production /app/.env
 RUN chmod +x /app/start.sh
 WORKDIR /app/
-CMD ["uv", "run", "/app/start.sh"]
+CMD ["/app/start.sh"]
--- a/docs/development.md
+++ b/docs/development.md
@@ -281,7 +281,7 @@ It is recommended to use `127.0.0.1` instead of `localhost` for local developmen
     ```
 5. Run the ingest command:
     ```bash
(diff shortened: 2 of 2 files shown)


eecavanna (approved):
I like this solution. Thanks!

I'm going to revisit how the Runtime is setup (`uv run --active` is currently required there, all over the place).

PR #2105, opened 2026-04-23
Add check to date histogram for empty data set

### LINKED ISSUE #2091: Date histogram does not update if there are zero results
If a search or filter returns zero results, the date histogram does not update to show a no result chart. Instead, it continues to show the previous state of the chart. This can cause confusion as it looks like there are results to display in the date chart when in fact there are none.

Image

JamesTessmer:
Resolves #2091 

Add a check to the `DateHistogram` to see if `facetSummary.bins` length is greater than 0 before displaying anything.
Tested with the same settings as shown in the issue and it appears to be resolved.


codytodonnell (approved):
Works great on my end! Thank you!

PR #2115, opened 2026-05-02
Update `nmdc-schema` package from `11.18.0` to `11.18.1`

### LINKED ISSUE #2114: Update `nmdc-schema` package from `11.18.0` to `11.18.1`
On this branch, I will leave the submission schema at its current version and only update the `nmdc-schema` package. A couple of us are discussing this on Slack (in this [DM](https://nmdc-group.slack.com/archives/C07FQKZPTS9/p1777679177186969)).

CC: @pkalita-lbl and @aclum

eecavanna:
This is my first time doing this since we switched to uv.

I manually edited `pyproject.toml` (as shown in the diff) and then ran:

```sh
uv sync --all-groups
```

That made changes to the `uv.lock` file, which I committed.

In hindsight, I think I could have just run `$ uv lock`. Docs: https://docs.astral.sh/uv/concepts/projects/sync/#creating-the-lockfile

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,7 +24,8 @@ dependencies = [
     "linkml-runtime",
     "nmdc-geoloc-tools",
     "nmdc-metadata-suggestor-ai-tool==1.1.1",
-    "nmdc-schema==11.18.0",
+    "nmdc-schema==11.18.1",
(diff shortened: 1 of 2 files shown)


eecavanna:
I forgot to also update the `nmdc-schema` version specified in `package.json` (and the corresponding lockfile). I'm doing that via PR https://github.com/microbiomedata/nmdc-server/pull/2117 now.

PR #2119, opened 2026-05-04
Store validation state inline in form objects

### LINKED ISSUE #1720: Update how form validation is tracked in submission portal
We want to move the validation state of each form from `validationState` into each form (e.g. `studyForm` tracks its own validity), which should make it easier down the line to have multiple sample sets per submission.

To dos:
Move all validation to its own form
Make a migrator to match that
update any tests/fakes

JamesTessmer:
This PR updates how tracking is done for validation (in preparation for having multiple sample sets per submission).
Now `StudyForm` and `MultiOmicsForm/AddressForm` track their own validation.
Validation for `SampleEnvironment` and `SampleData` have been moved to `MetdataSubmission`. These will probably need to be moved to the new table that will be made for `sampleSet` in the future, but this is where they would live for now.

All uses and references to `validationState` have been removed or updated as needed.
Resolves https://github.com/microbiomedata/issues/issues/1720


nmdc_server/migrations/versions/f9630762bd00_move_validation_to_each_form.py
```diff
@@ -0,0 +1,169 @@
...
+        package_name = _ensure_list(ms.get("packageName"))
+        templates = _ensure_list(ms.get("templates"))
+        study_form = _ensure_dict(ms.get("studyForm"))
+        multiomics_form = _ensure_dict(ms.get("multiOmicsForm"))
+        sample_data = ms.get("sampleData")  # keep as-is
+
+        validation_state = _ensure_dict(ms.get("validationState"))
+
+        # ALWAYS set per-form validation lists (empty list if missing)
+        study_form["validation"] = _ensure_list(validation_state.get("studyForm"))
```
pkalita-lbl: I think this changes the semantics of the validation field. The validation state of the normal forms (e.g. `studyForm`) is an empty list if the form has been successfully validated, a non-empty list of strings if the form has been unsuccessfully validated, but `null` if the form hasn't been validated at all. It looks like `_ensure_list` will change `null` (not validated) to an empty list (successfully validated).

nmdc_server/api.py
```diff
@@ -1506,7 +1510,7 @@ async def get_submission(
             submission
         )
         submission_metadata_schema.permission_level = permission_level
-
+        print(submission_metadata_schema)
```
pkalita-lbl: Remove debugging

web/src/views/SubmissionPortal/Components/TemplateChooser.vue
```diff
@@ -53,10 +56,10 @@ export default defineComponent({
       </template>
     </PageTitle>
     <SubmissionForm
-      @valid-state-changed="(state) => validationState.sampleEnvironmentForm = state"
+      @valid-state-changed="handleStateChanged"
```
pkalita-lbl: I would either keep this inlined or rename the function `handleValidationStateChange` in order to clarify that it is _not_ called on general form state changes.

nmdc_server/migrations/versions/f9630762bd00_move_validation_to_each_form.py
JamesTessmer: Ah yeah, that's a good catch. I was trying to be smart and avoid having to write a bunch of type guards and didn't think about how it would affect the validation semantics.

web/src/views/SubmissionPortal/Components/TemplateChooser.vue
JamesTessmer: Reverted this, went back to doing it inline. The function addition was for an initial first pass that was different from what we discussed.

JamesTessmer:
@pkalita-lbl Pushed up changes for what you pointed out.

JamesTessmer:
@pkalita-lbl Got main merged in to resolve some conflicts, should be ready for review again. I'd assume you're pretty busy with production release so no rush.

nmdc_server/migrations/versions/f9630762bd00_move_validation_to_each_form.py
```diff
@@ -0,0 +1,186 @@
...
+        # sampleData becomes an object with {data, validation}
+        sample_metadata_validation = validation_state.get("sampleMetadata")
+        new_sample_data = {"data": sample_data}
+        if isinstance(sample_metadata_validation, dict):
+            new_sample_data["validation"] = sample_metadata_validation
+        else:
+            # if old value missing, backfill an empty validation state object
+            new_sample_data["validation"] = {"invalidCells": {}, "tabsValidated": {}}
+
+        new_ms = ms
```
pkalita-lbl: I loaded up my local database with the latest production data and then started up my local dev stack. The backend logs indicated that it ran the migration (btw, you might want to consider updating the docstring at the top of this file to be short description, two line breaks, longer description otherwise the long multiline description gets printed to the logs -- see `09a705b0bc00_submission_images.py` for an example to follow). After the migration ran I **did not** see the restructuring that this migration performs. I believe it's because line 91 only makes a new reference to the same underlying dict. Therefore the assignment in line 105 is not enough to let SQLAlchemy know the `row` object has changed. Creating a deep copy here should fix the issue. I'd recommend always loading up with a recent production dump when testing out migrations locally. If you need help getting that set up let me know.

```suggestion
        new_ms = deepcopy(ms)
```

JamesTessmer: Ah dang it. Thanks for checking, good point on using a fresh dump. I didn’t get any errors when running the migrator but it makes sense that incremental changes during development would affect that.

I’ll make the changes today and load a fresh db dump locally to test.

JamesTessmer: @pkalita-lbl Made this update. Tested the migration with a fresh DB and doing the deep copy looks like it works.

web/src/views/SubmissionPortal/store/index.ts
```diff
@@ -649,7 +645,11 @@ const templateList = computed<string[]>((prevTemplates) => {
 /**
  * DataHarmonizer Step
  */
-const sampleData = shallowRef({} as Record<string, any[]>);
+const sampleDataDefault = {
+  data: {} as Record<string, any[]>,
+  validation: ref<SampleMetadataValidationState | null>(null),
```
pkalita-lbl: During testing I observed that if I opened an existing submission, returned to the Submission Portal Home page, and then created a new submission, the new submission would get the same sample metadata validation state that the _previously opened_ submission had. I believe the root cause is have a `ref` here in the default value. It should probably be:
```suggestion
  validation: null as SampleMetadataValidationState | null,
```

nmdc_server/schemas_submission.py
```diff
@@ -115,43 +117,44 @@ class AddressForm(BaseModel):
...
 
-class SubmissionValidationState(BaseModel):
-    studyForm: Optional[List[str]] = None
-    multiOmicsForm: Optional[List[str]] = None
-    sampleEnvironmentForm: Optional[List[str]] = None
-    senderShippingInfoForm: Optional[List[str]] = None
-    sampleMetadata: Optional[SampleMetadataValidationState] = None
+class SampleData(BaseModel):
+    data: Dict[str, List[Any]]
+    validation: SampleMetadataValidationState
```
pkalita-lbl: I believe this needs to be optional in order to match the frontend type's definition and the fact that the default is `null`.

```suggestion
    validation: Optional[SampleMetadataValidationState] = None
```

pkalita-lbl (commented):
A couple of things I ran into while testing this branch locally

pkalita-lbl (approved):
I think this is good to go now!

PR #2130, opened 2026-05-14
Automatically add PI as contributor with 'editor' role

cristina-stonepedraza:
Changes made in this MR: 
- Add watcher to add a PI as a contributor automatically with editor role access when the author adds their ORCID
- If the PI changes, edit the auto-entry accordingly
- Add watcher to sync up the PI name if the ORCID stays the same, but the name field is changed

This allows the PI to be automatically added as a contributor with edit access to the submission.


web/src/views/SubmissionPortal/store/index.ts
```diff
@@ -883,6 +883,44 @@ async function loadRecord(id: string) {
...
+      );
+      if (idx !== -1) {
+        studyForm.contributors.splice(idx, 1);
+      }
+    }
+    if (newPiOrcid && !studyForm.contributors.some((c: { orcid: string }) => c.orcid === newPiOrcid)) {
+      studyForm.contributors.push({
+        name: studyForm.piName,
+        orcid: newPiOrcid,
+        roles: [],
```
pkalita-lbl, line 903: Should we include the `Principal Investigator` role by default as well? @mslarae13

pkalita-lbl (commented):
One question but otherwise looks good.

web/src/views/SubmissionPortal/store/index.ts
cristina-stonepedraza: I think it makes sense to include the PI role by default. I can make this quick change. @mslarae13 @pkalita-lbl

mslarae13: If I'm understand the Q, yes, PI should have full access and edit access

mslarae13:
FIles changed is 0. Status?

cristina-stonepedraza:
Sorry for the delay with this one - I rebased and applied the previous changes, along with the auto-add PI role comment.

pkalita-lbl (approved):

PR #2149, opened 2026-05-26
Fix regression with gene function searching facet

### LINKED ISSUE #2146: Faceted functional search is not working
I'm not sure when this regression happened but a user should be able to search both by term name and term ID for functional search and the term name should show up when a user enters a term ID in the search box.

The last time this was a problem is documented in this issue: https://github.com/microbiomedata/nmdc-server/issues/1450

This problem appears to be even worse for COG & PFAM, terms that should be supported (COG0105 & PF00005) are not found.

codytodonnell:
Resolves #2146 

Restores Function facets search capability. Terms can be searched by text or term and both support partial matches.

Screenshot 2026-05-26 at 3 49 16 PM

Screenshot 2026-05-26 at 4 10 33 PM

### DIFF
--- a/nmdc_server/crud.py
+++ b/nmdc_server/crud.py
@@ -328,7 +328,7 @@ def kegg_text_search(db: Session, query: str, limit: int) -> List[models.KoTermT
     term = query.replace(pathway_prefix, "map") if pathway_prefix else query
     q = (
         db.query(models.KoTermText)
-        .filter(models.KoTermText.text.ilike(f"%{term}%") | models.KoTermText.term.ilike(term))
+        .filter(models.KoTermText.text.ilike(f"%{term}%") | models.KoTermText.term.ilike(f"%{term}%"))
         .order_by(models.KoTermText.term)
         .limit(limit)
     )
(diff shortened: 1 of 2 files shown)


nmdc_server/crud.py
```diff
@@ -328,7 +328,9 @@ def kegg_text_search(db: Session, query: str, limit: int) -> List[models.KoTermT
     term = query.replace(pathway_prefix, "map") if pathway_prefix else query
     q = (
         db.query(models.KoTermText)
-        .filter(models.KoTermText.text.ilike(f"%{term}%") | models.KoTermText.term.ilike(term))
+        .filter(
+            models.KoTermText.text.ilike(f"%{term}%") | models.KoTermText.term.ilike(f"%{term}%")
```
eecavanna, line 332: Note for other reviewers: The expression before the `|` says `.text` and the expression after the `|` says `.term` (took me a minute to realize this).

web/src/components/FilterGene.vue
```diff
@@ -63,9 +68,10 @@ export default defineComponent({
 
     async function getGeneResults() {
       const resp = await geneSearch();
-      const results = [];
-      if (resp) resp
-        .map((v: KeggTermSearchResponse) => ({ text: getTermDisplayText(v.term, v.text), value: v.term }));
+      let results: GeneResult[] = [];
+      if (resp) {
+        results = resp.map((v: KeggTermSearchResponse) => ({ text: getTermDisplayText(v.term, v.text), value: v.term }));
```
eecavanna, line 73: Looks like the previous code was discarding the value returned by `map`. The new code makes sense to me.

eecavanna (approved):
The changes look good to me.

PR #2150, opened 2026-05-27
Fix metaproteomic analysis facet error

### LINKED ISSUE #2147: Error throws when using Metaproteomic Analysis facet
There's an error thrown on Sentry on the `/study/search` endpoint when using the Metaproteomic Analysis facet.

> Don't know how to join to <Mapper at 0x7b998f925340; MetaproteomicAnalysis>. Please use the .select_from() method to establish an explicit left side, as well as providing an explicit ON clause if not present already to help resolve the ambiguity.

> nmdc_server/api.py:653

codytodonnell:
Resolves #2147 

Use of the Metaproteomics Analysis Category filter in the Data Portal was triggering this backend error:

> Don't know how to join to <Mapper at 0x7b998f925340; MetaproteomicAnalysis>. Please use the .select_from() method to establish an explicit left side, as well as providing an explicit ON clause if not present already to help resolve the ambiguity.

This caused the study and consortia results to be incorrect (unfiltered) when using this facet.

The fix targets the overarching filter classes: `BaseFilter`, `WorkflowExecutionFilter`, and `MetaproteomicAnalysisFilter`. There were two main changes made here:
1. The `matches` query in `BaseFilter` was resulting in ambiguous SQL because of the cases where the target table was an `AliasedClass`.
2. For  `WorkflowExecutionFilter` and `MetaproteomicAnalysisFilter`, the `join_self` function needed to be overridden so that the proper association tables were found/used for those subqueries.

I'll note, because it wasn't obvious to me at first, that these filters are invoked in query.py on various instantiated filters created with `create_filter_class()` (e.g. [line 477 in query.py](https://github.com/microbiomedata/nmdc-server/blob/main/nmdc_server/query.py#L477)). The joins are triggered from the `matches` function which gets called on those instantiated filter classes.

AI disclosure: Claude Sonnet 4.6 was used to assist the diagnosis and solution for this bug.

### DIFF
--- a/nmdc_server/filters.py
+++ b/nmdc_server/filters.py
@@ -79,6 +79,7 @@ class BaseFilter:
         target_table: Table,
     ) -> Query:
         """Get a query representing unique id's of the target table matching the conditions."""
+        # Note that select_from is necessary to avoid ambiguity when the target_table.model is an AliasedClass
         query = db.query(func.distinct(target_table.model.id).label("id")).select_from(
             target_table.model
         )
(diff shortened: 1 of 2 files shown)


pkalita-lbl (approved):

eecavanna (approved):
Looks good to me. Thanks for including the explanatory comments above the code changes!

PR #2151, opened 2026-05-27
Restore update of `modifiedDate` state

pkalita-lbl:
I noticed on dev (**not** on prod) that the Last Modified timestamp of submissions were not being presented on the submission summary page:
image

This was caused by the `modifiedDate` ref not being updated in `updateStateFromRecord`. I believe this was accidentally dropped in https://github.com/microbiomedata/nmdc-server/commit/f06ca7b238f688822acd7938db61eb7b662b8e8c#diff-04ee0d419cc3e4c8a706fb6a5783bfadb69775408126ee2d65033126815bf772L833.

### DIFF
--- a/web/src/views/SubmissionPortal/store/index.ts
+++ b/web/src/views/SubmissionPortal/store/index.ts
@@ -859,6 +859,7 @@ function updateStateFromRecord(record: MetadataSubmissionRecord) {
     Object.assign(sampleData, record.metadata_submission.sampleData);
   }
   createdDate.value = new Date(record.created + 'Z');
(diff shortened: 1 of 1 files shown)


pkalita-lbl:
@JamesTessmer hopefully you can confirm that the original removal was just an oversight

JamesTessmer:
Pulled the branch and the fix looks good. 
Yeah I think this was an oversight. If I had to guess that line just got accidentally overwritten, there's no functional reason it should have been removed.

JamesTessmer (approved):

PR #2172, opened 2026-06-03
Implement study consortia redesign

### LINKED ISSUE #2087: Sample display of alternative identifiers when there are many identifiers listed
For some of the newly load NEON samples it says `Sample...`, I think this is b/c there isn't enough room to show all the ncbi and img identifiers (there are 5 total in this example). Normally this says `Sample Identifiers` and then shows insdc and img identifiers. Is there something we can do to display this better?

Search filter: https://data-dev.microbiomedata.org/?q=CiAICBAaGAoiGCJubWRjOm9tcHJjLTExLXhxeTR5dzk3Ig

Image

### LINKED ISSUE #2153: Implement Study/Consortia redesign
Based on #2086 and feedback from the overall group, implement the updated study/consortia layout in the Data Portal UI.

codytodonnell:
Resolves #2153 and resolves #2087

### UI Changes

This PR modifies the UI and UX of the search results. Studies and consortia are now baked into a single section called "Studies." The samples results have been moved into a tab group that is combined with the studies results. Studies is the default visible tab.

The result cards themselves have also been updated. They now include IDs with click-to-copy behavior. The titles are now clickable links to the respective detail pages. Study children have been slightly modified visually (still collapsed/expandable). Alternative identifiers are now rendered in a section labeled "External" and if the width of this section is beyond 500px, then it turns into a [slide group](https://vuetifyjs.com/en/components/slide-groups/#usage).

Links app-wide have been given a default color for both visited and unvisited.

### UX and Functionality Changes

The page no longer scrolls to the top when a filter condition is added/removed. Sample filters will be reflected immediately even if you are beyond page 1 (this was an existing bug). If you are beyond page one and the filters change, your results will reset to page 1.

When filtering by a study ID that has children (using the checkbox feature or the Study ID LIKE filter), the samples associated with its child studies are automatically included in the results.

### Screenshots

Screenshot 2026-06-12 at 4 41 37 PM

Screenshot 2026-06-12 at 4 42 39 PM

### Still to Come

These issues will be addressed in a future PR:
- #2186
- #2187

### DIFF
--- a/web/src/views/Search/SearchLayout.vue
+++ b/web/src/views/Search/SearchLayout.vue
@@ -1,6 +1,6 @@
 <script setup lang="ts">
 import {
-  computed, ref,
+  computed, Ref, ref,
 
 } from 'vue';
 
@@ -126,10 +126,10 @@ const gatedEnvironmentVisConditions = useClockGate(
   computed(() => (visTab.value === 1)),
   stateRefs.conditions,
 );
-// const showChildren:Ref<any[]> = ref([]);
-// function toggleChildren(value:StudySearchResult) {
-//   showChildren.value.includes(value.id) ? showChildren.value.splice(showChildren.value.indexOf(value.id), 1) : showChildren.value.push(value.id);
-// }
+const showChildren: Ref<any[]> = ref([]);
+function toggleChildren(value:StudySearchResult) {
+  showChildren.value.includes(value.id) ? showChildren.value.splice(showChildren.value.indexOf(value.id), 1) : showChildren.value.push(value.id);
+}
 </script>
 
 <template>
@@ -219,7 +219,7 @@ const gatedEnvironmentVisConditions = useClockGate(
                   @set-page="study.setPage($event)"
                   @set-items-per-page="study.setItemsPerPage($event)"
                 >
-                  <template #prepend-action="{ result }">
+                  <template #action-left="{ result }">
                     <v-list-item-action>
                       <v-checkbox-btn
                         :model-value="studyCheckboxState"
@@ -227,6 +227,26 @@ const gatedEnvironmentVisConditions = useClockGate(
                         @click.stop
                         @change="setChecked($event.target.checked, result.id, result.children as StudySearchResult[])"
                       />
+                    </v-list-item-action>
+                  </template>
+                  <template #action-title-right="{ result }">
+                    <v-list-item-action
+                      v-if="result.children && result.children.length > 0"
+                      class="ml-2"
+                    >
+                      <v-btn
+                        variant="flat"
+                        color="grey"
+                        size="x-small"
+                        @click="toggleChildren(result as StudySearchResult)"
+                      >
+                        {{ result.children.length }} child {{ result.children.length > 1 ? 'studies' : 'study' }}
+                        <template #append>
+                          <v-icon>
+                            {{ showChildren.includes(result.id) ? 'mdi-chevron-up' : 'mdi-chevron-down' }}
+                          </v-icon>
+                        </template>
+                      </v-btn>
                     </v-list-item-action>
                   </template>
                   <template #action-right="{ result }">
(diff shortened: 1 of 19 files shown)


eecavanna:
Based on the PR description, alone, the changes sound good to me. I'll review the diff next chance I get.

eecavanna:
I think I can request reviews from `codex` in this repo, so I'll try to do that now.

@codex review

web/src/views/Search/SearchLayout.vue
```diff
@@ -20,212 +20,106 @@ import usePaginatedResults from '@/use/usePaginatedResults';
...
-    const showChildren:Ref<any[]> = ref([]);
-    function toggleChildren(value:StudySearchResults) {
+const biosample = usePaginatedResults(stateRefs.conditions, api.searchBiosample, dataObjectFilter, 10);
+const studyType = types.study;
+const studySummaryData = useFacetSummaryData({
+  field: ref('study_id'),
+  table: ref('study'),
+  conditions: stateRefs.conditions,
+});
+// const studyCondition = studySummaryData.otherConditions.value || [];
```
eecavanna: I don't know whether this comment was left in intentionally or by mistake.

web/src/views/Search/SearchLayout.vue
```diff
@@ -266,418 +160,244 @@ export default defineComponent({
...
-                        <v-btn
-                          icon
-                          variant="plain"
-                          size="large"
-                          :to="{ name: 'Study', params: { id: result.id } }"
-                        >
-                          <v-icon>
-                            mdi-chevron-right
-                          </v-icon>
+                          {{ result.children.length }} child {{ result.children.length > 1 ? 'studies' : 'study' }}
```
eecavanna, line 272: Side note: I typically check for `== 1` or `!= 1` when doing this (in any codebase) so that "0" is also grammatically correct. In this case, maybe the UI element would not appear for values of "0" anyway.

web/src/views/Search/SearchLayout.vue
```diff
@@ -266,418 +160,244 @@ export default defineComponent({
...
-                          </v-icon>
-                        </v-btn>
-                      </v-list-item-action>
-                    </template>
-                    <template #item-content="props">
-                      <div v-if="props.result.omics_processing_counts">
-                        <template
-                          v-for="item in props.result.omics_processing_counts"
+                        <v-chip
+                          v-if="(item as any).count"
```
eecavanna: Note to self: look into this `any` usage, which I assume is to appease TypeScript's type checker after it complained about something that would have been tedious to address the root cause of.

eecavanna (approved):
I reviewed the diff on my tiny iPhone screen. It looked good to me. I noticed several changes that I thought weren't technically necessary, but I think were made to make the code clearer and easier to maintain. Thanks for making those!

I left a few comments. None were deal breakers. I'm going to submit this review and then check what Codex reported.

eecavanna:
This PR looks good to me and I'm comfortable with it being merged in so it can undergo pre-release testing for the upcoming release. I'm also OK with the alternative: holding off on merging until after the release. The product owner may have a preference. CC: @aclum

web/src/views/Search/SearchLayout.vue
codytodonnell: Good catch!

codytodonnell: But yes, in this particular case, the button only shows if there's at least one item in the children array anyways

web/src/views/Search/SearchLayout.vue
codytodonnell: I think this was existing code, but I'm realizing the cast isn't necessary anymore because the search result type has become more specific. Removing.

web/src/views/Search/SearchLayout.vue
codytodonnell: Removed

PR #2179, opened 2026-06-04
Disable job that restarts Spin-hosted deployments via Rancher API

### LINKED ISSUE #2178: Update GHA workflow to refrain from trying to restart workloads on Spin
We no longer host things on Spin (with a few exceptions unrelated to this ticket), but the deployment GHA workflow still tries to restart workloads there, causing the GHA to workflow to fail. This ticket represents the task of disabling the Spin deployment functionality within the GHA workflow that restarts deployments. I plan to do it with GHA's `if: false` directive on the relevant step(s), so that we can easily switch back to Spin if necessary, and so we don't have to rename the workflows yet (we can do that in a future cleanup effort).

Issue cloned from https://github.com/microbiomedata/nmdc-runtime/issues/1521

eecavanna:
On this branch, I updated the `.github/workflows/deploy.yml` GHA workflow that builds container images and restarts workflows (on Spin and GKE), so that it no longer restart workloads on Spin. It still restarts them on GKE.

Nowadays, we host things on GKE instead of Spin. Attempting to restart workloads on Spin causes the job (and its containing workflow) to fail. That triggers a notification to the committer, and causes the GitHub UI to show a red icon next to the workflow run, both of which degrade the "developer experience."

Fixes #2178

This is the `nmdc-server` analogue to the following `nmdc-runtime` PR: https://github.com/microbiomedata/nmdc-runtime/pull/1522

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -123,7 +123,16 @@ jobs:
 
   # Use the Rancher API to redeploy the specified deployments,
   # causing them to use the newly-pushed container images.
+  #
+  # Note: 🚫 This job is intentionally disabled via `if: ${{ false }}`.
+  #          It may be removed entirely in a future commit.
+  #
+  # Note: This `redeploy` job is the Spin counterpart to the subsequently-added `restart-dev-deployments` job below.
+  #       TODO: Standardize the jobs' names, if we keep this `redeploy` job long term.
+  #
   redeploy:
+    if: ${{ false }}
+
     needs: build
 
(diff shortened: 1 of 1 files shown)


eecavanna:
The analogous change in `nmdc-runtime` worked as designed, per this screenshot of the first affected workflow run:

image

pkalita-lbl (approved):

eecavanna:
The change worked as designed [in `nmdc-server` also], as shown in this screenshot of the first affected workflow run:

image

Issue #2084, opened 2026-04-16
Submission Portal - UI changes to support isolate modeling


aclum:
Add a radio button for `Isolate` to questions about datatypes in the Multi-omics Data 
- [ ] if the facilty is `JGI` 
- [ ] if a user is not submitting data to a user facility (has data been generated can by `yes` or `no`) 

Update what tabs are shown based on a selection of `isolate`
- [ ] if users are shipping samples to a user facility &  facility is `JGI` & `isolate` is selected then show the `isolate` and `jgi_isolate` tabs
- [ ] users are not submitted data to a user facility & (has the data been generated is `yes` or `no`)  & `isolate is selected then show the `isolate` tab.

@mslarae13 to confirm requirements

pkalita-lbl:
Are we supporting isolates without sampling environments (I think someone described this as the "ordered something from a catalog" use case)? If yes, does this become not required if-and-only-if Isolate is selected in the previous step?

<img width="619" height="663" alt="Image" src="https://github.com/user-attachments/assets/73569e73-7c51-49a7-ac88-49609b7596cb" />

aclum:
I'd say yes. If a user only selects isolate then you wouldn't display the Sample Environment selection interface.

aclum:
Patrick has been working on mockups which were discussed yesterday. There was some discussion about 'isolate' not being a data type & the need to separate genome from transcriptiome so this could be 'isolate genome' & 'isolate transcriptome'.

ssarrafan:
Appears to be active with related MRs. Moving to current sprint. @aclum @pkalita-lbl

ssarrafan:
Patrick is OOO today so I'll move this to the next sprint for updates. Is this still active @pkalita-lbl? No updates in 2 weeks.

aclum:
This is done, changes are in dev and will be in prod after the nmdc-server release on Monday. Completed by https://github.com/microbiomedata/nmdc-server/pull/2185

PR #2185, opened 2026-06-11
Integrate new isolate schema classes into Submission Portal UI

pkalita-lbl:
Related to #2084 (intentionally not closing yet)

### Summary

These changes are the bare minimum integration of new isolate-related `nmdc-submission-schema` classes into the Submission Portal UI. I'm intentionally wanting to keep the issue open after this PR is merged because I want to ask squad members to test on dev before closing it out.

### Details

* Upgrade `nmdc-schema` and `nmdc-submission-schema` dependencies both to version `11.20.0`.
* Repeat existing patterns for adding "Isolate Genome" and "Isolate Transcriptome" checkboxes to the non-DOE data type options _and_ the JGI data type options.
* Update the form field dependency logic to add the "isolate" template if any of those analysis types were selected and add the correct JGI tab(s) to the Sample Metadata screen if one of the JGI types was selected.
* The only somewhat new idea here is in `web/src/views/SubmissionPortal/Components/TemplateChooser.vue` where the "isolate" template checkbox can be disabled for a reason _other than_ if there are rows in DataHarmonizer for that template. Because there are now multiple ways to get a disabled checkbox, I've added a message conditionally displayed below each checkbox explaining why it is disabled.


web/src/views/SubmissionPortal/Components/TemplateChooser.vue
```diff
@@ -18,7 +19,27 @@ export default defineComponent({
...
+      Object.entries(HARMONIZER_TEMPLATES).forEach(([key, template]) => {
+        if (templateHasData(template.sampleDataSlot)) {
+          notes[key] = 'This template cannot be deselected because there is data present in the Sample Metadata tab for this template.';
+        } else if (key === 'isolate' && (
+          multiOmicsForm.omicsProcessingTypes.includes('isolate-genome') ||
+          multiOmicsForm.omicsProcessingTypes.includes('isolate-transcriptome') ||
+          multiOmicsForm.omicsProcessingTypes.includes('isolate-genome-jgi') ||
+          multiOmicsForm.omicsProcessingTypes.includes('isolate-transcriptome-jgi')
+        )) {
+          notes[key] = 'This template cannot be deselected because an isolate omics processing type is selected in Multi-omics Data page.';
```
eecavanna: Quick note from iPhone:

selected "on the" instead of "in" ?

web/src/views/SubmissionPortal/Components/TemplateChooser.vue
```diff
@@ -45,7 +66,8 @@ export default defineComponent({
           target="_blank"
           rel="noopener noreferrer"
         >MIxS Extension</a>
-        for your samples.
+        for your samples. If isolation was performed, select the "isolate" option. If isolates were obtained from a
+        provider, select only the "isolate" option.
```
eecavanna, line 70: Maybe italicize the "only" (if possible here), assuming the words are all accurate.

web/src/views/SubmissionPortal/store/index.ts
```diff
@@ -530,6 +533,31 @@ watch(() => multiOmicsForm.omicsProcessingTypes, (newValue, oldValue) => {
   if (!newValue.includes('lipidome') && oldValue.includes('lipidome')) {
     multiOmicsForm.lipProtocols = null;
   }
+
+  // isolate genome was added, the isolate environment template is automatically added as well.
```
eecavanna, line 537: Maybe add "if" to the beginning of the comment

(applies to all of these)

eecavanna (approved):
LGTM! Reviewed carefully from iPhone. Left some comments from my tiny keyboard here.

web/src/views/SubmissionPortal/Components/DoeFacility.vue
```diff
@@ -186,6 +186,20 @@ export default defineComponent({
...
+        v-model="multiOmicsForm.omicsProcessingTypes"
+        label="Isolate Genome"
+        value="isolate-genome-jgi"
+        :disabled="templateHasData(HARMONIZER_TEMPLATES.jgi_isolate_genome?.sampleDataSlot) || undefined"
+        hide-details
+      />
+      <v-checkbox
+        v-model="multiOmicsForm.omicsProcessingTypes"
+        label="Isolate Transcriptome"
+        value="isolate-transcriptome-jgi"
```
codytodonnell, line 199: This value and the values in the other isolate checkboxes look like they are written in a very specific way which presumably aligns with how the db stores these data. I'm wondering if it would be good to have an enum or union type on the frontend with these acceptable values, just to make it more clear that they are intentionally formatted and to avoid mistakes.

web/src/views/SubmissionPortal/Components/TemplateChooser.vue
codytodonnell: I agree this reads a little bit confusing. Without knowing the full context or intent, this is how I'm interpreting the statement:

> If isolation was performed, include the "isolate" option in your selection. If isolates were obtained from a provider, select _only_ the "isolate" option and no others.

web/src/views/SubmissionPortal/Components/TemplateChooser.vue
pkalita-lbl: Fixed

web/src/views/SubmissionPortal/Components/TemplateChooser.vue
pkalita-lbl: This is basically draft text that I expect to be refined once people see it on dev.

web/src/views/SubmissionPortal/store/index.ts
pkalita-lbl: Leaving as is to match existing comments. I think the the intention was to capture the action that the subsequent block is handling.

web/src/views/SubmissionPortal/Components/DoeFacility.vue
pkalita-lbl: Whoops! I merged right as you were reviewing. But to answer your question the `omicsProcessingTypes` field is typed with this literal string union: https://github.com/microbiomedata/nmdc-server/blob/fac8a231ef560f97b7daf3b8f20cc997f8a5d9fe/web/src/views/SubmissionPortal/store/index.ts#L377-L382

codytodonnell: Ok nice, I saw that farther down the diff but I wasn't 100% it was connected to the checkbox values. Thanks!

PR #2203, opened 2026-06-22
Add regex enforcement for reviewer ORCiD

### LINKED ISSUE #2197: ORCID iD format is not enforced in Assign Reviewer popup
1. As an admin, go to the Submission Portal home page
2. For any submission, click the action overflow button (three dots) and then click "Assign Reviewer" button

**Expected**: In the Assign Reviewer popup, the ORCID iD field should minimally enforce that the user enters follows the ORCID iD format.
**Actual**: The user is allowed to enter anything.

Image

This can lead to subtle issues like what @mslarae13 reported in https://github.com/microbiomedata/issues/issues/1761. She had entered @snmiller5's ORCID iD with several trailing spaces (i.e. "0000-0000-0000-0000&nbsp;&nbsp;&nbsp;&nbsp;")

JamesTessmer:
Fixes https://github.com/microbiomedata/nmdc-server/issues/2197

This PR adds a check on the entered ORCiD and disables the `add reviewer` button if it doesn't match the expected regex, or if there is no ID entered.

### DIFF
--- a/web/src/views/SubmissionPortal/Components/StudyForm.vue
+++ b/web/src/views/SubmissionPortal/Components/StudyForm.vue
@@ -11,6 +11,7 @@ import {
   primaryStudyImageUrl,
   studyForm,
 } from '../store';
+import { validateOrcid } from '../utils.ts';
 import { PermissionTitle } from '@/views/SubmissionPortal/types';
 import { stateRefs } from '@/store';
 import SubmissionDocsLink from './SubmissionDocsLink.vue';
@@ -142,6 +143,7 @@ export default defineComponent({
       primaryStudyImageUrl,
       piImageUrl,
       revalidate,
+      validateOrcid,
     };
   },
 });
@@ -242,9 +244,7 @@ export default defineComponent({
           <v-text-field
             v-model="studyForm.piOrcid"
             label="ORCID iD"
-            :rules="[
-              v => !v || /(\d{4}-){3}\d{3}(\d|X)/.test(v) || 'ORCID iD must be in valid format (0000-0000-0000-0000)',
-            ]"
+            :rules="[validateOrcid]"
             :disabled="!isOwner() || currentUserOrcid === studyForm.piOrcid || undefined"
             variant="outlined"
             :hint="Definitions.piOrcid"
--- a/web/src/views/SubmissionPortal/Components/SubmissionList.vue
+++ b/web/src/views/SubmissionPortal/Components/SubmissionList.vue
@@ -1,10 +1,11 @@
 <script lang="ts">
 import {
   defineComponent, ref, watch, onMounted,
-  computed,
+  useTemplateRef,
 } from 'vue';
 import { useRouter } from 'vue-router';
 import { DataTableHeader } from 'vuetify';
+import { VTextField } from 'vuetify/components';
 import usePaginatedResults from '@/use/usePaginatedResults';
 import {
   SubmissionStatusEnum, editableByStatus, formatStatusTransitions,
@@ -26,6 +27,7 @@ import {
   StatusOption,
 } from '@/views/SubmissionPortal/types';
 import { stateRefs } from '@/store';
+import { validateOrcid } from '../utils.ts';
 import useRequest from '@/use/useRequest';
 
 const headers: DataTableHeader[] = [
@@ -165,9 +167,7 @@ export default defineComponent({
     }
 
     const reviewerOrcid = ref('');
-    const invalidOrcid = computed(() => {
-      return !/(\d{4}-){3}\d{3}(\d|X)/.test(reviewerOrcid.value);
-    });
+    const orcidTextFieldRef = useTemplateRef<InstanceType<typeof VTextField>>('orcidTextField');
     function openReviewerDialog(item: MetadataSubmissionRecordSlim | null) {
       isReviewerAssignmentDialogOpen.value = true;
       selectedSubmission.value = item;
@@ -230,7 +230,8 @@ export default defineComponent({
       isReviewerAssignmentDialogOpen,
       isReviewerForSubmission,
       reviewerOrcid,
-      invalidOrcid,
+      orcidTextFieldRef,
+      validateOrcid,
       IconBar,
       IntroBlurb,
       TitleBanner,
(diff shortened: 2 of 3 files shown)


pkalita-lbl (commented):
A couple of issues to take a look at:

1. The regex being used here `/(\d{4}-){3}\d{3}(\d|X)/` with the `test` method doesn't actually prevent the original issue. Because `test` searches for any _substring_ that matches, this will not prevent a value like `0000-0000-0000-0000    ` (with trailing whitespace). I would recommend adding `^` and `$` in the regex to do a full string match.
2. It feels a little off to do the validation in `rules` and again in a computed ref. I would recommend only using `rules`, getting a ref to the `v-text-field` with `useTemplateRef`, and disabling the button on the `isValid` [field](https://v3.vuetifyjs.com/en/api/v-text-field/#exposed-isValid).
3. I would define the ORCID validation in a function in `web/src/views/SubmissionPortal/utils.ts` that both `SubmissionList` and `StudyForm` can use.

web/src/views/SubmissionPortal/Components/SubmissionList.vue
```diff
@@ -537,6 +543,9 @@ export default defineComponent({
                 class="mt-4"
                 label="ORCiD"
                 variant="outlined"
+                :rules="[
+                  v => !v || /(\d{4}-){3}\d{3}(\d|X)/.test(v) || 'ORCID iD must be in valid format (0000-0000-0000-0000)',
```
codytodonnell: It might be good to define this rule function above with a descriptive name and then inject that function here:

```
:rules="[isValidOrcid]"
```

JamesTessmer: Good point I think that's in line with Patrick's suggestion as well

JamesTessmer:
> A couple of issues to take a look at:
> 
> 1. The regex being used here `/(\d{4}-){3}\d{3}(\d|X)/` with the `test` method doesn't actually prevent the original issue. Because `test` searches for any _substring_ that matches, this will not prevent a value like `0000-0000-0000-0000    ` (with trailing whitespace). I would recommend adding `^` and `$` in the regex to do a full string match.
> 2. It feels a little off to do the validation in `rules` and again in a computed ref. I would recommend only using `rules`, getting a ref to the `v-text-field` with `useTemplateRef`, and disabling the button on the `isValid` [field](https://v3.vuetifyjs.com/en/api/v-text-field/#exposed-isValid).
> 3. I would define the ORCID validation in a function in `web/src/views/SubmissionPortal/utils.ts` that both `SubmissionList` and `StudyForm` can use.

Ah thanks for pointing those out. I'll get on these changes

web/src/views/SubmissionPortal/Components/StudyForm.vue
```diff
@@ -242,9 +244,7 @@ export default defineComponent({
           <v-text-field
             v-model="studyForm.piOrcid"
             label="ORCID iD"
-            :rules="[
-              v => !v || /(\d{4}-){3}\d{3}(\d|X)/.test(v) || 'ORCID iD must be in valid format (0000-0000-0000-0000)',
-            ]"
+            :rules="[validateOrcid]"
```
pkalita-lbl, line 247: The rule previously had a `!v || ` at the beginning which essentially means "_if_ you provide a value then it must adhere to this pattern." The implementation of `validateOrcid` doesn't have that. It _always_ checks if the value matches the pattern. Since an empty string doesn't match the pattern, **this effectively makes this field required when it wasn't previously**.

This is where we want to take advantage of Vuetify's design pattern of specifying `rules` as an array of functions: https://v3.vuetifyjs.com/en/components/forms/#rules. The `validateOrcid` function should be accepting of empty strings (i.e. add the `!v || ` there). This makes it applicable to non-required fields. When an ORCID iD _is_ required (like in the `SubmissionList` modal), then you'd compose `validateOrcid` with another function that enforces the required-ness:

```vue
<v-text-field
  ...
  :rules="[
    (v) => !!v || "An ORCID iD is required",
    validateOrcid,
  ]"
/>
```

With that, if a value isn't provided, the first rule will return `An ORCID iD is required` and `validateOrcid` won't run. But if a value is provided, the first rule will return true and `validateOrcid` will run.

pkalita-lbl (changes requested):

web/src/views/SubmissionPortal/Components/StudyForm.vue
JamesTessmer: Ah thanks for explaining, that makes sense and I'll make those updates.
On this line `(v) => !!v || "An ORCID iD is required"` is there a need for the double negation or could it just be `(v) => v || "An ORCID iD is required"`?

After thinking about it I see why that doesn't work, there's not boolean conversion.

pkalita-lbl: What's necessary is that the [function](https://v3.vuetifyjs.com/en/components/text-fields/#validation) returns either a boolean (typically `true` to indicate the value is valid) or a string (indicating the value is not valid and providing the message to show to the user). Using `!!v` is a way to coerce `v` to a boolean. You could [do the same](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean/Boolean#return_value) with `Boolean(v)`. But without one of those, if you typed "hello" into the text box, the rule would evaluate `"hello" || "An ORCID iD is required"` and return "hello". That would indicate the value is invalid and the message to show the user is "hello".

JamesTessmer: Ah ok, I see in my head I was equating `v` to `Boolean(v)`. I pushed up the fixes for that.

web/src/views/SubmissionPortal/Components/SubmissionList.vue
```diff
@@ -533,10 +540,13 @@ export default defineComponent({
             </legend>
             <v-col cols="4">
               <v-text-field
+                ref="orcidTextField"
                 v-model="reviewerOrcid"
                 class="mt-4"
                 label="ORCiD"
                 variant="outlined"
+                :rules="[(v) => !!v || 'An ORCiD ID is required',
```
pkalita-lbl: Sorry, this is a pet peeve of mine -- and we mess it up all the time -- but the name of the organization is "ORCID". The identifier that they provide is an "ORCID iD". I know it looks dumb, but I don't make the rules: https://info.orcid.org/brand-guidelines/#The_ORCID_name

JamesTessmer: That's fair, if that's how they specify it to show up in usage.

pkalita-lbl (approved):
Looks good now!

PR #2207, opened 2026-06-23
Apply static column width to ribosomal sequence columns

### LINKED ISSUE #2205: Limit width of ribosomal sequence columns
The first ribosomal sequence (`isolate_ribosomal_seq`) and second ribosomal sequence (`isolate_second_ribosomal_seq`) columns are expected to hold quite long strings. By default DataHarmonizer attempts to make columns wide enough to fit their content. This means these columns can become unusably wide:

Image

This task is to override DataHarmonizer's default behavior for those particular columns. Those columns should ideally expand up to a maximum width and then wrap or visually truncate the contents thereafter.

pkalita-lbl:
Fixes #2205 

These changes use a DataHarmonizer feature we were already leveraging which allows clients (like `nmdc-server`) to make [Handsontable column](https://handsontable.com/docs/13.1/javascript-data-grid/configuration-options/#set-column-options) customizations on a per-slot basis. In this case the `width` of the column (which is normally automatically calculated) is being manually set for the columns representing the `isolate_ribosomal_seq` and `isolate_second_ribosomal_seq` slots. These columns are expected to contain very long string that aren't intended to be human-readable (i.e. they're just a series of ATCGNs), so I think visually truncating at a predefined width is okay here.

### DIFF
--- a/web/src/views/SubmissionPortal/harmonizerApi.ts
+++ b/web/src/views/SubmissionPortal/harmonizerApi.ts
@@ -146,7 +146,22 @@ export default class HarmonizerApi {
   }
 
   _getFieldSettings() {
-    const fieldSettings: any = {};
+    const fieldSettings: any = {
+      isolate_ribosomal_seq: {
+        getColumn: (_: any, col: {[key: string]: any}) => {
+          const newCol = { ...col };
+          newCol.width = 180;
+          return newCol;
+        }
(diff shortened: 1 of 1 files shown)


eecavanna:
@codex review

Pretty please

eecavanna (approved):

PR #2209, opened 2026-06-23
Add validation failure when well ID is provided but container is not plate

### LINKED ISSUE #2190: mechanism to enforce alignment between container type and well location
Testing the isolate changes last week highlighted a bug where linkml and data harmonizer aren't playing well together. There are two rules well_requires_plate and plate_requires_well which were written to check that if the container type is plate that a well location is populated and that if well location is populated that the container is a plate. @pkalita-lbl was able to get to the bottom of this & determined it was because the rule uses https://linkml.io/linkml-model/latest/docs/equals_string/ and `plate` and `well` are permissible values from an enumeration, not strings.

To be discussed at the metadata meeting, we either need the rule changed which may depend changes to the linkml metamodel, a cursory look doesn't turn up anything like equals_permissible_value OR we need to enforce this in python code in nmdc-server OR something else TBD

cc @sierra-moxon @cmungall @mslarae13

pkalita-lbl:
Fixes #2190 

The linked issue identifies two LinkML rules that are present in `submission-schema`. Because the rules fall into a bit of a LinkML grey area they aren't interpreted by DataHarmonizer they way one might expect. Therefore, the decision was to reimplement the rules in custom `nmdc-server` code. 

The existing code that checks container types, names, and well IDs for JGI standards already captured one of the LinkML rules (namely, when container type is `plate` then the well ID must be provided). These changes add the validation expressed by the other LinkML rule: if a well ID was provided the the container type must be `plate`.

### DIFF
--- a/web/src/views/SubmissionPortal/validation.test.ts
+++ b/web/src/views/SubmissionPortal/validation.test.ts
@@ -25,15 +25,17 @@ describe('validatePlateWellsForJgi', () => {
     ]);
   });
 
-  it ('flags rows with a well ID and cont_type other than "plate"', () => {
+  it('flags rows with a well ID and cont_type other than "plate"', () => {
     const issues = validatePlateWellsForJgi([
       { cont_type: 'tube', container_name: 'tube-1', cont_well: 'B1' },
       { cont_type: 'tube', container_name: 'tube-2', cont_well: 'C1' },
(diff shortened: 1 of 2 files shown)


eecavanna:
I'll review this now.

eecavanna (approved):
The PR description makes sense to me, and the code changes match it. They introduce a validation error when the user specifies a well when they did not specify ~a plate~ the container type as `plate`.

PR #2212, opened 2026-06-23
Fix bug where "Save" button does not dismiss email address modal

### LINKED ISSUE #2208: Users get stuck at email address entry model
Two users have reported that they have been unable to get past the email address entry modal that appears when the portal cannot retrieve an email address from the user's ORCID account.

The modal was designed to disappear once the user enters their email address into it. However, it is not disappearing. It is remaining visible and preventing the user from using the portal. 

We first received a report about this yesterday. We received a report about a second occurrence (with a different user) today. Both occurred in production. The latest production deployment took place on May 18, 2026.

We were able to help the first user pass the modal by manually populating the database with their preferred email address. Once we did that, the user was no longer stuck at the modal.

eecavanna:
On this branch, I made it so clicking the "Save" button and pressing the "Enter" key both cause the email address form to be submitted. To do that, I moved the "Save" button to _within_ the `v-form` (previously, it was outside of it), made the "Save" button be of type `"submit"`, and removed the "Save" button's sideband way of triggering a submission.

While doing that, I also extracted the email address validation function into a named helper function.

Finally, I made it so both the validation and the storage are happening with the same value (previously, validation was happening with a non-trimmed version of the entered string, but storage was happening with a _trimmed_ version of it).

Most of the changed lines in the diff are indentation-only changes (due to moving the `v-form` tag to an "outer" level of the tree).

### Testing

I tested both clicking the "Save" button and pressing the "Enter" key (when focused on the text field) locally, and both successfully updated the email address in the database. I also confirmed the value stored in the database was the trimmed one, even when the textbox contains leading/trailing whitespace.

### Tickets

Fixes #2208

### DIFF
--- a/web/src/components/UserEmailModal.vue
+++ b/web/src/components/UserEmailModal.vue
@@ -49,7 +49,7 @@ const updateUser = async (value:string) => {
 /**
  * Returns `true` if the specified string resembles an email address.
  * 
- * Note: This basic checks still accepts things like ` h e l l o @ example . com `.
+ * Note: This basic check still accepts things like ` u s e r @ example . com `.
  *
  * Docs: https://en.wikipedia.org/wiki/Email_address#Validation_and_verification
  */
--- a/web/src/components/UserEmailModal.vue
+++ b/web/src/components/UserEmailModal.vue
@@ -46,6 +46,13 @@ const updateUser = async (value:string) => {
   user.value = await updateUserRequest.request(() => api.updateUser(user.value!.id, update));
 };
 
+/**
+ * Returns `true` if the specified string resembles an email address (per a basic check).
+ *
+ * Docs: https://en.wikipedia.org/wiki/Email_address#Validation_and_verification 
+ */
+const validateEmailAddr = (s: string) => /.+@.+\..+/.test(s);
+
 const submitterEmail = ref(user.value?.email ?? '');
 
 // dialog is bound to the modal to determine whether to display or not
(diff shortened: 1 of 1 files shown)


eecavanna:
@codex review

web/src/components/UserEmailModal.vue
```diff
@@ -68,18 +80,19 @@ const updateEmail = async () => {
...
-      </v-card-text>
-      <v-card-text>
-        <v-form
-          ref="formRef"
-          class="my-6"
-        >
+      <v-form
+        ref="formRef"
+        class="my-6"
+        @submit.prevent="updateEmail"
```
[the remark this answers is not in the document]
eecavanna: Fixed in https://github.com/microbiomedata/nmdc-server/pull/2212/commits/f5979e3b0d81518409bb62b838dccd00851ce8df

web/src/components/UserEmailModal.vue
```diff
@@ -46,6 +46,15 @@ const updateUser = async (value:string) => {
   user.value = await updateUserRequest.request(() => api.updateUser(user.value!.id, update));
 };
 
+/**
+ * Returns `true` if the specified string resembles an email address.
+ * 
+ * Note: This basic checks still accepts things like ` h e l l o @ example . com `.
```
eecavanna: Fix typo

```suggestion
 * Note: This basic check still accepts things like ` u s e r @ example . com `.
```

web/src/components/UserEmailModal.vue
```diff
@@ -46,6 +46,15 @@ const updateUser = async (value:string) => {
...
+ *
+ * Docs: https://en.wikipedia.org/wiki/Email_address#Validation_and_verification
```
eecavanna: Make comment more accurate

```suggestion
 * Reference: https://en.wikipedia.org/wiki/Email_address#Validation_and_verification
```

pkalita-lbl (approved):
I didn't run this branch locally, but this is the change I was expecting to see.

eecavanna:
Thanks!

Merging (will go out to prod during next Monday's release).

Props to @pkalita-lbl and @mslarae13 for all the upstream work that preceded this PR: user support, investigation, and user research. Thank you!

PR #2219, opened 2026-06-24
Do not modify sample set status in submission finalize handler

pkalita-lbl:
Related to https://github.com/microbiomedata/nmdc-runtime/issues/1496

These are changes that go along with https://github.com/microbiomedata/nmdc-runtime/pull/1546. These changes can be merged into `feat-submission-sample-sets` independently of the `nmdc-runtime` PR.

These changes update the `POST /api/metadata_submission/{submission_id}/finalize` endpoint so that it does not modify any of the submission's sample sets. Instead `nmdc-runtime` will be in charge of calling the `PATCH /api/metadata_submission/sample_set/{sample_set_id}/status` endpoint and the necessary time.

### DIFF
--- a/nmdc_server/api.py
+++ b/nmdc_server/api.py
@@ -1855,11 +1855,11 @@ async def generate_signed_upload_url(
 
 
 @router.post(
-    "/metadata_submission/{id}/finalize",
+    "/metadata_submission/{submission_id}/finalize",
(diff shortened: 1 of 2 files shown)


JamesTessmer (approved):
Looks good to me. Pulled locally and tested, didn't encounter any errors.

PR #2233, opened 2026-07-01
Fix pasted Data Portal filter URLs

### LINKED ISSUE #2230: Data Portal - paste of a filter url does not apply the correct filter
I suspect this is related to the new tab behavior but copy and pasting a url from one window to another is not working.

to reproduce either in prod or dev apply a filter `PI Name [is] Kate Thibault` for example, should return 1 study with 3 child studies. Copy the URL to a new tab and paste and it returns all of the data instead of applying the filter.

candidate for hot fix

cc @codytodonnell @shreddd @eecavanna @pkalita-lbl

eecavanna:
> [!IMPORTANT]
> Codex opened this PR on @eecavanna's behalf. He has not reviewed its contents yet.

### Motivation
- Pasted/search-filter URLs that include decoded `conditions` were being overwritten at app startup by previously persisted local query state, causing pasted filtered links to show unfiltered results.
- The issue affects the search state initialization logic in `web/src/store/index.ts` and required a small change to prefer explicit route-provided conditions.
- The intent is to ensure explicit URL filters always take precedence while preserving the existing behavior of restoring saved local state when no URL filters are present.

### Description
- Add a `routeHasConditions(currentRoute)` helper in `web/src/store/index.ts` that detects whether the current route query contains decoded `conditions`.
- Change `init()` so `restoreState()` is only called for logged-in users when the current route does not contain conditions, preventing URL filters from being overwritten.
- Export `routeHasConditions` and add a focused unit test `web/src/store/index.test.ts` that verifies detection of decoded URL conditions, empty condition arrays, and routes without conditions.
- The changes are limited to search-store initialization and tests and retain the original behavior of restoring persisted state when no URL conditions exist.

### Testing
- Ran the focused unit test with `yarn test src/store/index.test.ts`, which passed.
- Ran the full test suite with `yarn test`, and all tests passed (`13` tests in this environment). 
- Ran type checking with `yarn type-check` (`vue-tsc`) and it completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_b_6a44a0974d4c8331afac0c1c84b51be5)

### DIFF
--- a/web/src/store/route.test.ts
+++ b/web/src/store/route.test.ts
@@ -1,25 +1,10 @@
-import { beforeAll, describe, expect, it } from 'vitest';
+import { describe, expect, it } from 'vitest';
 
-let routeHasConditions: typeof import('./index').routeHasConditions;
+import { routeHasConditions } from './route';
 
 function routeWithQuery(query: Record<string, unknown>) {
   return { query } as Parameters<typeof routeHasConditions>[0];
 }
-
-beforeAll(async () => {
-  // The store imports the API client, which reads window.location at module load time.
-  // Provide just enough browser global for this focused helper test.
-  globalThis.window = {
-    location: { origin: 'http://localhost' },
-    localStorage: {
-      getItem: () => null,
-      setItem: () => undefined,
-      removeItem: () => undefined,
-    },
-  } as unknown as Window & typeof globalThis;
-
-  ({ routeHasConditions } = await import('./index'));
-});
 
 describe('routeHasConditions', () => {
   it('detects decoded URL conditions so startup does not overwrite pasted filters', () => {
--- a/web/src/store/index.test.ts
+++ b/web/src/store/index.test.ts
@@ -1,0 +1,43 @@
+import { beforeAll, describe, expect, it } from 'vitest';
+
+let routeHasConditions: typeof import('./index').routeHasConditions;
+
+function routeWithQuery(query: Record<string, unknown>) {
+  return { query } as Parameters<typeof routeHasConditions>[0];
+}
+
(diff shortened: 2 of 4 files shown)


web/src/store/route.test.ts
```diff
@@ -0,0 +1,28 @@
...
+        table: 'study',
+        field: 'principal_investigator_name',
+        op: '==',
+        value: 'Kate Thibault',
+      }],
+    }))).toBe(true);
+  });
+
+  it('treats an empty decoded conditions list as no URL filter state', () => {
+    expect(routeHasConditions(routeWithQuery({ conditions: [] }))).toBe(false);
```
eecavanna, line 22: I don't know that I agree with this treatment. I think the implication is: Someone could copy/paste a link to what they are looking at, and send it to a friend; and the friend would visit the link and their locally-saved filter would get loaded and used, resulting them seeing something different from the sender.

web/src/store/route.test.ts
```diff
@@ -0,0 +1,28 @@
...
+  });
+
+  it('treats routes without conditions as restorable from saved local query state', () => {
+    expect(routeHasConditions(routeWithQuery({ view: 'data-types' }))).toBe(false);
```
eecavanna, line 26: Same comment as https://github.com/microbiomedata/nmdc-server/pull/2233#discussion_r3504575922

pkalita-lbl:
I don't think this fully resolves the issue. There are related problems with keeping the query state persisted between a logged-out and logged-in state. I'm going to take a closer look and either add to this PR or start a new one.

PR #2235, opened 2026-07-01
Make `Study Description` text area scrollable for review

### LINKED ISSUE #1641: BUG: can't see all submission context during review
See 

Image

When I'm reviewing a submission, I'd like to be able to read this whole description or have the ability to expand the window.

FYI
@JamesTessmer @pkalita-lbl

JamesTessmer:
This PR resolves https://github.com/microbiomedata/issues/issues/1641

Previously the `Study Description` text area was not scrollable or expandable once a submission was submitted for review.
To fix this, the text area is now no longer disabled, but is set to read only and styling has been added to make it visually align with the disabled components in the form.

### DIFF
--- a/web/src/views/SubmissionPortal/Components/StudyForm.vue
+++ b/web/src/views/SubmissionPortal/Components/StudyForm.vue
@@ -180,7 +180,6 @@ export default defineComponent({
             v-model="studyForm.description"
             label="Study Description"
             :hint="Definitions.studyDescription"
-            :class="{ 'readonly-as-disabled': formRef?.isDisabled }"
             persistent-hint
             variant="outlined"
           >
--- a/web/src/views/SubmissionPortal/Components/StudyForm.vue
+++ b/web/src/views/SubmissionPortal/Components/StudyForm.vue
@@ -180,8 +180,6 @@ export default defineComponent({
             v-model="studyForm.description"
             label="Study Description"
             :hint="Definitions.studyDescription"
-            :readonly="formRef?.isDisabled"
-            :disabled="false"
             :class="{ 'readonly-as-disabled': formRef?.isDisabled }"
             persistent-hint
             variant="outlined"
@@ -611,13 +609,8 @@ export default defineComponent({
     </div>
   </div>
 </template>
-<style scoped>
-.readonly-as-disabled :deep(.v-field),
-.readonly-as-disabled :deep(.v-input__details) {
-  opacity: var(--v-disabled-opacity, 0.6);
-}
-
-.readonly-as-disabled :deep(textarea) {
-  cursor: default;
+<style lang="scss">
+.v-input--disabled, .v-field--disabled {
+    pointer-events: initial;
 }
 </style>
--- a/web/src/views/SubmissionPortal/Components/StudyForm.vue
+++ b/web/src/views/SubmissionPortal/Components/StudyForm.vue
@@ -180,6 +180,9 @@ export default defineComponent({
             v-model="studyForm.description"
             label="Study Description"
             :hint="Definitions.studyDescription"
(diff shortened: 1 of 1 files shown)


eecavanna:
The description and diff both seem fine to me. I'm going to request a review from Codex now.

eecavanna:
@codex review

web/src/views/SubmissionPortal/Components/StudyForm.vue
```diff
@@ -608,3 +611,13 @@ export default defineComponent({
     </div>
   </div>
 </template>
+<style scoped>
+.readonly-as-disabled :deep(.v-field),
```
eecavanna: Side note: I didn't understand this class name at first. I think it's being used to convey "this is a 'read-only' field, and we are styling it to look like a 'disabled' field."

eecavanna (approved):
I reviewed the PR description and the diff. Both make sense to me.

I don't know that including `:disabled="false"` on the `v-textarea` element is necessary (I assume the element is already, by default, not disabled). Maybe it's being included to override something that gets applied at a higher level of the DOM tree.

I did not spin this up locally. Assuming the author did and tested it locally, I'm OK with this being merged in as is.

JamesTessmer:
> I reviewed the PR description and the diff. Both make sense to me.
> 
> I don't know that including `:disabled="false"` on the `v-textarea` element is necessary (I assume the element is already, by default, not disabled). Maybe it's being included to override something that gets applied at a higher level of the DOM tree.
> 
> I did not spin this up locally. Assuming the author did and tested it locally, I'm OK with this being merged in as is.

I found without explicitly setting `:disabled="false"` the state of the form as a whole propagated down and would disable the `textarea`. I wasn't sure if there is a better way to keep that from happening though.

web/src/views/SubmissionPortal/Components/StudyForm.vue
JamesTessmer: That's the interpretation I was going for, I was not 100% confident with the name so if there's an alternative that jumps to mind that sounds better I have no problems dropping that in.

pkalita-lbl:
I'm a little concerned that this fix applies to only the one `v-textarea` mentioned in the issue. There are other `v-textarea`s used elsewhere in the submission portal, plus more may be added later.

The reason that a disabled `v-textarea` becomes non-scrollable (unlike a disabled native `textarea`) is that Vuetify sets `pointer-events: none` on a few disabled-related CSS classes. We can try overriding those to handle this more generically. I did it by adding this to `SubmissionForm.vue` (to keep the change somewhat scoped):

```vue
<style lang="scss">
.v-input--disabled, .v-field--disabled {
    pointer-events: initial;
}
</style>
```

As far as I can tell it was just a stylistic choice on Vuetify's part, and it doesn't break anything to override it.

JamesTessmer:
> I'm a little concerned that this fix applies to only the one `v-textarea` mentioned in the issue. There are other `v-textarea`s used elsewhere in the submission portal, plus more may be added later.
> 
> The reason that a disabled `v-textarea` becomes non-scrollable (unlike a disabled native `textarea`) is that Vuetify sets `pointer-events: none` on a few disabled-related CSS classes. We can try overriding those to handle this more generically. I did it by adding this to `SubmissionForm.vue` (to keep the change somewhat scoped):
> 
> ```
> <style lang="scss">
> .v-input--disabled, .v-field--disabled {
>     pointer-events: initial;
> }
> </style>
> ```
> 
> As far as I can tell it was just a stylistic choice on Vuetify's part, and it doesn't break anything to override it.

Oh yup that seems like a better way to do it. Dropped that in and it worked on the `studyForm` as well so I'll go with this instead.

pkalita-lbl (approved):

PR #2236, opened 2026-07-01
Ensure search conditions are restored on app initialization

### LINKED ISSUE #2230: Data Portal - paste of a filter url does not apply the correct filter

pkalita-lbl:
Fixes #2230 

Supersedes https://github.com/microbiomedata/nmdc-server/pull/2233. Sadly codex really biffed it.

The relevant change in Release 2026.6 is the addition of this call to `router.replace` in `SearchLayout.vue`: https://github.com/microbiomedata/nmdc-server/blob/91cf2b6e95ca9eace5abfb03cb4cce18f5364bbd/web/src/views/Search/SearchLayout.vue#L138-L158

Calling `router.replace` (line 140) triggers a re-stringify-ing of the query params through our custom function: https://github.com/microbiomedata/nmdc-server/blob/91cf2b6e95ca9eace5abfb03cb4cce18f5364bbd/web/src/utils.ts#L40-L63

‼️ Note that this function **modifies the provided query params** by calling `delete params.conditions` (line 59).

Previously this wasn't a problem because the parsed conditions were loaded into application state before `stringifyQuery` was ever called: https://github.com/microbiomedata/nmdc-server/blob/91cf2b6e95ca9eace5abfb03cb4cce18f5364bbd/web/src/store/index.ts#L130

However now `router.replace` is called in an immediately-triggered watcher. This means that `stringifyQuery` gets called before the assignment of `state.conditions`.

So the fix here to **not** modify the params passed to `stringifyQuery`. Now it makes a shallow copy and operates on that, which makes the `delete` call safe. Vue Router's underlying params object is unchanged, and it is intact when `state.conditions` is assigned.

---

Some ancillary changes:

* Previously the `router.push`/`router.replace` call in `setConditions` only set the `conditions` query param. In Release 2026.6 this changed to pass through extra params (like the new `view` and `results` params). This also allowed the login token exchange code to remain the URL longer than needed. These changes add `code` to the list of query params intentionally dropped when calling `setConditions`.
* Previously the `persistState` was too aggressive about when it called `setQueryState`. This causes a bug where `setQueryState` is incorrectly called during the login process. These changes add an additional guard to ensure it is only called when the user is actually in the search interface.

### DIFF
--- a/web/src/store/index.ts
+++ b/web/src/store/index.ts
@@ -48,7 +48,7 @@ function setConditions(conditions: Condition[], push = false) {
       && a.op === b.op
       && a.table === b.table);
   if (router) {
-    const { conditions, q, ...rest } = router.currentRoute.value.query;
+    const { conditions, q, code, ...rest } = router.currentRoute.value.query;
     // @ts-ignore
     router[push ? 'push' : 'replace']({ query: { ...rest, conditions: state.conditions }, name: 'Search' }).catch(noop);
   }
--- a/web/src/store/index.ts
+++ b/web/src/store/index.ts
@@ -31,7 +31,7 @@ const unreactive = {
  */
 function persistState() {
   /* If the user is browsing anonymously, stash their state in case they log in */
-  if (!state.user) {
+  if (!state.user && router && router.currentRoute.value.name === 'Search') {
     setQueryState({
       conditions: state.conditions,
       bulkDownloadSelected: state.bulkDownloadSelected,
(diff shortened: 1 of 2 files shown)


pkalita-lbl:
@codex review

eecavanna (approved):
The diff looks good to me, and I found the PR description helpful in understanding the diff. Also, explaining the root cause in terms of recent code changes helped me understand why this issue appeared when it did. I also want to acknowledge that the base branch is (intentionally) not `main`.

PR #2238, opened 2026-07-01
Fix amplicon ingest failure on `LibraryPreparation` records missing `target_subfragment`

sujaypatil96:
Amplicon data-generation records fail to ingest from MongoDB into Postgres when their producing `LibraryPreparation` has a `target_gene` but no `target_subfragment`. 

`load_amplicon_data` accessed `material_processing_set["target_subfragment"]` unconditionally (guarded only by the presence of `target_gene`), raising `KeyError: 'target_subfragment'`. The error is caught by the top-level handler in `load()`, so each affected record is silently added to `errors["omics_processing"]` and **skipped entirely**.

Per the NMDC schema (nmdc-schema 11.21.0), `target_gene` and `target_subfragment` are both **optional** slots asserted only on `LibraryPreparation`, and are **independent** of one another — so neither can be assumed present, and one can exist without the other.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

<hr />

I looked through the existing test infrastructure in this repo and it looks like there is no Mongo-mocking infrastructure and hence no tests that check loading from a Mongo instance. That's a separate concern in itself so I won't try to tie that into this PR. We can think about and build a more robust mechanism in a separate PR.


eecavanna:
Will summon Codex for an initial review while I take a 2-minute break.

eecavanna:
@codex review

nmdc_server/ingest/omics_processing.py
```diff
@@ -243,27 +243,46 @@ def load_omics_processing(  # noqa: C901
 
 def load_amplicon_data(obj, input_ids, mongodb):
     """
-    Load amplicon-specific fields for omics processing records.
+    Load amplicon-specific fields onto an omics processing (data generation) record.
```
eecavanna: Is "data generation" the modern name for what used to be called "omics processing"? I think we have a mixture of terms in the ingest code, and that having that relationship be stated explicitly would be helpful for readers (including me).

nmdc_server/ingest/omics_processing.py
```diff
@@ -243,27 +243,46 @@ def load_omics_processing(  # noqa: C901
...
         mongodb: MongoDB database connection
     """
+    # `obj` is the destination data generation record being populated; `amplicon_lib_prep`
+    # (below) is the source LibraryPreparation the target_gene/target_subfragment are read from.
+    obj["target_gene"] = None
+    obj["target_subfragment"] = None
+
     for input_id in input_ids:
-        material_processing_set = mongodb["material_processing_set"].find_one(
+        amplicon_lib_prep = mongodb["material_processing_set"].find_one(
```
eecavanna: Note to self:

"Find the document in the `material_processing_set` collection whose `has_output` field contains this `input_id`." The document get saved into a variable named `amplicon_library_prep`.

When I look at the schema docs for https://microbiomedata.github.io/nmdc-schema/MaterialProcessing/, I see the range of the `has_output` slot is `ProcessedSample`. I see the same thing in the schema docs for https://microbiomedata.github.io/nmdc-schema/LibraryPreparation/ (i.e. the range of its `has_output` slot is also `ProcessedSample`). So, I assume `input_id` is the ID of a processed sample.

nmdc_server/ingest/omics_processing.py
```diff
@@ -243,27 +243,46 @@ def load_omics_processing(  # noqa: C901
...
             {"has_output": {"$in": [input_id]}}
         )
-        if material_processing_set and "target_gene" in material_processing_set:
-            obj["target_gene"] = material_processing_set["target_gene"]
-            target_subfragment = material_processing_set["target_subfragment"]
+        if not amplicon_lib_prep:
+            continue
+
+        target_gene = amplicon_lib_prep.get("target_gene")
```
eecavanna: Note to self:

- Schema docs for `target_gene` slot: https://microbiomedata.github.io/nmdc-schema/target_gene/ (range is `TargetGeneEnum`)

nmdc_server/ingest/omics_processing.py
```diff
@@ -243,27 +243,46 @@ def load_omics_processing(  # noqa: C901
...
+        if target_gene is not None:
+            obj["target_gene"] = target_gene
+
+        target_subfragment = amplicon_lib_prep.get("target_subfragment")
```
eecavanna: Note to self:

- Schema docs for `target_subfragment` slot: https://microbiomedata.github.io/nmdc-schema/target_subfragment/ (range is `TextValue`)

sujaypatil96:
Awesome, great catch @eecavanna (made a fix in d45bc6c0)

I fixed the logic such that the `target_subfragment` is always pulled from the same library preparation record that has the original `target_gene` (since subfragment is part of the gene - e.g. V4 is a subfragment of 16S rRNA)

sujaypatil96:
@codex review

nmdc_server/ingest/omics_processing.py
sujaypatil96: Yup, that's right, "data generation" (instances of its sub classes) is the modern name for what used to be called "omics processing".

You're right there is a mixture of language in the codebase at the moment, but I figured that we need to start using the new and correct jargon at some point so thought I would use the right language starting now. But yes, you're right, it'll be good to make a note of that change (as comments in code).

eecavanna:
Since the Codex summoning is working for me and for our teammate PK, but not for you (apparently), I think there is some setup involved in your ChatGPT/Codex app (I can walk you through it). I'll summon Codex now.

eecavanna:
@codex review

sujaypatil96:
I think this PR is ready for another round of review/merge @eecavanna.

nmdc_server/ingest/omics_processing.py
```diff
@@ -243,27 +243,44 @@ def load_omics_processing(  # noqa: C901
...
 def load_amplicon_data(obj, input_ids, mongodb):
     """
-    Load amplicon-specific fields for omics processing records.
+    Load amplicon-specific fields onto an omics processing (data generation) record.
+
+    ``target_gene`` and ``target_subfragment`` are both optional slots asserted only on
+    ``LibraryPreparation`` in the NMDC schema. Both are read from a *single* source
+    LibraryPreparation -- the first one in the input chain that declares a ``target_gene``
+    -- rather than being combined across records: a ``target_subfragment`` describes a
+    region of its ``target_gene``, so the two are only meaningful together. A record whose
```
aclum: This sentence doesn't make sense, target_subfragement can be populated independently of target_gene.

aclum (commented):
I'm skeptical this is doing the correct thing and i can't see any new unit tests. This code needs to be able to process the following situations:
analyte_category of amplicon_sequencing_assay, target_gene not specified, target_subfragment not specified
analyte_category of amplicon_sequencing_assay, target_gene not specified, ,target_subfragment specified
analyte_category of amplicon_sequencing_assay, target_gene specified, ,target_subfragment not specified
analyte_category of amplicon_sequencing_assay, target_gene specified, ,target_subfragment specified

nmdc_server/ingest/omics_processing.py
sujaypatil96: Oh interesting 👀 the schema slot description for [target_subfragment](https://microbiomedata.github.io/nmdc-schema/target_subfragment/) says "name of subfragment of a gene or locus" so I assumed that the gene needs to be identified/populated first and only if that's been populated we can specify a sub fragment region on it after and hence populate `target_subfragment`.

aclum: they are independent and there no linkml rules specifying that one requires the other.

sujaypatil96:
@aclum about the unit tests - I would have loved to add them, but there is no Mongo mocking test infrastructure setup in this repo and so I can't write tests against those cases exactly.

nmdc_server/ingest/omics_processing.py
sujaypatil96: Awesome, thanks for the clarification!

eecavanna:
I have an idea for making it easier to test this function, without introducing Mongo mocking test infrastructure. Since the function's use of Mongo is limited to finding a `material_processing_set` document that meets some criteria, I think changing the function's parameters so the function expects something more specific than a reference to the entire Mongo database (which is what it expects now) is practical and would make it easier for us to do dependency injection within doctests. I'll implement what I have in mind (on this branch) and then report back.

eecavanna:
OK, I made it so the function expects to receive (as one of its parameters) a reference to a callback function. All the Mongo interaction happens within that callback function (as opposed to in the function-under-test). The callback function can be mocked in the function-under-test's doctests.

I included 5 doctests as examples. For example:

```py
>>> obj = {}
>>> load_amplicon_data(obj, ["input_a"], lambda id_: {"target_gene": "MyGene", "target_subfragment": "MySubfragment"})
>>> obj
{'target_gene': 'MyGene', 'target_subfragment': 'MySubfragment'}
```

You can run them via:

```shell
python -m doctest nmdc_server/ingest/omics_processing.py
```

I recommend adding a 6th doctest—one in which the callback function returns a dictionary that has a `target_subfragment` field but lacks a `target_gene` field. The function currently ignores the `target_subfragment` in that situation, which I think is contrary to what @aclum posted in [this PR comment](https://github.com/microbiomedata/nmdc-server/pull/2238#discussion_r3532216495):

> they are independent and there no linkml rules specifying that one requires the other.

eecavanna:
@codex review

eecavanna:
simon says

@codex review

sujaypatil96:
Thanks for making the necessary method modifications to wire in the doctests strategy for unit tests @eecavanna! I modified the logic in accordance with @aclum's comment here: https://github.com/microbiomedata/nmdc-server/pull/2238#discussion_r3532216495

I also added in more new doctests (in addition to the ones you already have @eecavanna) for these use cases: https://github.com/microbiomedata/nmdc-server/pull/2238#pullrequestreview-4639449740

eecavanna:
@codex review

nmdc_server/ingest/omics_processing.py
```diff
@@ -241,29 +251,120 @@ def load_omics_processing(  # noqa: C901
...
+    value on any of them is picked up.
+
+    Following the convention of load_omics_processing, this mutates `obj` in place rather
+    than returning a value.
 
     Args:
-        obj: The omics processing object to populate with amplicon data
-        input_ids: List of input IDs to search for material processing data
-        mongodb: MongoDB database connection
+        obj: The data generation record to populate with amplicon data (mutated in place)
```
eecavanna: Thanks for clarifying the parameter descriptions.

Given the description of the `obj` parameter, here, I would like for the parameter to be renamed from `obj` to `data_generation`. This can either be done as part of this PR, or this PR could contain a "TODO" about it (like the one below), or we can just file a ticket about it and not continue iterating on this branch.

```py
# TODO: Rename `obj` to `data_generation` and add a type hint (`: dict`),
#       so the parameter is less ambiguous to maintainers and callers.
```

nmdc_server/ingest/omics_processing.py
```diff
@@ -241,29 +251,120 @@ def load_omics_processing(  # noqa: C901
...
+                                                      `material_processing_set` document, if any,
+                                                      whose `has_output` field contains the
+                                                      specified ID.
+
+    The doctests below cover the four combinations of ``target_gene`` and
+    ``target_subfragment`` being present or absent, plus edge cases. Run them via
+    `$ python -m doctest nmdc_server/ingest/omics_processing.py`; the Mongo lookup is
+    stubbed with a `lambda` (or a dict's `.get`) mapping an input ID to a LibraryPreparation.
+    A found LibraryPreparation is always non-empty, so the "neither present" case is a
+    document that simply lacks both slots:
```
eecavanna: The doctests already say what they do, so I recommend reducing this to:

```suggestion
    Doctests (these can be run via `$ python -m doctest nmdc_server/ingest/omics_processing.py`):
```

eecavanna (approved):
Looks good to me. Thanks for implementing this and iterating on it, @sujaypatil96! I left a couple comments about (a) comments and (b) variable names. Neither is a deal-breaker, from my perspective.

eecavanna:
In this repo, merging is blocked until all PR comments have been resolved. You can resolve them whenever you're comfortable.

image

nmdc_server/ingest/omics_processing.py
```diff
@@ -184,7 +184,17 @@ def load_omics_processing(  # noqa: C901
 
     # Get amplicon specific fields
     if obj["omics_type"] == "Amplicon":
-        load_amplicon_data(obj, input_ids, mongodb)
+
+        def find_material_processing_having_id_in_output(id_: str) -> Optional[dict]:
```
aclum: Is there logic somewhere else to make sure the material_processing_set record is of {'type': 'nmdc:LIbraryPreparation'? The schema limits the range of has_output to a processed sample which could  contain records which are not of the correct type.

nmdc_server/ingest/omics_processing.py
```diff
@@ -241,29 +251,120 @@ def load_omics_processing(  # noqa: C901
...
+    A found LibraryPreparation is always non-empty, so the "neither present" case is a
+    document that simply lacks both slots:
+
+    1. Neither ``target_gene`` nor ``target_subfragment`` present:
+    >>> obj = {}
+    >>> load_amplicon_data(obj, ["input_a"], lambda id_: {"id": "nmdc:libprep-1"})
+    >>> obj
+    {'target_gene': None, 'target_subfragment': None}
+
+    2. Only ``target_subfragment`` present:
```
aclum, line 299: why aren't tests 2 and 4 structured like doctest 5. target_subfragment has a range of TextValue, not of String.

aclum: seems like we can get rid of these two to avoid confusion given the other testing coverage

nmdc_server/ingest/omics_processing.py
```diff
@@ -241,29 +251,120 @@ def load_omics_processing(  # noqa: C901
...
+
+        # target_gene and target_subfragment are independent optional slots, so capture each
+        # separately the first time an input's LibraryPreparation declares it.
+        if obj["target_gene"] is None:
+            obj["target_gene"] = amplicon_lib_prep.get("target_gene")
+
+        if obj["target_subfragment"] is None:
+            target_subfragment = amplicon_lib_prep.get("target_subfragment")
+            # target_subfragment's schema range is TextValue, which ingests either as a plain
+            # string or as a `{"has_raw_value": ...}` dict; normalize both to the raw string.
```
aclum: the ingest behavior is not consistent?

nmdc_server/ingest/omics_processing.py
sujaypatil96: There's no explicit logic to ensure that the returned `material_processing_set` record is of type `nmdc:LibraryPreparation` - at least in the `load_omics_processing()` method where the `load_amplicon_data()` method is used. The logic being used is simply saying return the `material_processing_set` record where the processed sample id that is "input to" (asserted as `has_input` on) a `nmdc:DataGeneration` record is "output on" (asserted as `has_output` on) a `material_processing_set` record.

nmdc_server/ingest/omics_processing.py
sujaypatil96: True, that makes sense! good catch @aclum. `target_subfragment` is `TextValue` as enforced by the schema and there's no possibility of it ever just being a plain string so we are good to get rid of doctests that are testing this behavior.

nmdc_server/ingest/omics_processing.py
sujaypatil96: I modified the logic so that it doesn't deal with the case where it tries to "fallback" to an assumed case of "target_subfragment" possibly being a plain string. It strictly assumes that it is a `TextValue` (which it is) and parses out the expected string from its `has_raw_value`. Not sure if that's what you meant by "ingest behavior is not consistent" and if this fix addressed that concern?

nmdc_server/ingest/omics_processing.py
eecavanna: From a coding standing, we can **add** that condition (i.e. only consider `material_processing_set` documents whose `type` is `"nmdc:LibraryPreparation"`).

The current, type-agnostic approach was carried over from the original code (it is like that on the `main` branch).

I assume the more strict version @aclum is describing is the more **correct** approach. I'll check in with @sujaypatil96 on Slack about whether he's planning to implement that on this branch—if not, I'll implement it this evening.

eecavanna: Done via commit https://github.com/microbiomedata/nmdc-server/pull/2238/commits/5ee63378efd38664ad738da340a1e3f894de262d. Resolving conversation.

nmdc_server/ingest/omics_processing.py
eecavanna: Fixed via commit https://github.com/microbiomedata/nmdc-server/pull/2238/commits/ad87fdd3a51bb00de01ba2e9b8077111f38ab90f. Resolving conversation.

nmdc_server/ingest/omics_processing.py
eecavanna: This is another case where the original code (i.e. the code still on `main`) was written to handle `target_subfragment` values that were _either_ dicts or strings.

https://github.com/microbiomedata/nmdc-server/blob/0f13cb45096ef97a0c1fc9d5beed889b2f0dabb3/nmdc_server/ingest/omics_processing.py#L260-L263

Since we know the `target_subfragment` slot has a range of `TextValue` (per the [schema docs](https://microbiomedata.github.io/nmdc-schema/target_subfragment/)), we can **both** (a) remove the code that handles the case where it's a string, and (b) remove the corresponding doctests.

I'll do that now.

eecavanna: Sorry, I saw this conversation and assumed the issue was still outstanding. @sujaypatil96 already did (a) and (b) via commit https://github.com/microbiomedata/nmdc-server/pull/2238/commits/115479d06b3777fb5f2c2378c4e7404875928110. I will resolve this conversation now.

nmdc_server/ingest/omics_processing.py
eecavanna: I think @aclum was referring to the fact that the comment says:

> `target_subfragment` [...] INGESTS either as a plain string or as a [...] dict

I assume that wording came from Claude. If the comment said "CONTAINS" instead of "INGESTS", I don't think this would have come up.

The comment is gone now and the function is no longer designed to tolerate `target_subfragment` values that aren't dicts, so I'll resolve this conversation.

eecavanna:
@codex review

eecavanna:
Once the Codex review comes back clean, I plan to make one optimization to the function, which is to stop processing `input_ids` once `data_generation["target_gene"]` and `target_generation["target_subfragment"]` are both non-`None`. That will avoid unnecessary Mongo queries.

nmdc_server/ingest/omics_processing.py
```diff
@@ -241,29 +261,99 @@ def load_omics_processing(  # noqa: C901
...
+        # separately the first time an input's LibraryPreparation declares it.
+        if data_generation["target_gene"] is None:
+            data_generation["target_gene"] = amplicon_lib_prep.get("target_gene")
+
+        if data_generation["target_subfragment"] is None:
+            # target_subfragment's schema range is TextValue, so a present value is a
+            # `{"has_raw_value": ...}` dict; store its raw value (absent -> left as None).
+            target_subfragment = amplicon_lib_prep.get("target_subfragment")
+            if isinstance(target_subfragment, dict):
+                data_generation["target_subfragment"] = target_subfragment.get("has_raw_value")
```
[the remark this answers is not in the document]
eecavanna: This behavior is intentional. The schema docs say that the range of the `target_subfragment` slot is `TextValue`, which must be a dictionary, not a plain string. Resolving conversation.

eecavanna:
@codex review

eecavanna (approved):

nmdc_server/ingest/omics_processing.py
```diff
@@ -241,29 +261,107 @@ def load_omics_processing(  # noqa: C901
...
+        # value from the LibraryPrepation we found earlier.
+        if data_generation["target_gene"] is None:
+            data_generation["target_gene"] = amplicon_lib_prep.get("target_gene")
+
+        # If the `target_subfragment` field of the `DataGeneration` is still `None`, use the
+        # `target_subfragment.has_raw_value` value from the LibraryPrepation we found earlier.
+        if data_generation["target_subfragment"] is None:
+            target_subfragment = amplicon_lib_prep.get("target_subfragment")
+            if isinstance(target_subfragment, dict):
+                data_generation["target_subfragment"] = target_subfragment.get("has_raw_value")
```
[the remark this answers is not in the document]
sujaypatil96: According to the `nmdc-schema` documentation about the `target_subfragment` slot here: https://microbiomedata.github.io/nmdc-schema/target_subfragment/ you can see that the range of the slot is `TextValue` which in JSON is represented as a dict. So in theory, the value on that slot/key will never be a scalar/plain string in the database. For values such as `V4` that are quoted above, they will be stored on a `has_raw_value` key within the dict asserted on `target_subfragment`. So therefore, we don't need a handler for the scalar/plain string case.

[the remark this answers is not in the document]
eecavanna: I agree. Resolving convo.

eecavanna:
The changes on this branch look good (and make sense) to me. I think I heard the availability of one of the prospective reviewers will be limited tomorrow due to quarterly activities. Given that the scope of the change is (in my opinion) narrow and I know what to look for in the ingest logs (to see the impact of these changes on the original symptom), and given how early we are in the release cycle, I am in favor of merging this into `main` already, so that we can see its effects on real-world data (without setting up a local ingest environment, which is an even lower-risk option).

PR #2241, opened 2026-07-02
add unsaved image dialogue to `SubmissionForm` pages

### LINKED ISSUE #2237: Add popup for users that attempt to navigate away from PI/Study image upload before clicking upload
Per the parent issue https://github.com/microbiomedata/nmdc-server/issues/1788 , right now there's no warning for users in the case that they have an image selected for upload (but haven't actually uploaded it) and they navigate away from the page. 

To fix this some kind of notification or dialog should appear when a user tries to navigate away without uploading, informing them that they haven't actually uploaded. Once they resolve the dialog if they navigate away without uploading it shouldn't reappear.

Do we want to add a button to the dialog that allows them to click an upload button there?
Options could be 'go back', 'upload and continue', and 'continue without uploading' in that case.

JamesTessmer:
This PR resolves https://github.com/microbiomedata/nmdc-server/issues/2237 

These changes add a dialog that appears if a user tries to navigate away from a page where they've selected an image to upload but they did not click the `upload` button. These changes are applied to `SubmissionForm` so they should be present in any future image upload options added to other forms as well.

This functionality is added by adding a new bit of state to the store that tracks any pending selected images. When an image is uploaded that pending image is cleared from the set. 

Considerations:
Do we want to add an 'upload pending images and continue' option? 
Are there cases where `ImageUpload` would be included outside of Submission Portal forms?

### DIFF
--- a/web/src/views/SubmissionPortal/Components/SubmissionForm.vue
+++ b/web/src/views/SubmissionPortal/Components/SubmissionForm.vue
@@ -1,8 +1,9 @@
 <script setup lang="ts">
-import { computed, nextTick, onMounted, useTemplateRef, watch } from 'vue';
+import { computed, nextTick, onMounted, useTemplateRef, watch, ref } from 'vue';
 import { isEqual } from 'lodash';
 import { SubmissionEditorRole } from '@/views/SubmissionPortal/types.ts';
 import { useSubmissionStore } from '../store';
+import { onBeforeRouteLeave } from 'vue-router';
 
 const {
   allowedRoles = ['owner', 'editor'],
@@ -17,6 +18,8 @@ const emit = defineEmits<{
 }>();
 
 const store = useSubmissionStore();
+const leaveDialog = ref(false);
+const pendingNext = ref<(() => void) | null>(null);
 
 const formRef = useTemplateRef('formRef');
 const isDisabled = computed(() => store.getUneditableReason(allowedRoles, inSampleSetContext) !== undefined);
@@ -51,6 +54,25 @@ onMounted(() => {
   void validate();
 });
 
+onBeforeRouteLeave((to, from, next) => {
+  if (store.hasPendingImageUploads) {
+    leaveDialog.value = true;
+    pendingNext.value = next;
+  } else {
+    next();
+  }
+});
+
+function handleLeaveDialog(leave: boolean) {
+  leaveDialog.value = false;
+  if (leave) {
+    pendingNext.value?.();
+    pendingNext.value = null;
+  } else {
+    pendingNext.value = null;
+  }
+}
+
 defineExpose({
   validate,
   isDisabled,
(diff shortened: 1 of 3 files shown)


web/src/views/SubmissionPortal/Components/SubmissionForm.vue
```diff
@@ -65,4 +87,35 @@ defineExpose({
...
+    persistent
+    max-width="500"
+  >
+    <v-card>
+      <v-card-title class="text-h5">
+        Unsaved Image
+      </v-card-title>
+      <v-card-text>
+        You have an image selected that hasn't been uploaded yet. If you leave now, it will be lost.
+      </v-card-text>
```
pkalita-lbl: I don't think any extra styling is needed here, so I'd use the `title` and `text` props on `v-card`:

```vue
<v-card
  title="Unsaved Image"
  text="You have an image..."
>
```

web/src/views/SubmissionPortal/Components/SubmissionForm.vue
```diff
@@ -65,4 +87,35 @@ defineExpose({
...
+      <v-card-actions>
+        <v-spacer></v-spacer>
```
pkalita-lbl: I believe having the `v-card` in a `v-dialog` already takes care of the action button alignment, so this `v-spacer` likely isn't doing anything.

pkalita-lbl (commented):
I like the overall approach. My one concern is that the `next` callback function on Vue Router navigation guards (like `onBeforeRouteLeave`) is deprecated (see docs [1](https://router.vuejs.org/guide/advanced/navigation-guards.html#Optional-third-argument-next), [2](https://router.vuejs.org/api/interfaces/NavigationGuard.html)). So I'd like to avoid introducing a new use of it.

Instead of using `next`, use the return value of the guard function: `true` to proceed with navigation, `false` to prevent it. I think we can also take advantage of the fact that guards can be async. This could work by:

1. The navigation guard calls and awaits an async function that opens the confirmation dialog and then returns a promise for a boolean value representing the user's choice in the confirmation dialog.
2. Let the promise be resolved by the user's action in the dialog (i.e. `handleLeaveDialog` calls the promise's resolve function)

Something like this:

```ts
const leaveDialogOpen = ref(false);
// Note: no re-rendering needed when these change so these are not refs
let leaveDialogPromise: Promise<boolean> | null = null;
let leaveDialogResolve: ((value: boolean) => void) | null = null;

function confirmLeave() {
  if (leaveDialogPromise) {
    return leaveDialogPromise;
  }
  leaveDialogOpen.value = true;

  leaveDialogPromise = new Promise<boolean>((resolve) => {
    // Stash the resolve function for use in handleLeaveDialog
    leaveDialogResolve = resolve;
  }).finally(() => {
    // Tidy up whenever the promise settles
    leaveDialogOpen.value = false;
    leaveDialogPromise = null;
    leaveDialogResolve = null;
  });
  return leaveDialogPromise;
}

onBeforeRouteLeave(async () => {
  if (!store.hasPendingImageUploads) {
    // No pending uploads, allow navigation
    return true;
  }
  // There are pending uploads, confirmLeave will show the confirmation dialog and return a promise that resolves to
  // true or false based on the user's choice. Await that promise and return its result to determine whether to allow
  // navigation.
  return await confirmLeave();
});

function handleLeaveDialog(leave: boolean) {
  if (leaveDialogResolve === null) {
    return;
  }
  leaveDialogResolve(leave);
}
```

JamesTessmer:
I like the async implementation, I added it like you have outlined and made the rest of the updates. Tested again and didn't need to make any changes from what you provided for that.

pkalita-lbl (approved):

PR #2256, opened 2026-07-16
Improve UI for reviewer assignment

### LINKED ISSUE #1767: Improve UI for Reviewer assignment
@codytodonnell  had some good suggestions when reviewing #1757 :

> 1.  We could change the "Resume" button for reviewers to say something like "Review" or "View".
> 2. We could add some message in the dialog indicating that the reviewer was added successfully. It was a little unclear to me initially that the assignment went through.

Additionally it would be nice to be able to view if there are already reviewers assigned to a submission.

JamesTessmer:
Resolves https://github.com/microbiomedata/nmdc-server/issues/1767

This PR adds two new UI elements to the `SubmissionList` relating to reviewer assignments.

The first addition is a snackbar that provides feedback on whether or not assigning a reviewer was successful. Currently if there was an error then no error code is reported since they didn't seem very human readable, it just says 'something went wrong' which might be too vague.

The second addition is a `v-card` within the add reviewer dialog that displays the reviewers that are already assigned to that submission. Part of adding this included refetching the submissions after adding a reviewer so that this displayed list included the new reviewer (in the case that multiple reviewers might be added at the same time, so the user would see the correct list immediately)

### DIFF
--- a/web/src/views/SubmissionPortal/Components/SubmissionList.vue
+++ b/web/src/views/SubmissionPortal/Components/SubmissionList.vue
@@ -58,7 +58,7 @@ const options = ref({
 const isDeleteDialogOpen = ref(false);
 const deleteDialogSubmission = ref<SubmissionMetadataSlim | null>(null);
 const isReviewerAssignmentDialogOpen = ref(false);
-const reviewerAssignmentSnackbarOpen = ref(false);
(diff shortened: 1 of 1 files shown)


codytodonnell (approved):
Code looks good to me. I didn't open the UI for this branch but the changes you described sound good

pkalita-lbl (approved):

PR #2259, opened 2026-07-16
Send `permission_level` of `owner` with submission creation response

### LINKED ISSUE #2257: Incorrect permissions reported for newly created submission
1. Create a new submission
2. From the newly created submission's Submission Summary page, navigate to the Study Information page

**Expected**: I should be able to make updates to the forms on the Study Information page
**Actual**: I get a banner saying "Your current permission level for this submission does not allow editing of this page. Contact the submission author to request a change in permission level." and the forms are disabled.

Note: If I reload the page, the banner goes away and I'm able to edit the Study Information forms. This only happens immediately after creating the submission.

pkalita-lbl:
Fixes #2257 

Currently in `main` only the `GET /api/metadata_submission/{id}` endpoint calculates the permission level for a given submission and user combination, then populates the `permission_level` field. With the reworking of the submission state store in #2240, that `GET` request is no longer made after creating a submission. Instead the response from the `POST /api/metadata_submission` request directly hydrates the submission record in the store. That's good because it's one less unnecessary API request. However it was leaving `permission_level` unset and causing the user-facing bug described in the issue.

The fix here is to set the `permission_level` field on the response of `POST /api/metadata_submission`. In this case the value does not have to be calculated. The user who initiated the request is by definition the owner of the submission.

An existing test of the `POST /api/metadata_submission` endpoint has been updated to add the new expecation.

### DIFF
--- a/nmdc_server/api.py
+++ b/nmdc_server/api.py
@@ -1775,7 +1775,11 @@ async def submit_metadata(
     db.add(owner_role)
     db.commit()
     crud.try_get_submission_lock(db, submission.id, user.id)
-    return submission
+    submission_metadata_schema = schemas_submission.SubmissionMetadataSchema.model_validate(
+        submission
+    )
+    submission_metadata_schema.permission_level = models.SubmissionEditorRole.owner.value
+    return submission_metadata_schema
 
 
 @router.post(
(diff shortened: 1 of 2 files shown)


JamesTessmer (approved):
Tested locally, fixes the issue for me.

PR #2260, opened 2026-07-17
Update ingester to retry HTTP requests it sends to third-party servers

### LINKED ISSUE #2234: Update ingester to retry HTTP requests to third-party web servers
nmdc_server/ingest/kegg.py

Ingest sometimes fails with the error:

```
('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))
```

I think having the ingester retry those requests will prevent some of those failed. There is a lot of unnecessary overhead involved in restarting the entire ingest because an HTTP request failed once.

eecavanna:
On this branch, I updated the ingester so that it retries the HTTP requests it sends to third-party servers.

It will retry them up to 10 times, backing off some number of seconds each time (see formula below, and note that I have set the `backoff_factor` to `1`).

> image

Since the author wrote "most errors are resolved immediately by a second try without any delay", I interpret that to mean that the sequence of events will be: try 1, **no delay**, [try 2 or] **retry 1**, wait 2s, retry 2, wait 4s, retry 3, wait 8s, retry 4, etc.

It will only retry the request if the response code was one of these:

```py
        status_forcelist=[
            status.HTTP_502_BAD_GATEWAY,
            status.HTTP_503_SERVICE_UNAVAILABLE,
            status.HTTP_504_GATEWAY_TIMEOUT,
            status.HTTP_429_TOO_MANY_REQUESTS,
        ],
```

Fixes https://github.com/microbiomedata/nmdc-server/issues/2234

### DIFF
--- a/nmdc_server/ingest/common.py
+++ b/nmdc_server/ingest/common.py
@@ -5,6 +5,7 @@ from time import perf_counter
 from typing import Any, Dict, List, Optional, Set, Union
 
 import requests
+from fastapi import status
 from pydantic import BaseModel
 from requests.adapters import HTTPAdapter, Retry
 from sqlalchemy.exc import IntegrityError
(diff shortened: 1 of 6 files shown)


pkalita-lbl (approved):

codytodonnell (approved):
Looks clean and makes sense to me

PR #2262, opened 2026-07-17
Render a single empty row instead of no rows when no data is present

### LINKED ISSUE #2261: DataHarmonizer cell validation comments not shown consistently
1. Load the Sample Metadata screen for a submission for the first time. 
2. Fill in the "sample name" column for row one and click Validate.

At this point DataHarmonizer flags a number of cells as invalid (expected) and shows an indicator in the corner of the invalid cells denoting that the cells have a comment (expected). However, hovering over the cell **does not** show the comment containing the validation message:

Image

Additionally, if I reload the page and DataHarmonizer initializes with the previously saved data the validation comments **do** appear as expected:

Image

pkalita-lbl:
Fixes #2261 

The bug was caused by a DataHarmonizer/Handsontable quirk. Further explanation in code comment.

### DIFF
--- a/web/src/views/SubmissionPortal/harmonizerApi.ts
+++ b/web/src/views/SubmissionPortal/harmonizerApi.ts
@@ -234,7 +234,11 @@ export default class HarmonizerApi {
     if (!this.ready.value) {
       return;
     }
-    await this.dh.loadDataObjects(data);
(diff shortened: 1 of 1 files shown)


JamesTessmer (approved):
worked for me locally, thanks for the explanatory comment.

PR #2264, opened 2026-07-17
Fix amplicon data tables

### LINKED ISSUE #2247: Amplicon table layout requires horizontal scrolling to access rightmost columns
The amplicon table within a biosample's expanded "Amplicon" panel is narrower than its contents (columns). There is a lot of empty space to the right of the table.

Here is a "screencast" showing the issue. In this video, I am scrolling horizontally using trackpad gestures. 

https://github.com/user-attachments/assets/f6b51849-29dd-48c5-bb85-6cbc4940ae9e

codytodonnell:
Resolves #2247 

All other data object tables are wrapped in a `<v-card>` however the `<AmpliconObjectDataTable>` components were not. I've now added that wrapper so that the amplicon tables extend to the full width of the result box and have a border consistent with the other omics tables.

Screenshot 2026-07-17 at 2 56 59 PM

### DIFF
--- a/web/src/components/AmpliconObjectDataTable.vue
+++ b/web/src/components/AmpliconObjectDataTable.vue
@@ -80,20 +80,22 @@ export default defineComponent({
 
 </script>
 <template>
-  <v-data-table
-    :headers="headers"
-    :items="items"
-  >
-    <!-- eslint-disable-next-line -->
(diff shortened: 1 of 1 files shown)


eecavanna (approved):
Great! Thanks for fixing this.

PR #2267, opened 2026-07-18
Streamline post-ingest Slack message bullet points when possible

### LINKED ISSUE #2266: Streamline post-ingest Slack message sent by ingester
It currently looks like this:

Image

Proposal:
- If the "extracted" and "loaded" numbers equal one another, just show one number in that bullet, and label it accordingly (e.g. "extracted and loaded")
- If the "extracted" and "loaded" numbers do not equal one another, add an emoji (e.g. ❌ or ⚠️)

eecavanna:
On this branch, I made it so the Slack messages sent by the ingester at the end of the ingest process are more concise when possible (so that Slack users can process them at a glance, rather than having to spend time comparing numbers which are usually equal to one another).

Fixes #2266

### DIFF
--- a/nmdc_server/ingest/common.py
+++ b/nmdc_server/ingest/common.py
@@ -67,7 +67,14 @@ class ETLReport:
         self.num_loaded: int = 0
 
     def __str__(self) -> str:
-        """Get a single-line representation of the ETL report."""
+        """
+        Get a single-line representation of the ETL report.
+
+        >>> etl_report = ETLReport()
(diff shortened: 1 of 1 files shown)


eecavanna:
Merging without human review. Rationale:
- Changes are trivial and very minor, in my opinion
- Changes were tested in isolation by running `python -m doctest ...`
- AI review by Codex came back with no complaints
- I want to test them with tonight's nightly dev ingest, and it is the weekend

Issue #2213, opened 2026-06-23
Omics type bar chart bars have gray lines on right edge


eecavanna:
Omics type bar chart bars have gray, 0-value "slivers" on their right sides.

<img width="836" height="466" alt="Image" src="https://github.com/user-attachments/assets/ed5dfe5d-268c-47e4-b460-44b4bef1e0d6" />

App version:

<img width="446" height="167" alt="Image" src="https://github.com/user-attachments/assets/2caf5207-bc24-4857-b9a8-5a4ac0935e7f" />

```json
{
  "nmdc_server": "1.18.1.dev149+g21badc607",
  "nmdc_schema": "11.20.2",
  "nmdc_submission_schema": "11.20.1"
}
```

Browser: Chrome on macOS

codytodonnell:
I banged my head on this one for a while but as it turns out, those grey lines have existed the whole time, they just weren't very visible because of the purple-colored bars. I believe this is an intentional feature of google charts. It just indicates explicitly that there are zero "No match" items.

eecavanna:
Interesting!

> those grey lines have existed the whole time

You're right—I confirmed they exist in prod:

<img width="652" height="310" alt="Image" src="https://github.com/user-attachments/assets/76385891-6a42-4891-8c1f-25b8def2d58a" />

> intentional feature of google charts. It just indicates explicitly that there are zero "No match" items.

Hmm... I wonder whether it's a consequence of a "stacked" bar (instead of a non-stacked bar) being used here.

https://github.com/microbiomedata/nmdc-server/blob/eece3da2da3ccb1c57436881dc5898def0375084/web/src/components/Presentation/FacetBarChart.vue#L141

Anyway, thanks for looking into this!

Issue #1824, opened 2025-10-27
Develop a frontend test suite


codytodonnell:
It would be useful to have frontend tests that can be run locally and in CI that test UI functionality. We have discussed using a tool like [Playwright](https://playwright.dev/) to do this.

This would be a good thing to implement during or directly after the Vueitfy 3 upgrade (#1573) to give us some assurance that everything is working as expected.

eecavanna:
Here's a link to a ticket (in a different repository, which is private) that contains a **checklist** of user-facing aspects of the Data & Submission Portal that I try to confirm work normally before declaring a new deployment as "working." I'm posting it here for reference. This is the closest thing I'm aware of to a UAT (user acceptance testing) checklist for the Data & Submission Portals.

https://github.com/microbiomedata/nmdc-cloud-deployment/issues/113

PR #2278, opened 2026-07-22
Remove no match bars from data portal

JamesTessmer:
Addresses #2213 

This PR removes the 'no match' bars from the data portal 'Data Types & Map' page.
This is accomplished by removing the `excludedCount` variable and related code.

### DIFF
--- a/web/src/components/Presentation/FacetBarChart.vue
+++ b/web/src/components/Presentation/FacetBarChart.vue
@@ -71,9 +71,6 @@ const chartData = computed(() => {
       { label: 'Match', role: 'data' },
       { role: 'scope' },
       { role: 'style' },
-      { label: 'No Match', role: 'data' },
-      { role: 'scope' },
-      { role: 'style' },
       { role: 'annotation' },
     ]
   ];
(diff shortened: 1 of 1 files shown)


JamesTessmer:
@codytodonnell This might not be the right solution to removing those extra grey bars (or if we really care that much)

codytodonnell:
I'm not sure this is what we want to do. Does this remove the "No match" bars for all instances of the bar chart? We definitely want to have the stacked "No match" bars when there is a filter applied. When there are no filters applied (or all the data in the chart "match" the search) is when we could consider hiding the "No match" bars. Although, I'm starting to see them more as a feature than a bug so I would be in favor of keeping them intact.

JamesTessmer:
Gotcha, I think that's fair. Do we want to close the issue as well or I can take a shot at only removing the "No match" bars when there are no filters. I'm mostly a little light on things to do so I'm scrolling through the open issues.

codytodonnell:
I would say let's leave it for now until it becomes an issue. Something (slightly larger) that needs work is the frontend test suite issue https://github.com/microbiomedata/nmdc-server/issues/1824. Any progress on that would be really helpful. Feel free to add yourself to that.

PR #2280, opened 2026-07-22
Fix non-interactive sankey bug

### LINKED ISSUE #2265: Sankey diagram lacks interactivity when using filter: `Data Type [is] Amplicon`
When I have no search filters in place, the Sankey diagram is interactive.

Likewise, when I have the `Data Type [is] Proetomics` search filter in place, the Sankey diagram remains interactive.

Image

However, when I have the `Data Type [is] Amplicon` search filter in place, the Sankey diagram is no longer interactive. Also, when going from `Data Type [is] Proetomics` to `Data Type [is] Amplicon`, I didn't notice any change in the content of the Sankey diagram, which I think is a related symptom (e.g. maybe the filtered set of samples lacks the metadata that the Sankey diagram uses, and [maybe] the UI was not designed to handle that situation and instead shows "stale" data).

Image

codytodonnell:
Resolves #2265 

The ecosystem sankey diagram would break if it didn't have any data. This was observed when the Amplicon filter was applied because some (perhaps all) of these data do not have ecosystem metadata on their biosample records. The sankey component is now configured to show a "No results or no ecosystem data for this search" message when this occurs. Previously it would show a stale (non-interactive) diagram or a blank panel.

Screenshot 2026-07-22 at 2 05 01 PM

### DIFF
--- a/web/src/components/EcosystemSankey.vue
+++ b/web/src/components/EcosystemSankey.vue
@@ -8,6 +8,7 @@ import { api, type Condition } from '@/data/api';
 import { makeTree } from '@/util';
 import useRequest from '@/use/useRequest';
 import LoadingOverlay from '@/components/LoadingOverlay.vue';
+import { GoogleChartWrapper, GoogleVizEvents } from 'vue-google-charts/dist/types';
 
 export interface EcosystemSankeyProps {
   conditions?: Condition[];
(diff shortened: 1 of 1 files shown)


eecavanna (approved):
Thanks for diagnosing and fixing this! I support this being merged in today.

PR #2310, opened 2026-08-03
Update `SubmissionList` with features to make navigating between submissions easier

### LINKED ISSUE #2309: Add tools for reviewers to find the 'right' submission
Some feedback from Montana and reviewers since the multiple sample set updates have indicated that it can currently be frustrating to find the right submission to review. 

In this example, there are three submissions with the same name (which should be alleviated in the future by the multiple sample sets updates) and it's hard to tell which one is the 'correct' one for the reviewer to click on. This is compounded by the fact that resuming a submission removes all filters on the `SubmissionList` table, so if a reviewer clicks the wrong submission then they have to go back, re-enter the filters, and find the right submission.
Image

To fix this we can do a few things, first is to provide some way for reviewers (or anyone else) to get a 'preview' of the submission without having to change pages. The second is to add an option to resume a submission in a new tab or have filters persist.

JamesTessmer:
Resolves #2309 

These changes add some changes for reviewers (and other users) to use that provide some ease of use.

This PR updates `SubmissionList` to use route links instead of the resume button, and adds filter persistence. 
The first change is more in line with the data portal, adding consistency. It also allows users to open submissions in new tabs and windows.
The second change allows users to keep their selected filter values in `SubmissionList` if they navigate back to that page (in the case of a misclick on the wrong submission or any other reason).

Both of these changes should alleviate some user feedback associated with the linked issue.

Screenshot 2026-08-07 at 12 53 08 PM

### DIFF
--- a/web/src/views/SubmissionPortal/Components/SampleSetTable.vue
+++ b/web/src/views/SubmissionPortal/Components/SampleSetTable.vue
@@ -16,35 +16,51 @@ import {
 } from '@/views/SubmissionPortal/types.ts';
 import PageSection from '@/components/Presentation/PageSection.vue';
 
+const props = withDefaults(defineProps<{
+  sampleSets?: SubmissionSampleSetListItem[];
+  compact?: boolean;
+  showActions?: boolean;
+}>(), {
+  sampleSets: undefined,
+  compact: false,
+  showActions: true,
+});
 
-const headers: DataTableHeader[] = [
-  {
-    title: 'Sample Set Name',
-    value: 'name',
-    sortable: true,
-  },
-  {
-    title: 'Templates',
-    value: 'templates',
-    sortable: false,
-  },
-  {
-    title: 'Status',
-    value: 'status',
-    sortable: true,
-  },
-  {
-    title: 'Last Modified',
-    value: 'date_last_modified',
-    sortable: true,
-  },
-  {
-    title: '',
-    value: 'action',
-    align: 'end',
-    sortable: false,
-  },
-];
+const headers = computed<DataTableHeader[]>(() => {
+  const baseHeaders: DataTableHeader[] = [
+    {
+      title: 'Sample Set Name',
+      value: 'name',
+      sortable: true,
+    },
+    {
+      title: 'Templates',
+      value: 'templates',
+      sortable: false,
+    },
+    {
+      title: 'Status',
+      value: 'status',
+      sortable: true,
+    },
+    {
+      title: 'Last Modified',
+      value: 'date_last_modified',
+      sortable: true,
+    },
+  ];
+
+  if (props.showActions && !props.compact) {
+    baseHeaders.push({
+      title: '',
+      value: 'action',
+      align: 'end',
+      sortable: false,
+    });
+  }
+
+  return baseHeaders;
+});
 const router = useRouter();
 const store = useSubmissionStore();
 const isDeleteDialogOpen = ref(false);
@@ -72,7 +88,11 @@ const isContributor = computed(() => {
 
 const sampleSet = ref<SubmissionSampleSetListItem[]>([]);
 onMounted(async () => {
-  sampleSet.value = store.submission.record!.sample_sets;
+  if (props.sampleSets) {
+    sampleSet.value = props.sampleSets;
+    return;
+  }
+  sampleSet.value = store.submission.record?.sample_sets ?? [];
 });
 
 watch(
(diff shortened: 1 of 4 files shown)


web/src/views/SubmissionPortal/Components/SampleSetTable.vue
```diff
@@ -72,7 +88,11 @@ const isContributor = computed(() => {
 
 const sampleSet = ref<SubmissionSampleSetListItem[]>([]);
 onMounted(async () => {
-  sampleSet.value = store.submission.record!.sample_sets;
+  if (props.sampleSets) {
+    sampleSet.value = props.sampleSets;
```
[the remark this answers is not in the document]
JamesTessmer: Hmm I missed this while testing, but double checking it is something that's needed.

web/src/views/SubmissionPortal/Components/SampleSetTable.vue
```diff
@@ -206,7 +226,39 @@ async function handleStatusChange(item: SubmissionSampleSetListItem | null, newS
...
+          </div>
+        </template>
+
+        <template #[`item.date_last_modified`]="{ item }">
+          {{ new Date(item.date_last_modified + 'Z').toLocaleString() }}
+        </template>
+      </v-data-table>
+    </v-card>
+  </div>
+  <PageSection v-else>
```
pkalita-lbl: I'm concerned about the level of code duplication here. I can't quite see why `SampleSetTable` ever rendered a `PageSection`. In general, the `PageSection` component should only need to be rendered by "page-level" components like `SubmissionSummary`. If `PageSection` is removed, I think the differences needed for `compact` and `showActions` can be much more localized, e.g. 

```vue
<v-data-table
   ...
   :density="compact ? 'compact' : 'default'"
>
```

pkalita-lbl (commented):
I have an alternate suggestion regarding the "resume in new tab" feature. In general, I'm always in favor of following standard browser semantics, rather than inventing something new. In this case, that means I would question whether we need the "Resume" button at all anymore:

Image

Navigating via a button is less standard than navigating via a link. If we render a link, then users already know they can right click it to get standard new tab and new window options. This is already what we do in the Data Portal when presenting search results:

Image

The sample set preview feature seems mostly fine. I have some specific comments on the implementation. Although, overall I'm not entirely sure I understand what problem it's solving.

web/src/views/SubmissionPortal/Components/SampleSetTable.vue
JamesTessmer: Good point. I had kind of noticed that there was a lot of duplication as well and hadn't considered adjusting it to use a ternary operator or removing the use of `PageSection`.

JamesTessmer:
> The sample set preview feature seems mostly fine. I have some specific comments on the implementation. Although, overall I'm not entirely sure I understand what problem it's solving.

Montana had said she and some reviewers were having some trouble finding the 'right' submissions and being able to see that sample set metadata without fully navigating into a submission would be helpful, so there's some demand to see it on the `SubmissionList`. I think it's mostly of a side effect of transitioning to the multiple sample sets model, there's the combination of new UI and still having multiple submissions that should really all be sample sets within a single submission. It's also partly due to filters not persisting the same way they do in the data portal so it' a little more punishing to resume the 'wrong' submission. 
Actually adding filter persistence is probably a better solution here. @mslarae13 Do you think we need the preview, or is the combination of opening in a new tab + keeping filters/search text enough?


I like the idea of adding the router-link navigation to add some consistency with the data portal.

JamesTessmer:
@pkalita-lbl Made some updates. The more I was thinking about the more I think that the preview table is not the right solution so I reverted that. For some context I think the original ask from Montana was to have that preview with the sample set table with the functionality of setting sample set status from that preview. I moved away from that a little because I think being able to edit sample sets outside of the submission itself is a bad pattern, so we wound up with just the preview which was not as useful.

Instead the route links are present now, and there's filter persistence for the `SubmissionList` table. I *think* this should solve the original issue Montana was highlighting (using the example in the image).
Screenshot 2026-08-07 at 1 00 28 PM

I think the preview table does solve the issue of 'which of these identically named submissions is the one I want to look at', but is less useful outside of that (like you pointed out) whereas the link and filter persistence are more useful in other contexts as well.

pkalita-lbl (approved):
I think this looks good! I could go either way on whether we persist the filters _allll the way_ to local storage. Like, if I search for "grass", find the submission I want, work on it, close the page, come back tomorrow and "grass" is still in the search box, would I be surprised in a good or bad way? I'm not sure. But let's go with it for now and see if we get feedback on it.

JamesTessmer:
> I think this looks good! I could go either way on whether we persist the filters _allll the way_ to local storage. Like, if I search for "grass", find the submission I want, work on it, close the page, come back tomorrow and "grass" is still in the search box, would I be surprised in a good or bad way? I'm not sure. But let's go with it for now and see if we get feedback on it.

Yeah that was the initial implementation and after testing it felt unintuitive. I think with the change to use `sessionStorage` should keep the behavior more in line with the data portal filters.

PR #2335, opened 2026-08-21
Fix duplicate workflow executions and data objects in Data Portal UI

### LINKED ISSUE #2334: Data Portal- incorrect file and size counts for download when related to manifest_set records
Issues with workflow results display when records are associated with a manifest_set record.

duplicate records of the same analysis type for records that are associated with manifest set. ex https://data-dev.microbiomedata.org/?view=data-types&results=samples&q=Ch4ICBAaGAoiFiJubWRjOmJzbS0xMS1keDc3djc2OCI
ex nmdc:wfmgan-11-fjqht609.1 is shown twice if you select "All" for items per page.

codytodonnell:
Resolves #2334 

There are some cases where the same WorkflowExecution is part of multiple DataGenerations. We were not handling this in the Data Portal UI previously. Now, wfe and dobjs are deduped before the data object tables are displayed. Also, they are deduped on the backend when calculating the download summary file counts and sizes.

### DIFF
--- a/nmdc_server/aggregations.py
+++ b/nmdc_server/aggregations.py
@@ -278,57 +278,58 @@ def get_data_object_aggregation(
         for workflow in WorkflowActivityTypeEnum
     }
 
-    # TODO: we could join this into one query with a union, but it might not be worthwhile
-    # aggregate workflows
-    workflow_rows_query = (
+    # A data object can be associated with multiple matching data generations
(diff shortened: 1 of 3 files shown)


pkalita-lbl (approved):

aclum:
confirmed fixed on dev

PR #2348, opened 2026-08-26
Avoid "duplicating" `bulk_download` rows unnecessarily within SQL query (ingest)

### LINKED ISSUE #2347: Nightly ingest is failing with an OOM error (out of memory)
I already increased Memory by a whopping 6 Gi earlier today (from 10 Gi to 16 Gi), thinking that would leave us plenty of headroom. The OOM error is still happening. I have not managed to get an exact reading of how much memory is being used.

The error happens at this point in the ingest process:

```
INFO  [nmdc_server.jobs] Copying dependent data from the portal database to the ingest database.
INFO  [nmdc_server.jobs] Merging download-related data: Starting.
```

I suspect there's a bug somewhere because an increase of > 6 Gi of Memory usage since the previous day does not seem consistent to me with how much data is being handled.

CC-ing @pkalita-lbl so he knows there is a ticket about this.

eecavanna:
On this branch, I configured a query that takes place during ingest, to use lazy loading for a specific column. It had been using eager loading (and joining) by default, which was causing Memory consumption to spike (since the newly-introduced `rocrate_metadata_cache` column can contain "a lot" of data on a given row).

"a lot", when multiplied many times (once per related data object). It was enough to cause a Kubernetes node to terminate an ingest pod, reporting an OOM error.

I also made the logging statements more granular.

Fixes #2347

### DIFF
--- a/nmdc_server/jobs.py
+++ b/nmdc_server/jobs.py
@@ -108,9 +108,36 @@ def do_ingest(function_limit, skip_annotation) -> Dict[str, ETLReport]:
             #       and the latter is ingested from the MongoDB database.
             #
             logger.info("Copying dependent data from the portal database to the ingest database.")
-            with duration_logger(logger, "Merging download-related data"):
+            with duration_logger(logger, "Merging download-related data (FileDownload)"):
                 maybe_merge_download_artifact(ingest_db, prod_db.query(models.FileDownload))
-                maybe_merge_download_artifact(ingest_db, prod_db.query(models.BulkDownload))
+            with duration_logger(logger, "Merging download-related data (BulkDownload)"):
+                # Load `bulk_download` rows without automatically joining them with related
+                # `bulk_download_data_object` rows (which would consume memory unnecessarily).
+                #
+                # Note: `BulkDownload.files` is a so-called "back-reference" that gets created as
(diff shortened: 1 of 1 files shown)


nmdc_server/jobs.py
```diff
@@ -108,9 +108,36 @@ def do_ingest(function_limit, skip_annotation) -> Dict[str, ETLReport]:
...
+                # `bulk_download_data_object` rows (which would consume memory unnecessarily).
+                #
+                # Note: `BulkDownload.files` is a so-called "back-reference" that gets created as
+                #       part of the `bulk_download` relationship on `BulkDownloadDataObject`. The
+                #       former is configured to use "joined eager loading". That means that, by
+                #       default, whenever we query `BulkDownload`, each `bulk_download` row will
+                #       automatically be JOINed with each `bulk_download_data_object` related to it
+                #       (related via `BulkDownload.files`). Not only does it get _loaded_, it also
+                #       gets _joined_ with the original `bulk_download` row.
+                #
```
eecavanna, line 124: Oops. I did not mean to leave this last sentence in there.

```py
                #       [...] Not only does it get _loaded_, it also
                #       gets _joined_ with the original `bulk_download` row.
```

It's a remnant of an intermediate version of the paragraph.

PR #2364, opened 2026-09-04
Fix race condition with query parameter loading

### LINKED ISSUE #2337: Data Portal - incorrect behavior when copy & pasting a url with a filter
env: dev and prod, the example below is from dev.

When i have a filter applied (like search by biosample ID) sx ( ID [like] nmdc:bsm-11-dx77v768), which generates this url: https://data-dev.microbiomedata.org/?view=data-types&results=samples&q=Ch4ICBAAGAQiFiJubWRjOmJzbS0xMS1keDc3djc2OCI and then I copy the filter into a new window the results are a mix of correct and incorrect results. The number of samples and studies are correct, the numbers in the bar chart and the number of files, size of files is incorrect. This filter should select 1 metagenome from the bar chart and one set of analysis files of each type per workflow execution (e.g. I expect 1 Assembly Coverage Stats file,  not 4272).

observed:

Image

expected:

Image

codytodonnell:
Resolves #2337

Now we ensure that the app is not mounted until the router state is initialized with any existing query parameters from the URL. This ensures that the search components are initially rendered using filtered data instead of a possibly stale call for unfiltered data.


web/src/main.ts
```diff
@@ -25,4 +27,10 @@ if (typeof sentryDsn === "string" && sentryDsn.length > 0) {
   });
 }
 
-app.mount('#app')
+// Wait until the initial URL (including the encoded q parameter) has been parsed, then hydrate
+// the store before mounting any search components. Their immediate watchers can now make their
+// first requests with the correct conditions.
```
eecavanna, line 32: Helpful comment!

web/src/store/index.ts
```diff
@@ -100,6 +100,21 @@ const dataObjectFilter: ComputedRef<DataObjectFilter[]> = computed(() => state
...
+ * Initialize router-dependent search state before search components mount.
+ */
+function initializeRouteState(
+  _router: Router,
+  loginState = '' as string | (string | null)[],
+) {
+  // Retain the router so later condition changes can be synchronized back to the URL.
+  router = _router;
+  if (loginState !== 'submission') {
+    // @ts-ignore
```
eecavanna, line 113: I'm curious why this "ignore" was necessary (given that this is a "low level" thing like routing). No need to reply (thinking out loud).

eecavanna (approved):
I'm not too familiar with this part of the codebase (in particular, the `state` object and Vue's router). I found the comments (and PR description) helpful. Nothing in the diff jumps out at me as being problematic.

eecavanna:
I'm OK with this being merged in now. Thanks for taking care of the issue!

PR #2370, opened 2026-09-10
Implement metadata quality visualizations in the Data Portal

### LINKED ISSUE #2314: Implement metadata quality visualizations on the data portal
Implement the design from microbiomedata/nmdc-server#2313. Complete when users can see metadata ratings on biosamples on the data portal.

codytodonnell:
Resolves #2314 

This PR implements the new metadata quality badges into the Data Portal. This includes 4 main changes:

1. Badge count on sample results in the search page
2. Metadata Quality filter to filter by badge presence/absence
3. Badges displayed on the Sample Page
4. Updated Sample Page UI for readability

The `badges` field will be ingested into the `Biosample` table because currently all fields are ingested and attempted to be added to the `Biosample` table (see [this part of ingest](https://github.com/microbiomedata/nmdc-server/blob/main/nmdc_server/ingest/biosample.py#L129)). If the values don't match the permissible badge values specified in nmdc-schema, then ingesting those badges will fail. There is now a `badges` field in postgres that holds an array of strings.

Samples can be filtered by their presence or absence of 1 or more badges. The default selection for each badge filter is "Any" which means that the filtered samples will not care about the presence or absence of that particular badge.

Screenshot 2026-09-16 at 4 16 53 PM

----------------------------

The badge count on the result page helps users see metadata quality and verify their filter.

Screenshot 2026-09-16 at 4 16 36 PM

----------------------------

The new sample page formats metadata in a new `LabelValueTable` component which can be used to readably display a flat list of label-value pairs. There is also a new Metadata Quality section which displays each earned badge. The help link next to the page section title navigates to the `MetadataBadgeEnum` schema docs page. Once, docs URLs are added to the schema for each permissible value, we can also link out to individual badge pages when badges are clicked. Note that we also now display all metadata for a biosample. These are hidden by default in the table but can be toggled on/off by clicking the bottom row. This takes all the data that are shoved into the `annotations` blob of a Biosample. There are some annotation we fields we want to omit. Those fields are kept as a static list variable on the frontend right now.

Screenshot 2026-09-16 at 4 23 48 PM

----------------------------

Screenshot 2026-09-16 at 4 24 07 PM

----------------------------

AI disclosure: Codex with 5.6 Sol was used to assist with the code in this PR.


web/src/components/Presentation/BiosampleSearchResults.vue
```diff
@@ -128,6 +129,41 @@ function setExpanded(resultId: string, omicsProcessingId: string) {
...
+                v-bind="props"
+              >
+                <v-icon>
+                  mdi-medal
+                </v-icon>
+                <div>{{ (result as BiosampleSearchResult).badges.length }}</div>
+              </div>
+            </template>
+            <span class="d-flex flex-wrap">
+              <span>This biosample has {{ (result as BiosampleSearchResult).badges.length }} metadata quality badges:</span>
```
codytodonnell: Handle plural/singular

web/src/components/Presentation/FilterMetadataQuality.test.ts
```diff
@@ -0,0 +1,66 @@
...
+  const [badge, otherBadge] = badges;
+  const conditions: Condition[] = [
+    { table: 'biosample', field: 'badges', op: 'has', value: badge! },
+    { table: 'biosample', field: 'badges', op: 'lacks', value: otherBadge! },
+    { table: 'biosample', field: 'geo_loc_name', op: '==', value: 'USA' },
+  ];
+  const { emitted } = render(FilterMetadataQuality, { props: { conditions } });
+
+  await userEvent.click(screen.getByRole('radio', {
+    name: `${snakeToSentenceCase(badge!)} Any`,
```
codytodonnell, line 61: Can we change this to `aria-label`

web/src/components/Presentation/LabelValueTable.vue
```diff
@@ -0,0 +1,102 @@
+<script setup lang="ts">
+import { ref } from 'vue';
+import LabelValueTableRow from './LabelValueTableRow.vue';
+
+export interface LabelValuePair {
```
codytodonnell, line 5: Docstrings for props

eecavanna (approved):
We reviewed this live via Zoom and I expressed all my thoughts there. Nothing we encountered were things that I considered to be a deal breaker. I appreciate you taking the initiative to fix little things along the way (e.g. the tooltip padding). It is something I think we all benefit from.

web/src/components/Presentation/BiosampleSearchResults.vue
eecavanna: e.g.

```
0 badges
1 badge
2 badges
...
```

web/src/components/Presentation/LabelValueTable.vue
```diff
@@ -0,0 +1,102 @@
...
+  value?: string | number | null;
+  href?: string;
+}
+
+/**
+ * Displays a flat list of metadata as label-value pairs.
+ */
+withDefaults(defineProps<{
+  defaultRows?: LabelValuePair[];
+  hiddenRows?: LabelValuePair[];
```
eecavanna: Suggestions:

```
defaultRows -> baseRows or mainRows
hiddenRows  -> hideableRows or additionalRows or moreRows
```

Issue #1564, opened 2025-03-06
Prevent duplicate PIs on ingest


naglepuff:
### Background
In the postgres database, there is a table `principal_investigator`, which stores PIs as rows. These are populated during study ingest. This information is extracted from the `principal_investigator` slot. Note that in mongo, PI information is stored inline. This means that if multiple studies share a PI, that information is duplicated. On ingest, we do our best to represent each PI only once:

https://github.com/microbiomedata/nmdc-server/blob/8eb94dd111dbd23755983bcf40a84e715447523a/nmdc_server/ingest/study.py#L20-L23

### Problem
`name` is not a good equality check for PIs. If a PI is expressed differently in two different projects (e.g. "Mike" vs "Michael"), they will be represented twice in our database. The biggest consequence is that there may be two entries in the PI facet for the same person.

### Potential Solution
On our end, we could prevent duplicate PIs by checking ORCid ID instead of name. Each PI should really only have one ORCid ID.

pkalita-lbl:
Yeah the main issue has to do with the entire submission portal -> Mongo -> data portal trip once we allow user uploads in the submission portal.

Let's say a PI already has a study in the data portal with a crusty old headshot that they emailed us at some point. Now they submit a second study through the submission portal. They use the shiny new feature that allows them to upload a lovely recent headshot. They finish their submission, we accept it, it goes into Mongo.

But, uh oh, now we ingest that new study into the data portal. If that new study gets processed _after_ the original one it will show up with the _old_ headshot. The researcher is now rightfully confused thinking, didn't I attach my new headshot to that submission?

I completely agree that database normalization is a great thing. However. Since the upstream source (Mongo) is _not_ normalized in this respect, I feel we're a bit at its mercy. And maybe a bit of denormalization is okay if it makes the submission portal -> Mongo -> data portal flow more predictable?

None of this is set in stone, by the way. It's just in the discussion phase at the moment. But I came across this issue while backlog grooming and figured the issues should be linked.

turbomam:
Measured against prod 2026-08-12, in support of the ORCID-based fix proposed here.

The live `principal_investigator_name` facet has 74 distinct values with exactly two duplicate pairs:

```
'Eoin Brodie'         orcid='orcid:0000-0002-8453-8435'   nmdc:sty-11-dcqce727
'Eoin Brodie'         orcid='orcid:0000-0002-8453-8435'   nmdc:sty-11-2zhqs261
[...]

PR #2376, opened 2026-09-18
Make PrincipalInvestigator many-to-many with Study

### LINKED ISSUE #2332: Update postgres db population to source PI info for study page from `has_credit_associations` instead of `principal_investigator`
See superissue for more information.

Always source information from the `has_credit_associations` slot.  Use the `name` slot (not `has_raw_value`) going forward.

Note that with the next release the class and slot names will change (therefore ingest will need to fix):
`PersonValue` is now called `Person`, and `applies_to_person` slot is now called `applies_to_agent` (see https://github.com/microbiomedata/nmdc-schema/pull/3385)

pkalita-lbl:
Fixes #2332 

These changes give the `PrincipalInvestigator` search table a many-to-many relationship with the `Study` search table (a given study can have multiple PIs). Previously this was a many-to-one relationship (a given study could only have one PI). 

This is primarily implemented through the new `study_principal_investigator_association` table (line 323 of `nmdc_server/models.py` and the associated Alembic migration). The SQLAlchemy `Study` model now has a `principal_investigators` (plural) `relationship` (line 457 of `nmdc_server/models.py`), and the computed `principal_investigator_image_url` property has been moved to the `PrincipalInvestigator` model (line 346). 

The ingest changes start at line 121 of `nmdc_server/ingest/study.py` where we now traverse the `has_credit_associations` list instead of looking at the removed `principal_investigator` object. Any item in `has_credit_associations` with an `applied_roles` of `Principal Investigator` gets put into the `PrincipalInvestigator` table. The `PrincipalInvestigator` instances are attached to the `Study` instances in `create_study` (line 216 of `nmdc_server/crud.py`).

To make querying work, SQLAlchemy just needs to know which relationship to follow to get from a Study to a PI name (it can no longer be inferred like with a direct foreign key relationship). That's why `_association_proxy_keys` needed a third "relationship" element in the tuples (line 104 of `nmdc_server/query.py`) and a minor update to the query at line 764.

The rest of the changes are pretty mechanical renames of `principal_investigator -> principal_investigators` and `applies_to_person -> applies_to_agent`. `web/src/components/TeamInfo.vue` looks like it changed more than it really did because I converted it to use `<script setup>` instead of `defineComponent`; its logic is still fundamentally them same (enforced by a new test).


nmdc_server/models.py
```diff
@@ -430,19 +454,15 @@ class Study(Base, AnnotatedModel):
     omics_counts = query_expression()
     omics_processing_counts = query_expression()
 
-    principal_investigator_id = Column(
-        UUID(as_uuid=True), ForeignKey("principal_investigator.id"), nullable=True
+    principal_investigators = relationship(
+        "PrincipalInvestigator",
+        secondary=study_principal_investigator_association,
+        back_populates="studies",
```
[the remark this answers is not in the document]
pkalita-lbl: That was exactly the purpose of adding `models.Study.principal_investigators` to `_association_proxy_keys` in `nmdc_server/query.py`. I didn't find anywhere else `Study` to `PrincipalInvestigator` joining was needed for queries.

nmdc_server/ingest/study.py
```diff
@@ -46,7 +49,27 @@ def get_or_create_pi(db: Session, name: str, url: Optional[str], orcid: Optional
...
+
+def get_principal_investigator_ids(db: Session, credit_associations: List[dict]) -> List[UUID]:
+    principal_investigator_ids = []
+    for credit_association in credit_associations:
+        if PRINCIPAL_INVESTIGATOR_ROLE not in credit_association.get("applied_roles", []):
+            continue
+
+        pi_obj = credit_association.get("applies_to_agent")
+        if not pi_obj:
+            continue
```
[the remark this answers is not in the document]
pkalita-lbl: `applies_to_person` is truly gone. I added `AliasChoices` as a temporary workaround and should probably just remove it.

nmdc_server/schemas.py
```diff
@@ -246,7 +254,7 @@ class DataObjectAggregationElement(DataObjectAggregationNode):
 class OrcidPerson(BaseModel):
     """https://microbiomedata.github.io/nmdc-schema/PersonValue/"""
 
-    name: Optional[str] = None
+    name: str
```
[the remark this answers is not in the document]
pkalita-lbl: Existing data on dev and prod has already been cleaned up via other processes. Everyone has a name.

nmdc_server/ingest/study.py
```diff
@@ -27,10 +30,10 @@
 )
 
 
-def get_or_create_pi(db: Session, name: str, url: Optional[str], orcid: Optional[str]) -> str:
+def get_or_create_pi(db: Session, name: str, url: Optional[str], orcid: Optional[str]) -> UUID:
     pi = db.query(PrincipalInvestigator).filter_by(name=name).first()
```
codytodonnell, line 34: Is this able to handle PIs who have the same name?

pkalita-lbl: No 😢 

It also doesn't handle the same person with name variants (Mike vs Michael). But that's the way it has always been, and I didn't want to open that can of worms in these changes.

codytodonnell (approved):

nmdc_server/ingest/study.py
codytodonnell: Okay, maybe we should open a ticket for that for posterity

pkalita-lbl: I think it's already tracked in https://github.com/microbiomedata/nmdc-server/issues/1564

nmdc_server/models.py
```diff
@@ -430,19 +454,15 @@ class Study(Base, AnnotatedModel):
...
     )
-    principal_investigator = relationship("PrincipalInvestigator", cascade="all")
-    principal_investigator_name = association_proxy("principal_investigator", "name")
+    # Keep the singular virtual attribute for backwards compatibility
```
eecavanna, line 462: > backwards compatibility 

I'm wondering, "with what?"

eecavanna (approved):
Looks good to me! Thanks for taking care of this! I left a comment that is me thinking out loud. I'm OK with this being merged in as is.

nmdc_server/models.py
pkalita-lbl: Haha intentionally vague. I was sort of unsure how the full text search works, specifically this:https://github.com/microbiomedata/nmdc-server/blob/89cf6460cb971e08bd553c365f2a1343fabb68cc/nmdc_server/ingest/search_index.py#L10-L11

So I thought it might be better to leave the `principal_investigator_name` in.

eecavanna: Ah, gotcha! I thought it might have had something to do with search, but not necessary full text search; just search in general, since people can have URLs bookmarked, and the search query can be encoded in the URL. I'll resolve this convo.

